PR #22, opened 2019-12-19
Rich init

### LINKED ISSUE #21: Project initilisation with metadata
As we gather a good deal of information during the project creation flow it would be convenient to pass metadata during the init of a `GitProject`.

kim:
Closes #21


librad/src/git/mod.rs
```diff
@@ -138,35 +138,34 @@ impl GitProject {
...
-            } else {
-                src_path.file_name()
-            };
-            file_name
-                .map(|f| f.to_string_lossy().to_string())
-                .ok_or(Error::NeedProjectDir)
-        }?;
-
-        let mut pgp_key = key.clone().into_pgp(&profile.nick)?;
+        // TODO: resolve URL ref iff rad://
```
xla:
```suggestion
        // TODO: resolve URL ref if rad://
```

kim: `iff = if and only if` :smartass_emoji:

kim: The reason I wrote that is that, in theory, one can also use any other URL -- but then we need to go out to the internet to fetch it. That sucks. 

Perhaps we can create a separate `Contributor` type which forces the caller to resolve the profile, and also specify if it should be stored inline or as a URL.

xla: :exploding_head:

librad/src/meta/mod.rs
```diff
@@ -10,3 +10,6 @@ pub use common::*;
 pub use contributor::{Contributor, ProfileRef};
 pub use profile::{Geo, ProfileImage, UserProfile};
 pub use project::{Project, Relation};
+
+pub use sequoia_rfc2822::AddrSpec;
+pub use url::Url;
```
xla: We also need `ParseError` to deal with issues when parsing user input.

PR #26, opened 2019-12-20
Wrap `Url`

kim:
As @xla [points out](https://github.com/radicle-dev/radicle-link/pull/22#discussion_r359914612), Rust admits multiple versions of the same dependency to be linked into a program. Friends and acquaintances of `java.lang.ClassLoader` will notice that the smartness of presenting this as a solution to "dependency hell" is bounded. In particular, re-exporting from a dependency allows the dependency to break the public API (pinning to a patch version is a non-solution for a library).

Hence, we wrap the `url::Url` type (exposing only the minimally needed functionality for now). Access to underlying `ParseError`s is permitted via the `.cause()` method, although implementing control flow in them is discouraged.

There might be similar cases already introduced, which will be addressed at a later point.

### DIFF
--- a/librad/src/meta/common.rs
+++ b/librad/src/meta/common.rs
@@ -1,6 +1,7 @@
 use std::fmt;
 use std::str::FromStr;
 
+use ::url as the_url;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
 pub const RAD_VERSION: u8 = 2;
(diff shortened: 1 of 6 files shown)


librad/src/meta/common.rs
```diff
@@ -1,3 +1,68 @@
...
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct Url(url::Url);
+
+impl Url {
+    pub fn parse(s: &str) -> Result<Self, urls::ParseError> {
+        Self::from_str(s)
+    }
+}
+
+pub mod urls {
```
FintanH: Why plural here? Is it to disambiguate from `url`? Could we alias the one above differently possibly?

PR #27, opened 2019-12-20
Provide builder-pattern-style git project initialisation

kim:


kim:
I will only really buy into Rust once it has a `lens` library :trollface:

FintanH:
> I will only really buy into Rust once it has a lens library 

Oh you haven't heard, `lens` isn't boring and is too complicated :upside_down_face:

PR #31, opened 2020-01-09
Surfin'

### LINKED ISSUE #29: Get project handle by project id (for surfin')

kim:
* Provide conversion from GitProject to surf::GitRepository
* Expose top-level `open_project`

Fixes #29


kim:
This is not entirely satisfying, imho: the backend should be opaque for the user of `librad`. To make any `Project` surfable, we'd now need another enum to wrap `surf::GitRepository`, such that we could implement something like `From<Project> for Browsable`.

kim:
@FintanH do you want to start publishing to crates.io? Not sure what happens when using transitive git dependencies, I assume the one here equals to disjoint types from the ones `upstream` is using.

FintanH:
> do you want to start publishing to crates.io? Not sure what happens when using transitive git dependencies, I assume the one here equals to disjoint types from the ones upstream is using.

Ya, I can cut a release version `0.0.1` :)

Issue #16, opened 2019-12-06
Non-empty Set


kim:
`NonEmpty` is not enough for some cases (e.g. maintainer keys). Instead of runtime `dedup`, a proper non-empty set would be nice.

massimiliano-mantione:
For reference, [this](https://crates.io/crates/non-empty-vec) is an alternative implementation of a non empty vector which I find more ergonomic to use and more practical in general since the first element is not special.

In the same way we could have non empty HashSet and HashMap implementations backed by a regular HashSet or HashMap. The main advantages would be:
* the implementation of _all_ read operations would come for free through the Deref trait
* the memory impact and read performance would be _identical_ to the regular imlementation, mutations would have a tiny overhead

kim:
I don't have a _strong_ opinion -- the drawback of this other solution is that the constructor returns `Option`, so you're tempted to use `unwrap` all the time. 

In any case, I am leaning towards having such fundamental datatypes inline in `librad` (as opposed to external crates), so we don't have to jump through hoops for trait impls we might need.

PR #33, opened 2020-01-09
Fixed maintainers dedup.

massimiliano-mantione:
`dedup` works like the unix shell `uniq` cmd: it needs the vector to be sorted.

### DIFF
--- a/librad/src/meta/project.rs
+++ b/librad/src/meta/project.rs
@@ -79,6 +79,7 @@ impl Project {
 
     fn dedup_maintainers(&mut self) {
         let mut xs: Vec<PeerId> = self.maintainers.iter().cloned().collect();
+        xs.sort();
         xs.dedup();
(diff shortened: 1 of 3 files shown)


librad/src/meta/project.rs
```diff
@@ -79,6 +79,7 @@ impl Project {
 
     fn dedup_maintainers(&mut self) {
         let mut xs: Vec<PeerId> = self.maintainers.iter().cloned().collect();
+        xs.sort();
```
FintanH: It seems _sort_ of silly to `sort` every time to de-duplicate. Should we just keep the list in sorted order and use binary_search with insertions to maintain order and avoid insertions if they already exist?

kim: I think we want #16 at some point, but since this list is supposed to be small, the sort is negligible

PR #35, opened 2020-01-16
Use open over open_bare

FintanH:
[`open`](https://docs.rs/git2/0.11.0/git2/struct.Repository.html#method.open) says:

```
Attempt to open an already-existing repository at path.

The path can point to either a normal or bare repository.
```

So if the project already exists then we will have all the refs and what not. Whereas with bare we have nothing :(


FintanH:
Misunderstanding on my part.

PR #36, opened 2020-01-16
Add .rustfmt.toml and matrix build against rust stable + nightly

### LINKED ISSUE #28: rustfmt.toml
Should we put a `.rustfmt.toml` in the repo?

I found that most of the settings either have reasonable defaults, or aren't usable with the stable toolchain. We may still want to run `cargo fmt --check` on CI, if only to avoid "can you reformat this to 100 instead of 180 colums?" reviews :)

kim:
Fixes #28

### DIFF
--- a/.buildkite/docker/rust-nightly/Dockerfile
+++ b/.buildkite/docker/rust-nightly/Dockerfile
@@ -1,0 +1,26 @@
+FROM rustlang/rust:nightly-slim
+
+# Install additional packages
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        clang \
+        curl \
+        file \
+        libclang-dev \
+        libgmp-dev \
+        libssl-dev \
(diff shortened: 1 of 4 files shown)


kim:
@FintanH I actually changed my mind: pinning a library to a toolchain doesn't make much sense. I'm setting up a matrix build now, need to figure out how to not contend too much on build server resources.

FintanH:
> @FintanH I actually changed my mind: pinning a library to a toolchain doesn't make much sense. I'm setting up a matrix build now, need to figure out how to not contend too much on build server resources.

Fair enough, but pinning the formatting and all that seems like a good idea. I'll be interested in what comes out of this :)

kim:
Ok, I think this works until [the machines take over](https://github.com/rust-lang/homu): the slow builds run sequentially (across all agents), but the fast one is the only one that's required as per GH status checks.

PR #38, opened 2020-01-17
Pin to Rust nightly, after all

kim:
libp2p currently requires it.


.buildkite/pipeline.yaml
```diff
@@ -1,38 +1,14 @@
...
-  concurrency: 1
-
-- label: "Nightly Rust"
-  commands:
-  - "cargo test --all"
-  agents:
-    production: "true"
-    platform: "linux"
-  env:
+    #DOCKER_IMAGE: "gcr.io/opensourcecoin/radicle-link-build@sha256:"
```
FintanH: Intentional?

kim: Yah, waiting for that sha to be mined

kim:
Remaining question: do you prefer freedom of artistic expression or should `clippy` also pass on CI? @FintanH @massimiliano-mantione

.buildkite/pipeline.yaml
FintanH: Ya, saw the next commit as soon as I commented :sweat_smile:

FintanH:
@kim: Minimising bike shedding would be good. If someone wants to express themselves then they can add a lint rule and argue for it in review :)

PR #42, opened 2020-01-23
CLI: Refactor option parsing and effectful state setup

kim:
I got a bit annoyed by the way I set this up the first time, namely that I was
instantiating effectful state (storage, pinentry, ...) ad-hoc everywhere, so I
centralisised that.

There is also now the option to override the (equivalent of) XDG_* root by
setting the RAD_ROOT environment variable or passing --paths (oof feel free to
bikeshed that naming choice).

Additionally, the pinentry (i.e. password prompt) for the keystore is now
threaded as part of the type, and doesn't have to be passed to the methods. This
is a breaking change @xla.

### DIFF
--- a/librad/src/keys/mod.rs
+++ b/librad/src/keys/mod.rs
@@ -1,3 +1,5 @@
 pub mod device;
 pub mod pgp;
 pub mod storage;
+
+pub use storage::Storage;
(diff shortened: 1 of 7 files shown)


rad2/src/keystore.rs
```diff
@@ -0,0 +1,15 @@
...
+pub struct Keystore(FileStorage);
+
+impl Keystore {
+    pub fn new(paths: &Paths) -> Self {
+        Self(FileStorage::new(paths.clone()))
+    }
+}
+
+impl Storage for Keystore {
+    fn put_device_key(&mut self, k: &device::Key
```
kim: lol

PR #43, opened 2020-01-23
fmt: Enable match_block_trailing_comma

kim:


kim:
This one triggers my OCD every time: I can accept that the Rust parser doesn't want to consider layout and thus can't parse match arms without any separator at all. But then, ffs, put that damn comma at the end consistently! Punctuation is an anchor when reading.

OCD

xla:
Amen! Will adopt for upstream.

PR #44, opened 2020-01-24
Radicle Re-Export

FintanH:
Expose `radicle-surf` from `librad`

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -19,6 +19,8 @@ fn_args_layout = "Tall"
 
 edition = "2018"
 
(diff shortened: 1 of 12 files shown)


FintanH:
> Ok

Are you sure? You don't sound so sure :stuck_out_tongue_closed_eyes:

kim:
> Are you sure?

Ya, you will need to make 2 PRs still. But a good middleground I guess.

FintanH:
> Ya, you will need to make 2 PRs still. But a good middleground I guess.

Ya it least it's transparent to upstream. Have to keep our clients happy :)

PR #45, opened 2020-01-31
Fix `Paths` to ignore platform conventions when being told so

kim:
I mistakenly assumed `ProjectDirs::from_path` would ignore the platform-dependent
roots (e.g. XDG_* on Linux), but it instead overrides any paths _below_ that.

What we actually want is to be able to store all configuration and state under
a user-defined root directory, which is useful for testing or otherwise
constrained deployments. So this patch fixes that by precomputing things
either in a platform-dependent (`new()`) or user-defined (`from_root`) way.

### DIFF
--- a/librad/src/paths.rs
+++ b/librad/src/paths.rs
@@ -1,4 +1,5 @@
 use std::{
+    collections::HashMap,
     fs,
     io,
     path::{Path, PathBuf},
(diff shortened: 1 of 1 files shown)


librad/src/paths.rs
```diff
@@ -7,38 +8,92 @@ use std::{
 use directories::ProjectDirs;
 
 #[derive(Clone)]
-pub struct Paths(ProjectDirs);
+pub struct Paths {
+    pub keys_dir: PathBuf,
```
FintanH: Is it ok for users to manipulate a constructed `Paths`?

kim: Fair

librad/src/paths.rs
```diff
@@ -7,38 +8,104 @@ use std::{
...
+            Self {
+                keys_dir,
+                projects_dir,
+                profiles_dir,
+            } => [
+                ("keys_dir", keys_dir.clone()),
+                ("projects_dir", projects_dir.clone()),
+                ("profiles_dir", profiles_dir.clone()),
+            ]
+            .iter()
```
FintanH: I think `iter().cloned()` is the same as doing `into_iter()` but not 100% positive on that

kim:
I realised we can get away with borrowing those "base directories" instead of cloning them every time -- partly because one will `join` them with another path anyway.

librad/src/paths.rs
kim: Doesn't seem so, `into_iter()` yields `Iterator<Item=&(&str, &Path)>`, but I need `Iterator<Item=(&str, &Path)>` (i.e. the tuple needs to be cloned, not the values inside).

FintanH: Ah ok, my bad

PR #47, opened 2020-01-31
Factor out keystorage

kim:
As agreed during the offsite.

**Breaking changes**:

* The `Pinentry` method is no longer part of the top-level API, but tied to the `Crypto` used for securing key material. Which in turn only applies to the provided storage implementations -- the `Keystore` trait may interface directly with a system keychain, HSM, or whatever.
* The `FileStorage` implementation requires the path to the key **file** (previously, we were just passing `Paths`, which knows about directories, and chose the file name via magic spell).


radicle-keystore/src/file.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    type PublicKey = PK;
+    type SecretKey = SK;
+
+    type Metadata = M;
+
+    type Error = Error<P::Error>;
+
+    fn put_key(&mut self, key: Self::SecretKey) -> Result<(), Self::Error> {
+        if self.key_file_path().exists() {
```
FintanH: I wonder if there's anything to be said about the fact that the setup of `put_key` is generally the same. There's some existence test, followed by the same storage logic, and ending in how to "persist" the data. This could potentially be abstracted into higher-order functions, but it could also be viewed as over-engineering :upside_down_face: Food for thought

FintanH: Same goes for `get_key` and `show_key` as well I suppose

radicle-keystore/src/memory.rs
```diff
@@ -0,0 +1,122 @@
...
+        Self {
+            key: None,
+            pinentry,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub enum Error<P> {
+    KeyExists,
```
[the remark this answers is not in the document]
kim: There are more variants in the file backend (IO, for one). I know of no way to factor out commons sans structural typing.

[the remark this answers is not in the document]
FintanH: Ya, so if the duplication is worth it avoiding the structural typing then fair enough.

kim:
I'm not exactly sure if we actually want to re-use this for registry keys. If yes, we might want to move it to a separate repo, otherwise maybe not.

/cc @geigerzaehler @NunoAlexandre

radicle-keystore/src/lib.rs
```diff
@@ -0,0 +1,56 @@
...
+pub struct Keypair<PK, SK> {
+    pub public_key: PK,
+    pub secret_key: SK,
+}
+
+pub trait IntoSecretKey<M> {
+    fn into_secret_key(bytes: SecStr, metadata: &M) -> Self;
+}
+
+pub trait HasMetadata<M> {
```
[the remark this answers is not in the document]
kim: Yeah it doesn't really fit: `IntoSecretKey` was the inverse of `From`, but `Into` wouldn't cut it because we need the creation timestamp eventually to pull a PGP key out of thin air. I moved things into a single trait now, but then the `From` bound is still kinda dangling.

radicle-keystore/src/file.rs
kim: I'm not sure I want to argue with `rustc` about higher-order functions :upside_down_face: 

I think with the `Crypto` factored out, it doesn't look so bad: each impl just deals with the details of how things are stored, and additional serialisation where needed.

geigerzaehler:
> I'm not exactly sure if we actually want to re-use this for registry keys. If yes, we might want to move it to a separate repo, otherwise maybe not.

The registry code base does not need a key store. The client is set up in a way that the user needs to provide the key, meaning that upstream could use the key store.

radicle-keystore/src/file.rs
FintanH: lol, sure thang! I just missed Haskell for a moment y'know..

radicle-keystore/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "radicle-keystore"
+description = "Abstraction over storage of cryptographic keys within the Radicle stack"
+version = "0.1.0"
+authors = ["Monadic GmbH <radicle@monadic.xyz>"]
```
xla: Should we try to consistently migrate to dev@radicle.xyz?

radicle-keystore/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+
+[dependencies]
+secstr = "0.3"
+serde = { version = "1.0", features = ["derive"] }
+serde_cbor = "0.10"
+sodiumoxide = "0.2"
+rpassword = "4.0"
+
+[dev-dependencies]
+tempfile = "3.1"
```
xla: Classy!

kim:
@xla are you happy with it staying in this repo?

radicle-keystore/Cargo.toml
kim: Did not know this exists. Such things should probably also be carved into the stone tablets of `radicle-decisions`

xla: Agreed.

xla:
@kim Pretty happy.

PR #51, opened 2020-02-13
[DO NOT MERGE] quic git

kim:
This implements a very simple git HTTP/3 server and client transport using custom TLS certificate validation rules (basically, a self-signed cert using the device key is presented). See inline commentary for pointers where we're not there yet.

One major peculiarity is registering the transport with `libgit`, which needs to happen only once. We will need to rework the library such that it forces the user to run some initialisation logic.

I also thought it might be overall the most straightforward to just use HTTP/3 instead of a custom protocol over plain QUIC, but I'm not sure how smart this is. Basically, my thoughts were: ALPN (application-level protocol negotiation) happens on the connection level, not streams. So, if we negotiate a custom protocol, it will be a bit tricky to "upgrade" an individual stream to HTTP/3 if so desired. Otoh, we could just use HTTP/3 and use normal request matching to dispatch different protocol concerns. This feels a bit clunky, however. Specifically, support for streaming request bodies seems a bit tricky due to having to switch between synchronous (`libgit` transport) and asynchronous code (`quinn`).

Let me know what you think.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,990 +4,1563 @@
 name = "adler32"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5d2e7343e7fc9de883d1b0341e0b13970f764c14101234857d2ddafa1cb1cac2"
 
 [[package]]
 name = "aho-corasick"
 version = "0.7.8"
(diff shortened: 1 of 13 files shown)


kim:
Uhm yeah, the e2e test fails on CI. It works locally though, check it out.

kim:
Superseded by #70

PR #57, opened 2020-03-03
Issue Tracker

### LINKED ISSUE #53: Issue Tracker Semantics
Some relevant conversation happened in [radicle.community](https://radicle.community/t/coco-jammin-on-dags/140).

The definition of an Issue Tracker will exist at the data/application layer. The definition of this tracker and its semantics should be discussed an worked upon.

This task will track the first draft of these semantics and will be agreed up once both the CoCo and Upstream parties are happy with the draft. The first draft will be a Rust library that defines what an Issue is made up of and can be passed over to the Upstream code for use.

FintanH:
Fixes #53 

This is a complete first draft of an issue tracking view. Here's the breakdown of the PR.

* `lib.rs` - contains the export of sub-modules and the `Issue` struct and methods. Look at the crate documentation for a good overview of creating an `Issue`. It shouldn't be too much of a leap to go from a single issue to multiple.
* `metadata.rs` - contains all the metadata around an issue. This includes:
  * `Comment`
  * `Title`
  * `Reaction`
  * `Label`
  * `Assignees`
* `thread.rs` - is where the money's at. It builds an immutable, reply-only thread of "items" which get instantiated to `Comment` in `lib.rs`. It is a zipper structure, so it carries an internal finger to where you're pointing to in the structure. You navigate around by calling `next_reply` or `previous_reply` along with a `ReplyTo` value. The values inside are immutable so they start off as `Live`, and can be marked `Dead` upon deletion (but the data lives on).

Things that are left to do:
* Consider if we want a `Modified` `Status`
* Actually write strategies for the property based testing of `Thread`
* Round out unit testing for `Issue`
* Consider a trait for generating IDs for issues and comments

I think we can probably add these kinds of things as follow up tickets so that others may work on them in parallel.


radicle-tracker/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+impl<Author> Issue<Author> {
+    /// ```
+    /// use radicle_tracker::Issue;
+    ///
+    /// let issue = Issue::new(
+    ///     String::from("Monadic"),
+    ///     String::from("Buggy Boeuf"),
+    ///     String::from("We have bugs in our boeuf"),
```
xla:
```suggestion
    ///     String::from("We have boeuf in our beef."),
```

radicle-tracker/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+    ///
+    /// assert_eq!(issue.author, String::from("Monadic"));
+    /// assert_eq!(issue.title, String::from("Buggy Boeuf"));
+    /// assert_eq!(issue.thread.first().content, String::from("We have bugs in our boeuf"));
+    /// assert!(issue.thread.first().author == issue.author);
+    /// assert_eq!(issue.meta, ());
+    /// ```
+    pub fn new(author: Author, title: String, content: String) -> Self
+    where
+        Author: Clone + Eq,
```
xla: This shold be a CoCo identity once materialised, right?

radicle-tracker/src/lib.rs
```diff
@@ -0,0 +1,63 @@
+use std::hash;
+
+pub mod thread;
+use thread::*;
```
xla: wildcard imports

Wildcard imports...

radicle-tracker/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+
+        Issue {
+            author,
+            title,
+            thread: Thread::new(comment).0,
+            meta: (),
+        }
+    }
+
+    pub fn reply_to_thread(&mut self, path: Path, comment: Comment<Author>) -> Result<Path, ()> {
```
xla: Some doctests for the reply would help illustrate how that works. Esp. keen on understanding the `Path` part. I imagine that to be quintessential to refer into threads in issues.

radicle-tracker/src/lib.rs
```diff
@@ -0,0 +1,63 @@
+use std::hash;
+
+pub mod thread;
+use thread::*;
+
+pub struct Issue<Author> {
```
xla: What about incremental issue numbers known from many trackers? Or differently what is the identity of an issue, how will it be referred to and referenced?

radicle-tracker/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+    pub author: Author,
+    pub title: String,
+    pub thread: Thread<Comment<Author>>,
+    pub meta: (), // TODO(fintan): fill in meta data
```
xla: Do you need input on what could go in here?

radicle-tracker/src/lib.rs
FintanH: You don't like french?

radicle-tracker/src/lib.rs
FintanH: Ya that would be the plan, I would say :)

radicle-tracker/src/lib.rs
FintanH: Ha :smile: They're wildcard internally, but sure I can make it explicit.

radicle-tracker/src/lib.rs
FintanH: For sure, that's the aim for the next iteration :grin:

radicle-tracker/src/lib.rs
FintanH: Ya, I think there's another level above `Issue` which is `Tracker` or something. They'll definitely need an identifier.

xla:
@FintanH Forgot in my initial review, use of `mod.rs` is [discouraged](https://doc.rust-lang.org/edition-guide/rust-2018/module-system/path-clarity.html#no-more-modrs) in `2018` edition. :kissing_cat:

radicle-tracker/src/lib.rs
FintanH: I think I have them documented in https://radicle.community/t/cocoda-issues-codecollab/114 so if there's anything missing there lemme know :ok_hand:

massimiliano-mantione:
The main issue I see is that path elements cannot be simple numbers: we are in a distributed system where comments are created by independent actors, and they cannot generate unique numbers without coordination.

If we go towards the "one event feed per user" model, then each user event (and therefore also each comment) might have one unique id *local to the user*, but even there, a single user could operate on more than one device and those devices might not be online at the same time.

Maybe one sequential integer per device would be feasible but I guess a naturally unique identifier would be beter.

So the type of `Path` elements should be either properly analyzed or abstracted away.

Another issue is that the data structure assumes (or seems to assume) that an issue is *always* loaded in memory fully.
Again, thinking about the "one event feed per user" storage model, this data structure would be the merge of all the feeds, and its physical representation will likely be a git `tree` object (the `RoseTree` would be roughly mapped to nested git `tree` objects).
Loading it requires traversing several nested git `tree`s with `blob`s at the leaves.
I am not sure if the application layer would be ok with not being able to load it partially.

FintanH:
@massimiliano-mantione: I might not have communicated this properly (or at all :grimacing:), but this is essentially the "view" of an issue and its threads. This should happen post user-feed merging and will be the output of the fold over the DAG and operations, so will be the in-memory representation when a user loads it in upstream. Sorry if this was confusing. 

The actions that a user can take on this issue, such as `reply`, will generate operations in the future.

The `Path` is just a way of navigating the issue threads from the root. If it helps you can think of `Issue` and `Threads` to the in-memory `Directory` in `radicle-surf`. The model is defined and you can do a few operations on this model, but the actual generation of the `Directory` structure from `git` is done as a separate process. In the same way, the generation of the `Issue` and `Thread`s will be left up to another layer.

radicle-tracker/src/lib.rs
xla: That makes a lot of sense, so `Tracker` likely be the actual collab-tool package.

FintanH: Ya, likely so. I imagine it would be an organisation of open and closed issues. Or something of the like. It'll need more thought :) Getting a single issue done for now :grin:

massimiliano-mantione:
> The `Path` is just a way of navigating the issue threads from the root. If it helps you can think of `Issue` and `Threads` to the in-memory `Directory` in `radicle-surf`. The model is defined and you can do a few operations on this model, but the actual generation of the `Directory` structure from `git` is done as a separate process. In the same way, the generation of the `Issue` and `Thread`s will be left up to another layer.

I see, and I now see that this is the in memory view of the fold result.
But... does it make sense to generate those synthetic `u32` path components, and then having to map them onto the real IDs that we will have anyway?

My point is that the mapping is not obvious here, but it will be there when we will write the code that converts mutations to the in-memory representation into CRDT operations that we can persist and propagate.
Specifically, we will need to map each of those `u32` values back onto its corresponding global identifier.
I would prefer using the global identifiers in the memory model as well.

radicle-tracker/src/lib.rs
cloudhead: He didn't like that it was a possibly real life scenario :thinking:

radicle-tracker/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        let comment = Comment {
+            author: author.clone(),
+            content,
+        };
+
+        Issue {
+            author,
+            title,
+            thread: Thread::new(comment).0,
```
cloudhead: It's a little unexpected that `Thread::new` doesn't return `Thread`

radicle-tracker/src/thread.rs
```diff
@@ -0,0 +1,162 @@
...
+    pub fn delete(&mut self, path: &Path) -> Result<A, ()> {
+        match self.lut.remove(&path) {
+            Some(ix) => self.tree.remove_node(ix).ok_or(()),
+            None => Err(()),
+        }
+    }
+
+    pub fn edit<F>(&mut self, path: &Path, f: F) -> Option<A>
+    where
+        F: FnOnce(&A) -> Option<A>,
```
cloudhead: Shouldn't this take a `&mut A`?

radicle-tracker/src/lib.rs
FintanH: Well it returns `Thread` and the `Path` to it. Does this break the idiom? It's useful to return the `Path` so that you can continue `reply`ing, `view`ing, or `delete`ing.

I can call it `create` instead to avoid confusion, whatya think?

radicle-tracker/src/thread.rs
FintanH: I thought that at first, but because in the future there may be some ACL semantics, the function should not be able to mutate the `A`.

Unless... the functions hides this away anyway and mutates the `A` if the user accessing is allowed otherwise don't mutate it at all. God mutation semantics are so dirty :grimacing:

radicle-tracker/src/thread.rs
```diff
@@ -0,0 +1,162 @@
+use rose_tree::{NodeIndex, RoseTree, ROOT};
+use std::{collections::HashMap, hash};
+
+/// Laws:
+/// Thread::new(comment).first() === comment
+/// thread.reply(comment).delete(comment) === thread
+/// Thread::new(comment).delete(comment) === None
+/// Thread::new(comment).edit(f, comment) === Thread::new(f(comment).unwrap())
```
cloudhead: Maybe clearer if you say `{delete,edit}(comment_path)`, because it looks like you're passing the actual comment in here.

radicle-tracker/src/lib.rs
cloudhead: Yeah I think it does break the idiom. Can you not call `path()` to get the path?

radicle-tracker/src/thread.rs
FintanH: Ya they need updating! The comments and code in the `test` module below will be more accurate :grin:

radicle-tracker/src/lib.rs
cloudhead: Or if you want to be fancy, you could `impl Into<Path> for Thread`

cloudhead: But not sure if that makes sense, semantically :thinking:

radicle-tracker/src/thread.rs
cloudhead: Oh, but then why isn't `self` mutable? Guess I'm confused :grimacing:

radicle-tracker/src/thread.rs
```diff
@@ -0,0 +1,348 @@
...
+    pub fn paths(&self) -> Vec<Path> {
+        self.lut.keys().cloned().collect()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// forall a. Thread::new(a).first() === a
```
cloudhead: Love the haskellisms

radicle-tracker/src/thread.rs
```diff
@@ -0,0 +1,348 @@
...
+    /// assert_eq!(thread.view(&root_path), Ok(&String::from("Discussing rose trees")));
+    /// assert_eq!(root_path, Path::new(0));
+    /// ```
+    pub fn new(a: A) -> (Self, Path)
+    where
+        A: Eq,
+    {
+        let (tree, root) = RoseTree::new(a);
+        let mut lut = HashMap::new();
+        let path = Path::new(0);
```
cloudhead: Maybe more self-documenting to have a `ROOT` const under `Path`, so you can do `let path = Path::ROOT`. It just requires `Path::new` to be a `const fn`

FintanH:
> But... does it make sense to generate those synthetic u32 path components, and then having to map them onto the real IDs that we will have anyway?
My point is that the mapping is not obvious here, but it will be there when we will write the code that converts mutations to the in-memory representation into CRDT operations that we can persist and propagate.
Specifically, we will need to map each of those u32 values back onto its corresponding global identifier.
I would prefer using the global identifiers in the memory model as well.

That's a good point @massimiliano-mantione. But unfortunately we need some way of navigating the `RoseTree`, so we would like to know what `NodeIndex` (which is generated by `RoseTree`) the 2nd comment on the main thread is. It won't always be the case that it is `NodeIndex(2)`. Another global identifier doesn't help me navigate a tree here. 

So really in the end I think we'll probably end up with indirection as: `CommentId -> Path -> NodeIndex -> Node`. Maybe we could add a few more levels of indirection :thinking: :trollface:

radicle-tracker/src/thread.rs
FintanH: Because if we get Some(new_edit) then you can replace the old value.

Would you say is more Rustier?

radicle-tracker/src/thread.rs
```diff
@@ -0,0 +1,380 @@
...
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum Error {
+    MissingPath(Path),
+    MissingNode,
+}
+
+#[derive(Debug, PartialEq, Eq, hash::Hash, Clone)]
+pub struct Path(Vec<u32>);
```
kim: Could you elaborate a bit on your choice of datastructure, and what might be possible alternatives?

The `Path` type in particular strikes me as a little odd. My mental model of a "thread" (which, admittedly, originates in the 90s) _basically_ only considers the parent -- I'm looking at an entry, and then may decide to reply to it, or I'm interested in the replies to it. If you consider a pathologically long thread, I might want to _collapse_ nodes below a certain threshold, or "pin" a certain node at the root.

In other words, is navigating a tree structure not more like a zipper?

FintanH: Mmmm that could be a good thing to explore.

So you're saying that we could have:
```haskell
Thread = (Thread, [Comment], Thread)

reply :: Comment -> Thread -> Thread
reply comment (left, comments, right) = (left, comments <> [comment], right)
```

But hmmmm... I still think you need a way to "navigate" to a thread. I can think over this a bit and give more elaboration to my choice. We might want to pair on the possible alternatives, since I'm not entirely sure what there is :smile: The `Zipper` version looks isomorphic to the RoseTree version, and at the moment not entirely sure what it buys us; but at the same time I'd need to explore the rest of the functionality more.

kim: >  I still think you need a way to "navigate" to a thread

Sure... but isn't it like browsing any tree-shaped structure? Like, you know a root, list the outgoing children, then focus on the one you're interested in, etc. An absolute path is potentially useful as a "permalink" sort of thing -- but then, you can also address it as `root,left,left,right,left,left` etc.

A zipper representation seems to also map much more obviously to a partially-materialised view (i.e. you don't have to keep the entire tree in memory).

FintanH: Ya great point, lemme work with it!

radicle-tracker/src/thread.rs
cloudhead: I'm probably missing something -- maybe a usage example would help? I just don't understand how you can use this function to mutate the value :confused:

radicle-tracker/src/metadata.rs
```diff
@@ -0,0 +1,240 @@
...
+    collections::{HashMap, HashSet},
+    convert::Infallible,
+    hash::Hash,
+    str::FromStr,
+};
+
+/// The metadata that is related to an issue.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Metadata<User: Eq + Hash> {
+    labels: HashSet<Label>,
```
cloudhead: Are we supposed to use labels to mark an issue as open or closed?

radicle-tracker/src/metadata.rs
```diff
@@ -0,0 +1,240 @@
...
+/// A comment of an issue is composed of its [`Comment::author`], the
+/// [`Comment::content`] of the comment, and its [`Comment::reactions`].
+///
+/// It has a unique identifier (of type `Id`) chosen by the implementor.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Comment<Id, User: Eq + Hash> {
+    identifier: Id,
+    author: User,
+    content: String,
+    reactions: HashSet<Reaction<User>>,
```
cloudhead: Timestamp maybe good to have

radicle-tracker/src/thread.rs
```diff
@@ -0,0 +1,827 @@
...
+
+/// A `Thread` is the root item followed by a series of non-empty replies to the
+/// root item. For each item in reply to the root item there may be 0 or more
+/// replies.
+#[derive(Debug, Clone)]
+pub struct Thread<A> {
+    // A finger points into the `main_thread` structure.
+    // If it is `Left` then it is pointing to the main thread.
+    // If it is `Right` then it is pointing to a reply to a comment in the main thread.
+    _finger: Finger,
```
cloudhead: What's the reason this is prefixed with `_`? Normally that means it's unused..

radicle-tracker/src/thread.rs
```diff
@@ -0,0 +1,827 @@
...
+/// The "liveness" status of some data.
+///
+/// TODO: we may want to consider `Modified`.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum Status<A> {
+    /// The data has been created.
+    Live(A),
+    /// The data has been created and also deleted.
+    Dead(A),
+}
```
cloudhead: This type is a bit confusing, especially given the comment: It seems to represent the data *and* the status, not just a status, even though the type is called `Status`. :thinking: I guess I would expect a type like:
```
struct Data<A> {
  status: Status,
  data: A,
}
```
It would also avoid having to match on the status to get the data. :man_shrugging:

radicle-tracker/src/thread.rs
```diff
@@ -0,0 +1,827 @@
...
+    ///
+    /// thread.root();
+    /// assert_eq!(thread.view(), Ok(&Status::Live(String::from("Discussing rose trees"))));
+    /// #
+    /// #     Ok(())
+    /// # }
+    /// ```
+    pub fn root(&mut self) {
+        self._finger = ROOT_FINGER;
+    }
```
cloudhead: Based on the name, I kind of expect this to return the root, since `root` is one of the fields, and a thread has a root. One suggestion would be to have instead a function like `Thread::navigate_to(finger)` (or call it `goto`), and pass in `Finger::Root` to goto the root.

radicle-tracker/src/thread.rs
```diff
@@ -0,0 +1,827 @@
...
+    /// assert_eq!(thread.view(), Ok(&Status::Live(String::from("Discussing rose trees"))));
+    /// ```
+    pub fn view(&self) -> Result<&Status<A>, Error> {
+        match self._finger {
+            Finger::Root => Ok(&self.root),
+            Finger::Main(main_ix) => Ok(self.index_main(main_ix).first()),
+            Finger::Thread((main_ix, replies_ix)) => Ok(self
+                .index_main(main_ix)
+                .get(replies_ix)
+                .unwrap_or_else(|| panic!("Reply index is out of bounds: {}", replies_ix))),
```
cloudhead: That's kind of like `expect(format!("{}", ...))` (yes it's slightly less efficient but maybe more concise)

cloudhead:
I'm not commenting too much on the product side of things, because we're still shaping that, but it may be that certain fields will have to be added to `Comment` or `Issue`.

radicle-tracker/src/metadata.rs
FintanH: _oof_ no forgot about closing an issue :sweat_smile: I'll add that :)

radicle-tracker/src/metadata.rs
FintanH: Ya, that occurred to me too, thanks! Is there UTC time in Rust? I would imagine that's best, then the app can convert based on the user's local settings

radicle-tracker/src/thread.rs
FintanH: I wanted to convey that this is only something the data structure carries around and uses internally but has nothing to do with the actual data itself. Does that make sense? Should I remove the `_`?

radicle-tracker/src/thread.rs
FintanH: Well, those things are equivalent, but I see your point. I'll think about it :thinking:

radicle-tracker/src/thread.rs
FintanH: This _does_ return to the root. I was thinking about adding a `navigate_to` (HAH! `goto`), but it'll be a similar concept of assigning the `self._finger`. Having a `root` function is still useful.

radicle-tracker/src/thread.rs
FintanH: Clippy will give out to you for that because of evaluation. It's what I had originally :sweat_smile:

radicle-tracker/src/metadata.rs
cloudhead: There's `SystemTime` which could be good enough for now :thinking:

radicle-tracker/src/thread.rs
cloudhead: Ah I see. I would remove it still. Maybe distinguish it from the other fields with a blank line? `_finger` is ignored by the compiler if unused, it has special semantics.

radicle-tracker/src/thread.rs
cloudhead: Yeah if you want to keep the type as-is I'd probably sill suggest a better name and to fix the comment :)

radicle-tracker/src/thread.rs
cloudhead: "return the root" != "return *to* the root" :grin: 

What I find confusing is that the thread has a root and a field called root, yet the method called root isn't a getter, it's actually something that does internal mutation.

radicle-tracker/src/thread.rs
cloudhead: hah I see.

radicle-tracker/src/metadata.rs
kim: Please do _not_ use UTC but a proper zoned local time representation (I would assume the [time](https://docs.rs/time/0.2.7/time/) crate has one). UTC is subject to leap seconds, which are announced by committee decision like approx. 6 months in advance, and that will eventually cause trouble.

radicle-tracker/src/metadata.rs
kim: This is not representing the stored data, right? I think we still haven't made up our minds how we want to deal with issue lifecycle.

If we think of an issue with attached code (a "PR" or "revision"), merging the code to some mainline branch would consequently entail merging the conversation as well (to some _other_ mainline branch? idk). But then, what if it gets rejected? Does that disappear from the project history? What does it mean to "reopen" an issue?

Anyways, since we're concerned with the _surface_ API here (not how it gets stored), maybe a good representation would be to wrap the issue in a newtype `ClosedIssue(Issue)` so as to restrict what the caller _can do_ with that issue. Obviously I haven't thought this through. Probably more complicated, like a `Lifecycle` FSM or something.

radicle-tracker/src/thread.rs
FintanH: Fair, the language level semantics could probably be improved :sweat_smile: 

> What I find confusing is that the Thread has a root and a field called root, yet the method called root isn't a getter for that field (as you would expect), it's actually something that does internal mutation.

That's true... I guess the intention is that `thread.root().view() === thread.root`, but also we want the user to be able to navigate back to the root to execute further navigation from there.

Again, it's probably just the language of the code that needs work :thinking:

radicle-tracker/src/metadata.rs
FintanH: Ah, ok thanks for the call out :)

radicle-tracker/src/metadata.rs
FintanH: Nah, this is just the materialized view of the data. The stored data will come to a head when we talk about the op-based semantics and possibly the Merkle-DAG stuff.

> Anyways, since we're concerned with the surface API here (not how it gets stored), maybe a good representation would be to wrap the issue in a newtype ClosedIssue(Issue) so as to restrict what the caller can do with that issue. Obviously I haven't thought this through. Probably more complicated, like a Lifecycle FSM or something.

This is pretty much what I was thinking too :)

radicle-tracker/src/metadata.rs
FintanH: Would this be what we're looking for? https://docs.rs/time/0.2.7/time/struct.OffsetDateTime.html I haven't been bitten by time yet in my career so some guidance would help :sweat_smile:

FintanH:
For some reason, during some cleanup I brought these changes in: `librad/src/git/mod.rs → librad/src/git.rs `

I sincerely apologise for the confusion.

radicle-tracker/src/metadata.rs
kim: Ya so.. it turns out the Rust time libraries stuff their haggis with... well, you don’t want to look too closely. In particular, they don’t handle leap second nor time zone conversions correctly (this has to take DST into account, which they don’t).

Since the goal here is to merely _display_ completely unreliable timestamps in a halfway consistent manner in the _viewers_ local time zone, I’d amend my previous comment and suggest we type this in a sufficiently imprecise way such that the API doesn’t suggest you can do things with the timestamp which are wrong. Namely, the type shouldn’t implement `Ord` or `PartialOrd`, and should have a precision of no more than, say, 60secs. In other words, a newtype around `SystemTime::duration_since(UNIX_EPOCH)`, which can only be used to render a _relative_ time (“X units ago”).

FintanH: So just to clarify, you're suggesting that we add a newtype for something that we can represent as a date and time local to the user, but can't say use for ordering at all?

We'll also need some kind of `PartialOrd` timestamp field for when it comes to the merkl-dag and operation based stuff so that we have some total order to fall back on.

radicle-tracker/src/metadata/clock.rs
```diff
@@ -0,0 +1,263 @@
...
+
+    fn current_time() -> Self {
+        RadicleClock(SystemTime::now())
+    }
+
+    fn duration_since(&self, other: &Self) -> Result<Duration, Self::Error> {
+        let timestamp = self.0.duration_since(SystemTime::UNIX_EPOCH)?;
+        let other_timestamp = other.0.duration_since(SystemTime::UNIX_EPOCH)?;
+
+        Ok(other_timestamp - timestamp)
```
FintanH: Something I'm not sure about is that this will panic if they're not ordered right :/

FintanH: Should we take the `abs` value or introduce an `enum`, i.e.
```rust
pub enum When<A> {
  Before(A),
  After(A),
}
```

kim: Yeah, the Rust time API is pretty strange. I'm guessing that it is like this because `std` does not have arbitrary precision arithmetic types -- for whatever reason. Luckily, there's a crate for it!
I think we can make all operations total, and more like Haskell `time`, which gets it right:

```rust
use std::{
    fmt,
    ops::{Neg, Sub},
    thread,
    time::{Duration, SystemTime, UNIX_EPOCH},
};

use num_bigint::BigInt;

#[derive(Debug)]
struct TimeDiff(BigInt);

impl fmt::Display for TimeDiff {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        self.0.fmt(f)
    }
}

impl Sub for TimeDiff {
    type Output = TimeDiff;

    fn sub(self, rhs: TimeDiff) -> Self::Output {
        TimeDiff(self.0 - rhs.0)
    }
}

impl From<SystemTime> for TimeDiff {
    fn from(t: SystemTime) -> Self {
        Self(
            t.duration_since(UNIX_EPOCH)
                .map(|d| d.as_secs().into())
                .unwrap_or_else(|e| BigInt::from(e.duration().as_secs()).neg()),
        )
    }
}

trait Clock {
    fn current_time() -> Self;
    fn diff_since(&self, other: &Self) -> TimeDiff;
}

#[derive(Debug)]
struct RadClock(SystemTime);

impl Clock for RadClock {
    fn current_time() -> Self {
        Self(SystemTime::now())
    }

    fn diff_since(&self, other: &Self) -> TimeDiff {
        TimeDiff::from(self.0) - TimeDiff::from(other.0)
    }
}

fn main() {
    let now = RadClock::current_time();
    println!("now: {:?}", now);

    thread::sleep(Duration::from_secs(1));

    let later = RadClock::current_time();
    println!("later: {:?}", later);

    println!("diff: {}", later.diff_since(&now));
}
```

FintanH: Nice :o told you you're a smart :cookie:

One thing I'm not sure about is how the end-user of this is supposed to determine which time came later though? Or does it even matter :thinking:

Thinking about it, I suppose not. Basically they'll grab a comment and do `RadClock::current_time().diff_since(comment.timestamp())` to display the difference. #rubberduck :duck:

FintanH: Ok, I ran with it. This looks better now, imho. Lemme know if you see anything funny

FintanH:
Ping @xla :eyes: Lemme know if you wanna go over the code during a call :call_me_hand:

radicle-tracker/src/metadata/clock.rs
```diff
@@ -0,0 +1,263 @@
...
+}
+
+/// A minimal clock for getting a moment in time and calculating the duration up
+/// to another moment in time.
+///
+/// It does this through its instance of [`Clock`] trait.
+///
+/// For usability, it also provides a function [`RadicleClock::elapsed`] which
+/// returns an [`Elapsed`] value for presentation purposes.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
```
cloudhead: Might be a good idea to explain why there's no `PartialOrd` instance, so that someone doesn't in the future add it, thinking you forgot it :wink:

PR #75, opened 2020-03-17
License headers

### LINKED ISSUE #48: Chore: License headers
The GPL [requires to put a license header into every source file](https://www.gnu.org/licenses/gpl-howto.html), which is also true for many other licenses, should we decide to switch.

So we gotta do what we gotta do. Maybe `rustfmt` / `cargo fix` can [help](https://rust-lang.github.io/rustfmt/#license_template_path).

kim:
Fixes #48

### DIFF
--- a/librad/Cargo.toml
+++ b/librad/Cargo.toml
@@ -1,8 +1,9 @@
 [package]
 name = "librad"
 version = "0.1.0"
-authors = ["Monadic <team@monadic.xyz>"]
+authors = ["The Radicle Team <dev@radicle.xyz>"]
 edition = "2018"
+license = "GPL-3.0-or-later"
 
 [dependencies]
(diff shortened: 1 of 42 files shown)


librad/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "librad"
 version = "0.1.0"
-authors = ["The radicle Team <dev@radicle.xyz>"]
+authors = ["The Radicle Team <dev@radicle.xyz>"]
```
FintanH: Isn't it _supposed_ to be lowercase?

kim: No, I checked with ingrrish and brand rranguage native :)

FintanH: I thought the branding people said it's lowercase :thinking:

PR #76, opened 2020-03-18
Errors: thiserror + anyhow instead of failure

### LINKED ISSUE #72: Use "thiserror" in library code
Instead of `failure`, as per radicle-dev/radicle-decisions#2

It's still not so clear how we want to deal with opaque errors -- perhaps a library `OpaqueError(String)`?

kim:
Fixes #72

_Note that this is on top of #70_


librad/src/git.rs
```diff
@@ -42,72 +43,39 @@ const CONTRIBUTOR_METADATA_FILE: &str = "contributor.json";
...
+    Serde(#[from] serde_json::error::Error),
 
-    #[fail(display = "{}", 0)]
-    Pgp(keys::pgp::Error),
+    #[error(transparent)]
+    Pgp(#[from] keys::pgp::Error),
 
-    #[fail(display = "{:?}", 0)]
+    #[error("Surf error: {0:?}")]
     Surf(surf::error::Error),
```
kim: Nb. `surf::error::Error` is not `std::error::Error`, so we can't use `#[from]` or `#[source]` here. radicle-dev/radicle-surf#87 should fix this

kim:
I'm a bit unsure about the backtrace support -- `failure` adds them automatically, whereas `thiserror` apparently requires the error variant to have a field of type `Backtrace`, which would add quite a bit of annoying boilerplate. Otoh, backtraces are extremely valuable for `ExceptT`-style error handling. I might be misunderstanding how this works, though.

FintanH:
> I'm a bit unsure about the backtrace support -- failure adds them automatically, whereas thiserror apparently requires the error variant to have a field of type Backtrace, which would add quite a bit of annoying boilerplate. Otoh, backtraces are extremely valuable for ExceptT-style error handling. I might be misunderstanding how this works, though.

There's a curious paragraph in `anyhow`:
> A backtrace is captured and printed with the error if the underlying error type does not already provide its own. In order to see backtraces, they must be enabled through the environment variables described in std::backtrace:

The way I read this is that they should be added automatically via `anyhow` as well. We would only add `backtrace` for the convenience of other downstream users :man_shrugging:

librad/src/git.rs
```diff
@@ -42,72 +43,39 @@ const CONTRIBUTOR_METADATA_FILE: &str = "contributor.json";
...
-    #[fail(display = "Invalid PGP key: missing UserID packet")]
+    #[error("Invalid PGP key: missing UserID packet")]
     MissingPgpUserId,
 
-    #[fail(display = "Invalid PGP key: missing address")]
+    #[error("Invalid PGP key: missing address")]
     MissingPgpAddr,
 
-    #[fail(display = "Project already exists")]
+    #[error("Project already exists")]
```
FintanH: Would it help users to know which project doesn't exist?

kim: Yaahaha, that variant doesn’t actually seem to be used anymore :D

FintanH: :fire:

PR #79, opened 2020-03-23
Parametrise gossip message payload

kim:
There's a few rustisms I don't really get, see self-review

### DIFF
--- a/librad/examples/mesh.rs
+++ b/librad/examples/mesh.rs
@@ -26,6 +26,7 @@ use std::{
 use futures::stream::{self, StreamExt};
 use git2::{build::RepoBuilder, FetchOptions, RemoteCallbacks, Repository};
 use log::{info, warn};
+use serde::{Deserialize, Serialize};
 use tempfile::{tempdir, TempDir};
 use tokio::task;
 
@@ -45,6 +46,11 @@ use librad::{
     project::{Project, ProjectId},
     sync::Monitor,
 };
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+struct ProjectCreated {
+    id: ProjectId,
+}
 
 #[derive(Clone)]
 struct MiniPeer {
@@ -76,22 +82,20 @@ impl MiniPeer {
 }
 
 impl gossip::LocalStorage for MiniPeer {
-    fn put(
-        &self,
-        provider: &PeerId,
-        gossip::Update::Project { project, .. }: gossip::Update,
-    ) -> gossip::PutResult {
-        info!("LocalStorage::put: {}", project);
-        let repo_path = self.paths.projects_dir().join(project.path(&self.paths));
+    type Update = ProjectCreated;
+
+    fn put(&self, provider: &PeerId, ProjectCreated { id }: ProjectCreated) -> gossip::PutResult {
+        info!("LocalStorage::put: {}", id);
+        let repo_path = self.paths.projects_dir().join(id.path(&self.paths));
         if Repository::open_bare(&repo_path).is_ok() {
             // Note: we would want to fetch here
-            info!("{}: Project {} already present", self.peer_id(), project);
+            info!("{}: Project {} already present", self.peer_id(), id);
             gossip::PutResult::Stale
         } else {
             info!(
                 "{}: Cloning project {} from {} into {}",
                 self.peer_id(),
-                project,
+                id,
                 provider,
                 repo_path.display()
             );
@@ -108,7 +112,7 @@ impl gossip::LocalStorage for MiniPeer {
                 .bare(true)
                 .fetch_options(fetch_opts)
                 .clone(
-                    &format!("rad://{}@{}/{}", self.peer_id(), provider, project),
+                    &format!("rad://{}@{}/{}", self.peer_id(), provider, id),
                     &repo_path,
                 )
                 .map(|repo| {
@@ -127,17 +131,17 @@ impl gossip::LocalStorage for MiniPeer {
         }
     }
 
-    fn ask(&self, gossip::Update::Project { project, .. }: &gossip::Update) -> bool {
+    fn ask(&self, ProjectCreated { id }: &ProjectCreated) -> bool {
         self.paths
             .projects_dir()
-            .join(project.path(&self.paths))
+            .join(id.path(&self.paths))
             .exists()
     }
 }
 
 struct Bootstrap<'a> {
     peer: MiniPeer,
-    proto: Protocol<MiniPeer>,
+    proto: Protocol<MiniPeer, ProjectCreated>,
     endpoint: BoundEndpoint<'a>,
 }
 
@@ -164,7 +168,7 @@ async fn bootstrap<'a>(
 
     let gossip = gossip::Protocol::new(
         &PeerId::from(key),
-        gossip::PeerAdvertisement::new(endpoint.endpoint.local_addr().unwrap()),
+        gossip::PeerAdvertisement::new(endpoint.local_addr().unwrap()),
         gossip::MembershipParams::default(),
         peer.clone(),
     );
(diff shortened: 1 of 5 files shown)


librad/src/net/gossip.rs
```diff
@@ -64,16 +66,19 @@ pub use storage::*;
...
-pub enum ProtocolEvent {
-    SendAdhoc(Box<Hello>),
-    Connect(Box<Hello>),
+pub enum ProtocolEvent<A> {
+    SendAdhoc(Box<Hello<A>>),
+    Connect(Box<Hello<A>>),
     Disconnect(PeerId),
 }
 
+// Nb.: this type exists only because clippy otherwise complains that
```
kim: I don't understand why it's better to have this `Box` indirection -- isn't `Hello` already a box of sorts?

librad/src/net/gossip.rs
```diff
@@ -265,9 +271,10 @@ impl<S: Clone, R, W> Clone for Protocol<S, R, W> {
     }
 }
 
-impl<S, R, W> Protocol<S, R, W>
+impl<S, A, R, W> Protocol<S, A, R, W>
 where
-    S: LocalStorage + 'static,
+    S: LocalStorage<Update = A> + 'static,
+    for<'de> A: Serialize + Deserialize<'de> + Clone + Debug + Send + Sync + 'static,
```
kim: The `'static` lifetime doesn't seem right here

librad/src/net/gossip/rpc.rs
```diff
@@ -17,25 +17,22 @@
...
-    project::ProjectId,
-};
+use crate::net::gossip::types::{PeerAdvertisement, PeerInfo};
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
-pub enum Rpc {
-    Membership(Membership),
-    Gossip(Gossip),
+pub enum Rpc<A> {
+    Membership(Box<Membership>),
```
kim: Another one of these large enum variant things

librad/src/net/protocol.rs
```diff
@@ -461,9 +462,10 @@ where
 }
 
 #[async_trait]
-impl<S> GitStreamFactory for Protocol<S>
+impl<S, A> GitStreamFactory for Protocol<S, A>
 where
-    S: gossip::LocalStorage + 'static,
+    S: gossip::LocalStorage<Update = A> + 'static,
```
kim: It seems really strange to me that I need to put these trait bounds here -- I get that it's because the `impl` has them, but actually they could be just `forall`. What's the proper way to deal with this?

librad/examples/mesh.rs
```diff
@@ -164,7 +168,7 @@ async fn bootstrap<'a>(
 
     let gossip = gossip::Protocol::new(
         &PeerId::from(key),
-        gossip::PeerAdvertisement::new(endpoint.endpoint.local_addr().unwrap()),
+        gossip::PeerAdvertisement::new(endpoint.local_addr().unwrap()),
```
FintanH: :heart_eyes:

librad/src/net/gossip.rs
FintanH: Yikes, what's the `Debug` for?

librad/src/net/gossip.rs
FintanH: > isn't `Hello` already a box of sorts

I'm not sure they're equivalent. `Box` is a pointer to heap memory right? So it's the size of a pointer. Whereas `Hello` is the size of `PeerInfo` + `Rpc` (give or take). So `Box`ing it would reduce the size to align on, _I think_.

librad/src/net/gossip.rs
kim: Because of `trace!`s -- that's perhaps an indication that these things should be treated first-class (I think I lamented about this before). Although intended for slightly different kinds of applications, I was wondering if we should settle on this: https://docs.rs/tracing/0.1.13/tracing/

FintanH: Ah yes, we did have a chat about this. I don't have strong opinions about this. Maybe we need a radicle-decision? :stuck_out_tongue:

kim: Anyone's got an idea here? It seems to me like the lifetime of `A` should somehow be bound to the lifetime of `LocalStorage`, but I can't seem to figure out how. /cc @cloudhead @FintanH

librad/src/net/gossip.rs
kim: Hm ya. I'll leave it like this for now, perhaps some bestestest practices / idiomaticness folks can explain how this is supposed to be done at some point

librad/src/net/gossip.rs
kim: Not sure this applies to anyone but us (+ upstream as a consumer). Created #81

FintanH: My lifetime-fu is not good enough to tell what's going on here. My _guess_ is that the `for<'de>` is the culprit and because it gets introduced the inference for `A` gets messed up.

What would it usually be inferred as? Would it not usually be `'static`? :thinking:

librad/src/net/gossip.rs
```diff
@@ -64,16 +66,19 @@ pub use storage::*;
...
-    Connect(Box<Hello>),
+pub enum ProtocolEvent<A> {
+    SendAdhoc(Box<Hello<A>>),
+    Connect(Box<Hello<A>>),
     Disconnect(PeerId),
 }
 
+// Nb.: this type exists only because clippy otherwise complains that
+// `ProtocolEvent`'s size might be sub-optimal (enum size is determined by the
+// largest variant).
```
cloudhead: I think that lint isn't so useful here: it's just saying that there's high variance in the size of your enum fields, but that's ok imo. With the `Box` you're creating an indirection which shouldn't be necessary unless you have a very specific performance requirement around this enum, or you want to optimize for the `Disconnect` case here.

librad/src/net/protocol.rs
cloudhead: Yeah, afaik the only way to deal with this is to have no bounds on the `impl`, and keep bounds on the methods only.

kim:
> keep bounds on the methods only.

Oof ok. I went down this road at some point, but then decided I would
need an IDE to maintain such a thing XD

librad/src/net/gossip.rs
cloudhead: Yeah so in my understanding the problem is as follows:
1. `A` is some arbitrary type with some bounds.
2. *iff* the concrete type implementing `A` contains references, then those references are going to need some kind of lifetime.
3. To "prove" to the compiler that it should not worry about any possible references inside of `A` and their lifetime, you have to say that references within `A` are valid for the whole program (`'static`). And this is the case because you are holding `A` in a struct with no lifetime parameters, so it must be `'static`.

cloudhead: Nb. `'static` applied as a *bound* means that the type cannot have any references within it with a lifetime shorter than `'static`

librad/src/net/gossip.rs
kim: Yeah... I wonder _when_ this actually matters, as the lint is enabled by default. I would rather disable it for the entire crate if this is something one should only pay attention to in very specific cases, instead of annotating the exceptions everywhere.

cloudhead: Ah so this is part of the "default" lints? Sounds "pedantic" to me :thinking: 

I'm thinking: if you need to pack this enum in contiguous memory, eg. a `Vec`, then you don't want to waste space if only one of your variants takes lots of space, and all the others are small. In that case it would make sense to have an indirection for the "large" variant, especially if you don't access it as often as the "small" variants. Eg:

```rust
enum Command {
   OpenConnection(Box<HugeStruct>), // only used once per connection
   Send(SmallStruct), // used a lot
}
struct CommandQueue(Vec<Command>);
```

Maybe in the above case it makes sense to box the `HugeStruct`, if you have millions of these, as you'll rarely pattern-match against `OpenConnection`?

kim: Hm yeah. It's the other way round in this case: the small variant should be used less. 

Anyways, I reverted this .. and guess what? `clippy` decided to no longer complain about it :dumbfounded:

cloudhead: Yeah that's what I thought, so really don't think it makes sense.

PR #80, opened 2020-03-24
Box away the git transport stream.

kim:
Previous attempts failed due to poor type inference (see comment of
`impl GitStreamFactory for Protocol`).


cloudhead:
Ahh, great! :fox_face:

PR #83, opened 2020-03-31
Tracing

### LINKED ISSUE #81: Evaluate tracing
# Problem

The `log` crate is alright for ad-hoc output of what a program is doing, but is a bit too _unstructured_ to observe behaviour of a long-running program. It is also the case that some events might be interesting on the application layer (e.g. `pull` progress, peer connection status, ...).

# Potential Solution

While distributed tracing is not exactly applicable to our use-case, the [tracing](https://crates.io/crates/tracing) crate offers nice features, such as setting a context, timing, and structured data. It also appears that by choosing an appropriate subscriber, an application can decide what events they're interested in, and what to do with them.

# Alternatives

A more elaborate solution would be to thread some type of channel through the library, onto which we publish a predefined set of structured-data events, and forego logging entirely (except perhaps at `trace` level).

FintanH:
Fixes #81 

This introduces tracing into librad to replace log.
We import it qualified so that it makes it easy to use any of the
functionality from the lib with updating the imports.

tracing allows us to introduce spans at a top-level so that common
information can percolate down.

The extra benefit to tracing is that we can pass in key, value pairs to
trace statements for structured logging purposes.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,25 @@
+# copied expressions from https://nixos.wiki/wiki/Rust
+# and Mozilla's nix overlay README
+# https://www.scala-native.org/en/latest/user/setup.html
+let
+  moz_overlay = import (builtins.fetchTarball https://github.com/mozilla/nixpkgs-mozilla/archive/master.tar.gz);
+  pkgs = import <nixpkgs> { overlays = [ moz_overlay ]; };
+in
+  with pkgs;
+  stdenv.mkDerivation {
+    name = "clang-env-with-nightly-rust";
+    buildInputs = [
+      (pkgs.rustChannelOf { rustToolchain = ./rust-toolchain; }).rust
+      clang
+      llvmPackages.libclang
+      olm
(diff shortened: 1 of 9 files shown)


default.nix
```diff
@@ -0,0 +1,25 @@
+# copied expressions from https://nixos.wiki/wiki/Rust
```
kim, line 1: You! Sneaky!

librad/src/git/server.rs
```diff
@@ -55,7 +55,7 @@ impl GitServer {
         let mut recv = BufReader::new(recv);
```
kim, line 55: Creating a span here would be useful.

default.nix
FintanH: :grimacing: actually didn't mean to add this :sweat_smile: But while we're here... Would you prefer it didn't live in the repo or is it ok to merge? :eyes:

xla: Nix los hier.

PR #86, opened 2020-04-03
Introduce uri module

### LINKED ISSUE #65: Uniform Resource Locators
## Objective

Formal definition of a URI/URN scheme for entities (tbd) in the radicle network.

### "Physical" Addresses

`radicle-link` can be seen as a forest of **projects**, where the stable **project identifier** maps to a repository name. The "physical" address, however, is in the context of a currently connected-to peer:

`rad://PEER_ID/PROJECT_ID`

where the network address mapped to the `PEER_ID` is transparently handled by the networking layer.

This is so because we are **not** employing a DHT-based network -- the reasons for this are outside the scope of this discussion, but the short answer is that for this to perform acceptably, one has to assume at least `R` permanently connected nodes, where `R` is the replication factor. This is one of the main issues with IPFS.

The path components _after_ `PROJECT_ID` are implementation-defined: in `git` and `pijul`, we can physically address down to named-branch (`ref`) level. `git` actually negotiates the mapping from name to head hash, so in principle it is also possible to content-address the branch. It is **not** currently possible to address down to tree- or blob-level, because the merkle proof of the result requires a commit object. `pijul` could, in principle, address individual patches, but that functionality is not provided as of this writing (and it's also unclear what the benefits would be).

### Logical Addresses

Logically, we want to address:

* **projects**
* **identity documents**, along with their history
    
    The well-known identity documents are those of a **project** and a **user**, but we can generalise this to any document validated by the TUF rules, e.g. modelling ACLs.

* documents containing **signatures over the advertised `refs` of a peer** 
* histories with application-defined semantics, e.g.
  * **issues**
  * **pull requests** ("revisions")
  * **activity feeds**
* histories without interpreted semantics, i.e. **code branches**

All of these have in common that they are named roots of merkle DAGs,e xcept for projects, which are named _forests_ of merkle DAGs. Somewhat unclear is how **strongly bound** some of them are to the context of a **project**. Specifically, the **user identity** is supposed to be globally stable. One might expect, for example, to navigate to the **activity feed** bound to a **user**, without knowing the "canonical" **project** that user is bound to. Also, **issues** (isomorphic to revisions, but without references to code) could arguably appear within _any_ project, while semantically describing a different one.

_Note: a non-issue in this context is that two different projects could be names for the same, well, object database. They could be seen as forks in the original sense of the term, but be combined into the same local checkout as different remotes._ 

## Solutions

1. Overload the project identifier

    Currently, we have `HASH(metadata) '.' STORAGE_BACKEND`. This could be generalised to:

    `VARINT_TYPE_TAG:PAYLOAD[:VARINT_TYPE_TAG:PAYLOAD] `

    This would result in potentially very long identifiers.

2. Interplanetary Merkle-DAG

    Essentially, a global namespace where every node is content addressed. This is what IPFS does, and is a non-solution: we are in all cases interested in the _most recent, named head_ of a history leading back to a well-known "genesis" node. This requires an external naming system, or a DHT. The latter is called IPNS in the IPFS world, and predictably (and unsurprisingly) times out after 10s of minutes.

3.  Local indexing

    Preserve the project semantics ("just a name for a bunch of branches"), but locally maintain an inverted index `identifier -> project/path` as they are discovered. The local node should be able to answer queries from the network for this mapping. 

    Identifying what is an identifer (hah) needs to resort to convention. For example, a branch prefix like `.rad/identity/` could treat everything below as an indexable identifier, without knowing what kind of identifier. This is bounded by `PATH_MAX` on Linux (which is 4096 chars last time I checked), and probably some incomprehensible behaviour on Windows. That should still be enough, as I can think only of two levels atm: `user`, `user/{feed,..}`.

kim:
URN and URL types as per #65

Closes #65

### DIFF
--- a/librad/src/uri.rs
+++ b/librad/src/uri.rs
@@ -1,0 +1,531 @@
+// This file is part of radicle-link
+// <https://github.com/radicle-dev/radicle-link>
+//
+// Copyright (C) 2019-2020 The Radicle Team <dev@radicle.xyz>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 or
+// later as published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use std::{
+    fmt::{self, Display},
+    ops::Deref,
+    str::{FromStr, Utf8Error},
+};
+
+use multibase::Base;
+use multihash::Multihash;
+use percent_encoding::{percent_decode_str, percent_encode, AsciiSet};
+use regex::RegexSet;
+use thiserror::Error;
+use url::Url;
+
+use crate::peer::{self, PeerId};
+
+/// https://url.spec.whatwg.org/#fragment-percent-encode-set
+const FRAGMENT_PERCENT_ENCODE_SET: &AsciiSet = &percent_encoding::CONTROLS
+    .add(b' ')
+    .add(b'"')
+    .add(b'<')
+    .add(b'>')
+    .add(b'`');
+
+/// https://url.spec.whatwg.org/#path-percent-encode-set
+const PATH_PERCENT_ENCODE_SET: &AsciiSet = &FRAGMENT_PERCENT_ENCODE_SET
+    .add(b'#')
+    .add(b'?')
+    .add(b'{')
+    .add(b'}');
+
+/// Protocol specifier in the context of a [`RadUrn`] or [`RadUrl`]
+///
+/// This pertains to the VCS backend, implying the native wire protocol.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum Protocol {
+    Git,
+    //Pijul,
+}
+
+impl Protocol {
+    /// The "NSS" (namespace-specific string) of the [`Protocol`] in the context
+    /// of a URN
+    pub fn nss(&self) -> &str {
+        match self {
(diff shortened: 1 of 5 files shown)


librad/src/uri.rs
```diff
@@ -0,0 +1,535 @@
...
+        if !self.0.is_empty() {
+            self.0.push('/')
+        }
+        self.0.push_str(&other.0);
+        self
+    }
+
+    pub fn join<S: AsRef<str>>(self, segment: S) -> Result<Self, path::ParseError> {
+        let segment = Self::parse_str(segment)?;
+        Ok(self.append(&Self(segment)))
```
FintanH, line 154: I expected it to join with a separator like `/`. I think this is the usual behaviour of functions called `join` on path-like structures. Is there a reason we don't? Should the functionality be called out in a doc string?

librad/src/uri.rs
```diff
@@ -0,0 +1,535 @@
...
+/// The authority of a rad URL is a [`PeerId`], from which to retrieve the
+/// `radicle-link` repository and branch identified by [`RadUrn`].
+#[derive(Clone, Debug, PartialEq)]
+pub struct RadUrl {
+    authority: PeerId,
+    urn: RadUrn,
+}
+
+impl RadUrl {
+    // TODO: we should be able to open a `RadUrl` from local storage
```
FintanH, line 347: Are we tracking this work in a ticket?

librad/src/uri.rs
kim: It does that, in `append`. I don't actually need `append` anymore, though, so might as well remove it.

librad/src/uri.rs
kim: Ya well, this has been transplanted from #82

FintanH: Right, makes sense :+1:

librad/src/uri.rs
FintanH: Ah, `append` is the confusing part then :joy:

>  I don't actually need append anymore, though, so might as well remove it.

Okay :)

PR #89, opened 2020-04-07
Hashing

### LINKED ISSUE #88: Canonical Hasher
Provide a way to abstract over the hashing algorithm used for digests _not_ depending on the VCS backend.

For compatibility reasons, the canonical hash function shall be Blake2b-256. A faster (but not cryptographically secure) hash function shall be devised for testing purposes (FNV-1 seems fine).

All serialisation shall use [multihash](https://github.com/multiformats/multihash) encoding. Note, however, that the [multihash](https://crates.io/crates/multihash) _crate_ only supports a small subset of the hash algorithms defined in the multicodec table..

kim:
Closes #88

Very simple, but not super satisfying having to implement all hashing in terms of multihash, as this also doesn't give much opportunities to type more parameters of the hashing function.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1067,6 +1067,7 @@ dependencies = [
  "directories",
  "env_logger",
  "failure",
+ "fnv",
  "futures",
  "futures-await-test",
  "futures-timer 3.0.2",
--- a/librad/Cargo.toml
+++ b/librad/Cargo.toml
@@ -12,6 +12,7 @@ bs58 = "0.3"
 bytes = "0.5"
 directories = "2.0"
 failure = { version = "0.1", features = ["backtrace"] }
+fnv = "1.0"
 futures = { version = "0.3", features = ["thread-pool"] }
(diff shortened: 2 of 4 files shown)


massimiliano-mantione:
For me it makes sense.

I still don't fully understand all the implications of handling *several* hashing functions at the same time, like when transitioning from one to another.
Ideally it should be enough to change the default function so that all subsequent hashes will be computed with the new one.
However existing documents (and URNs) will still use the old hash, and URNs are supposed to be immutable...

kim:
I think it makes more sense now: there is one `Hash` type, and we just happen to use `multihash` for serde. This way, we have control over which hash functions we support, and how to evolve them.

It does show, however, one of my many complaints about `serde`: it does not seem like there is a way to control encoding for different output formats. I'd prefer to have the JSON format use a `multibase`-encoded _string_, while the CBOR format uses a byte array -- what it does now is to use an array of numbers for JSON.

PR #105, opened 2020-04-17
Make RadUrn identified by Hash instead of Multihash

kim:
Note that `Hash` is not forward-compatible, ie. an earlier version of the software will refuse to handle a hash algorithm introduced later. Perhaps this is what @massimiliano-mantione's doubts were. 

I don't have a good solution for how to solve this, since we assume that we must stay compatible with `registry`. I am, otoh, also not sure where this requirement came from, respectively what hash the `registry` wants to verify, and based on which input data.

### DIFF
--- a/librad/src/hash.rs
+++ b/librad/src/hash.rs
@@ -15,6 +15,13 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+use std::{
+    convert::TryFrom,
+    fmt::{self, Display},
+    str::FromStr,
+};
+
+use multibase::Base;
 use multihash::{Blake2b256, Multihash};
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use thiserror::Error;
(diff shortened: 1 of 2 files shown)


geigerzaehler:
I’m also not aware of any requirements on the hash from the registry. The registry considers data from the registry opaque at the moment and there are now plans to change this. We just provide some blob storage for projects and the app determines how to encode coco data and store it.

FintanH:
> I don't have a good solution for how to solve this

```Error: please upgrade `librad`, thank you.```

cloudhead:
In the future, we will want to consolidate the TUF verification such that it can be performed both in the registry, and offline, in `radicle-link`. This will ensure that similar functionality (multi-sig project metadata updates) isn't duplicated with different semantics, and will also allow the registry to operate as a trusted project metadata/identity verifier. For this to work though, the registry needs to be able to run the same code as radicle-link, namely the hashing function and signature verification.

This will come about when we start allowing users to anchor project metadata state on the registry.

kim:
@cloudhead Well, for it to work, the entire history of an identity branch must be stored on-chain. Or, a full-node acts as a radicle-link node, in which case the actual verification is not an on-chain computation.

Anyways, I'll leave it like this for now, but add a follow-up ticket to accept _any_ multihash when deserialising.

Issue #102, opened 2020-04-15
Lament about serde


kim:
This is a kind of meta-issue about the problems I found when working with `serde`. It's not urgent, but we should find ways to improve the situation. It is also specific to us (I think), and not really applicable to `decisions`, because we are the only ones to deal with different serialisation formats.

## The Good

`serde` is ubiquitous in Rust. Many times, it is really convenient to use, as all that's needed is to slap a `#[derive]` and possibly some field attributes. Also, a lot of crates come with `serde` impls (behind a feature or not), so there is really not much to think about. 

## The Bad

When you _do_ have to write your own `Serialize`/`Deserialize` (because of coherence or whatever), things get rather convoluted. But you know, it's Rust, get over it.

## The Ugly

The problem with `serde` is that it tries to abstract over _any_ serialisation format, and that is inevitably leaky. The concrete issues I found are:

* There is no obvious way to define `Serialize`/`Deserialize` in a way that _makes sense_ for more than one format. Case in point is #89: JSON doesn't like binary data, so one typically stores base64-encoded strings. CBOR, otoh, **does** like binary data. CBOR tags (useful e.g. for bignums and datetimes) currently require a newtype.
* The data model basically exposes _strings_ to denote constructors, fields, and enum variants. That's how JSON works, but not very compact. In fact, our CBOR wire types carry _more_ metadata than payload. Since there is apparently no way to annotate or otherwise infer field indices, the method to use those as per [RFC 7049, Section 3.7](https://tools.ietf.org/html/rfc7049#section-3.7) is not supported by the available library. This is also the reason there is no `serde` backend for protocol buffers.
* There does not seem to be any form of streaming parsing, e.g. for CBOR indefinite length arrays 
* There does not seem to be a way to generate IDL documents (swagger or whatever) from annotated types, which I find a bit surprising. There's two crates promising to allow [CDDL](https://tools.ietf.org/html/rfc8610) parsing and validation, but not sure how well they work.

## What now?

Treating `serde` as an object-to-JSON-mapper is probably fine for now. I'd expect the amount of handwritten impls to grow over time anyways, but auto deriving is very convenient for now. The CBOR story is very sad imho. I'd be leaning towards [minicbor](https://crates.io/crates/minicbor) for the wire types, although that comes at a lot of handwritten definitions and still doesn't solve the IDL problem (although perhaps not that important there).

FintanH:
[...]

PR #106, opened 2020-04-17
Stateful peer

### LINKED ISSUE #92: Expose stateful datatype like in mesh example
All the wiring, options, callbacks

kim:
Closes #92

### DIFF
--- a/librad/src/git/repo.rs
+++ b/librad/src/git/repo.rs
@@ -1,0 +1,624 @@
+// This file is part of radicle-link
+// <https://github.com/radicle-dev/radicle-link>
+//
+// Copyright (C) 2019-2020 The Radicle Team <dev@radicle.xyz>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 or
+// later as published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use std::{
+    collections::{BTreeMap, HashMap, HashSet},
+    fs,
+};
+
+use serde::{de::DeserializeOwned, Serialize};
+use thiserror::Error;
+
+use crate::{
+    canonical::{Cjson, CjsonError},
+    git::{
+        refs::{self, Oid, Refs},
+        url::GitUrlRef,
+    },
+    hash::{self, Hash},
+    keys::device,
+    meta::entity::{
+        self,
+        data::{EntityBuilder, EntityData},
+        Entity,
+        Signatory,
+        VerificationStatus,
+    },
+    paths::Paths,
+    peer::PeerId,
+    uri::{self, RadUrl, RadUrn},
+};
+
+const RAD_REFS: &str = "rad/refs";
+
+/// A git repository with `radicle-link` specific operations
+pub struct Repo {
+    urn: RadUrn,
+    key: device::Key,
+    repo: git2::Repository,
+}
+
+#[allow(clippy::large_enum_variant)]
+#[derive(Debug, Error)]
+#[non_exhaustive]
+pub enum Error {
+    #[error("Repo already exists")]
+    AlreadyExists,
+
+    #[error("Unknown repo")]
+    NoSuchRepo,
+
+    #[error("Branch {0} not found")]
+    NoSuchBranch(String),
+
+    #[error("Blob {0} not found")]
+    NoSuchBlob(String),
+
+    #[error("Identity VerificationStatus must be {expected:?}, got {actual:?}")]
+    Verification {
+        expected: VerificationStatus,
+        actual: VerificationStatus,
+    },
+
+    #[error(
+        "Identity root hash doesn't match resolved URL. Expected {expected}, actual: {actual}"
+    )]
+    RootHashMismatch { expected: Hash, actual: Hash },
+
+    #[error(transparent)]
+    HashAlgorithm(#[from] hash::AlgorithmMismatch),
+
+    #[error(transparent)]
+    Entity(#[from] entity::Error),
+
+    #[error(transparent)]
+    Refsig(#[from] refs::signed::Error),
+
+    #[error(transparent)]
+    Json(#[from] serde_json::error::Error),
+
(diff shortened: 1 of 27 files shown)


librad/src/net/peer.rs
```diff
@@ -0,0 +1,195 @@
...
+}
+
+impl Serialize for Rev {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        match self {
+            Self::Git(oid) => {
+                let mut seq = vec![0];
```
FintanH, line 49: Could you leave a comment saying that this is a tag for the variant? Just to make sure it's called out and not something "magic" :)

librad/src/net/peer.rs
```diff
@@ -0,0 +1,195 @@
...
+                _ => Err(serde::de::Error::custom("Unknown tag")),
+            },
+
+            None => Err(serde::de::Error::custom("No data")),
+        }
+    }
+}
+
+impl Into<uri::Protocol> for Rev {
+    fn into(self) -> uri::Protocol {
```
FintanH, line 77: I think this would be better as a method on `Rev` that borrows instead of owns. It would avoid cloning the data inside just to map it to the protocol.

librad/src/net/peer.rs
```diff
@@ -0,0 +1,195 @@
...
+        assert_eq!(
+            rev,
+            serde_cbor::from_slice(&serde_cbor::to_vec(&rev).unwrap()).unwrap()
+        )
+    }
+
+    #[test]
+    fn test_gossip_serde() {
+        let rev = Rev::Git(git2::Oid::hash_object(git2::ObjectType::Commit, b"chrzbrr").unwrap());
+        let gossip = Gossip::new(Hash::hash(b"cerveza coronita"), Path::new(), rev);
```
FintanH, line 189: lol

librad/src/net/peer.rs
kim: Yes, but no. This is actually the standard CBOR encoding for a tuple, so instead of writing a letter to future self whenever we have to convince `serde` to output proper CBOR, we should rather fix the abstraction leak (hint, hint #102)

librad/src/net/peer.rs
kim: Good catch. Source of idiomatic polizei is tell me it need programme like this deed82ec9bc243eae29042030690e3e39c0a0636

librad/src/net/peer.rs
FintanH: Heh, fair enough :smile:

librad/src/net/peer.rs
FintanH: :police_officer:

librad/src/git/repo.rs
```diff
@@ -0,0 +1,589 @@
...
+
+    #[error(transparent)]
+    Cjson(#[from] CjsonError),
+
+    #[error(transparent)]
+    Git(#[from] git2::Error),
+}
+
+impl Repo {
+    pub fn create<T>(paths: &Paths, key: device::Key, meta: &Entity<T>) -> Result<Self, Error>
```
FintanH, line 101: To check I'm understanding correctly: this allows you to create a repo for any entity, whether it's a project or a user? And this becomes tracking information in the current project you're working on?

librad/src/uri.rs
```diff
@@ -519,14 +521,41 @@ impl<'de> Deserialize<'de> for RadUrl {
     }
 }
 
+pub struct RadUrlRef<'a> {
+    pub authority: &'a PeerId,
+    pub urn: &'a RadUrn,
+}
+
+impl<'a> RadUrlRef<'a> {
+    pub fn to_owned(&self) -> RadUrl {
```
FintanH, line 530: You could probably add a https://doc.rust-lang.org/std/borrow/trait.ToOwned.html impl for this.

librad/src/git/repo.rs
kim: Yup. As I said in RTC, I'm not fully convinced that knowning `T` for `clone` is always practical -- the caller may have just a URN, so it might not be clear from the context what `T` should be. We can fix that by having a value-level bit in the URN which tells us what `T` is.

librad/src/uri.rs
kim: I tried that, but couldn't get it to typecheck. I might be missing something, but the issue seems to be: `ToOwned` has a constraint that the associated type `Owned` must be `Borrow`. Now, `Borrow` is parametrised over the borrowed type, and `fn borrow(&self) -> &Borrowed`. There is no way I can construct a `Borrowed` and then lend it out -- `borrow` owns it.

Since I see this `*Ref` pattern with inherent methods being used all over the place in Rust, I figure this _is_ actually an issue. Or I'm not interrigent enough, but either way I lost patience after a while :grimacing:

kim:
Unfortunately, the tests fail consistently when run inside a docker container, but succeed consistently using a real network stack. Will need to investigate separately.

librad/src/uri.rs
FintanH: Fair enough, I think I'd have to give it a go myself to understand better. But we shouldn't waste both of our time :smile:

librad/src/git/repo.rs
```diff
@@ -160,13 +166,21 @@ impl Repo {
...
         })
     }
 
-    pub fn clone<T>(paths: &Paths, key: device::Key, url: &RadUrl) -> Result<Self, Error>
+    /// Clone a [`Repo`] from the given [`RadUrl`]
+    ///
+    /// The repo must not already exist.
+    ///
+    /// TODO: entity verification is not currently functional, pending #95
+    pub fn clone<T>(paths: &Paths, key: device::Key, url: RadUrl) -> Result<Self, Error>
```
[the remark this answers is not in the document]
kim: Ya, `clone()` is a trait method, whereas this `clone` is an inherent -- so it's not actually overloaded. 

I'm also not keen on using anything but git terminology here, because, you know, the module is called `git::repo`. `replicate` would also not be totally accurate, it's more like `like_git_clone_but_with_some_special_rules_see_the_spec_draft1_rev69_section_42`.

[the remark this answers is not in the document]
FintanH: Fair enough :smile:

kim:
It turns out the docker issue was simply that the build image didn't have `git` installed, which we need for `git upload-pack`. This error would have surfaced when running a peer standalone, but not in the tests because the peers are `tokio::task::spawn`ed. I put some logging in place which will print `error` logs next to the test, and also makes running `RUST_LOG=whatever cargo test` actually do what you'd expect (but only in the integration tests for now).

PR #111, opened 2020-04-21
Factor out protocol upgrade

kim:
Also:

* introduce a simple mock stream and add a basic test for upgrades
* abstract over `SocketAddr` in gossip, to pave the way for testing
  using the mock stream

### DIFF
--- a/librad/src/net/gossip.rs
+++ b/librad/src/net/gossip.rs
@@ -24,9 +24,9 @@
 use std::{
     collections::{HashMap, HashSet},
     fmt::Debug,
+    hash::Hash,
     iter,
     marker::PhantomData,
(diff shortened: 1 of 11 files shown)


librad/src/net/gossip.rs
```diff
@@ -279,18 +310,28 @@ impl<S: Clone, A: Clone, R, W> Clone for Protocol<S, A, R, W> {
...
+    for<'de> Addr: Serialize
+        + Deserialize<'de>
+        + Clone
+        + Debug
+        + Hash
+        + PartialEq
+        + Eq
+        + Send
+        + Sync
+        + 'static,
```
FintanH, line 326: :clap:

PR #114, opened 2020-04-29
Metadata entities

massimiliano-mantione:
Ported existing code to use new `Entity`-based metadata types.


librad/src/git.rs
```diff
@@ -286,12 +287,15 @@ pub mod project {
             paths: &Paths,
             sources: &git2::Repository,
         ) -> Result<ProjectId, Error> {
-            let mut meta = meta::Project::new(&self.name, &PeerId::from(self.key.clone()));
-            meta.default_branch = self.default_branch;
-            meta.description = self.description;
-            meta.rel = self.rel;
-
-            GitProject::init(paths, &self.key, sources, meta, self.founder.clone())
+            // FIXME[ENTITY]: add certifier instead of peer
```
kim, line 290: I think this will resolve itself with #106: the repo init will just take an `Entity`.

librad/src/git.rs
```diff
@@ -149,16 +152,14 @@ impl GitProject {
...
             .find_branch(&default_branch, git2::BranchType::Local)
-            .map_err(|e| Error::MissingDefaultBranch(default_branch.clone(), e))?
+            .map_err(|e| Error::MissingDefaultBranch(metadata.default_branch().to_owned(), e))?
             .into_reference()
             .peel_to_commit()?;
 
-        // Ensure the signing key is a maintainer
-        let mut metadata = metadata;
-        metadata.add_maintainer(&PeerId::from(key.clone()));
+        // FIXME[ENTITY]: Verify entity signatures
```
kim, line 162: Same: we will just init a project from any valid `Entity`

rad2/src/commands/profiles.rs
```diff
@@ -115,6 +115,7 @@ impl Commands {
     }
 }
 
+// FIXME[ENTITY] resolve a User URN and return the User, not just a profile
```
kim, line 118: Yeah I think profile is obsolete anyway: you're supposed to have exactly one `User` unless you're using multiple keypairs, so we can just store the URN in some config.

librad/src/meta/profile.rs
```diff
@@ -64,23 +64,93 @@ pub enum ProfileImage {
     Url(Url),
 }
 
-#[derive(Deserialize, Serialize, Debug, Clone, PartialEq)]
+#[derive(Deserialize, Serialize, Debug, Clone, Copy, PartialEq, Eq)]
+pub struct GeoArc {
```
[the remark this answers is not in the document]
FintanH: I was looking at this resource: https://astro.unl.edu/naap/motion1/tc_units.html

+/- could mean North and South or East and West. Something I found interesting is that the degrees max for N/S is 90 and for E/W is 180. 

It might make sense to have two separate types here one for latitude and one for longitude, `deg`, `min`, and `sec` could be all be `u8`, and we carry an `enum NS` or `enum EW` indicating the direction. All the while ensuring the invariants of those values above via smart constructors.

librad/src/meta/user.rs
```diff
@@ -0,0 +1,215 @@
...
+    }
+}
+
+pub type UserData = EntityData<UserInfo>;
+
+impl UserData {
+    fn profile_ref_mut(&mut self) -> Result<&mut ProfileRef, Error> {
+        match &mut self.info.profile {
+            Some(p) => Ok(p),
+            None => Err(Error::MissingRootHash),
```
FintanH, line 63: I don't understand this error in relation to this function :thinking: Same goes for below in `user_profile_mut`.

What's the `RootHash` have to do with the `ProfileRef` not being set?

librad/src/git.rs
```diff
@@ -149,16 +152,14 @@ impl GitProject {
         let mut pgp_key = key.clone().into_pgp(&nickname, fullname)?;
 
         // Link all metadata to the tip of the default branch
-        let default_branch = metadata.default_branch.clone();
+        let default_branch = metadata.default_branch().to_owned();
```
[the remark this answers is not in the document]
massimiliano-mantione: Yes, but not too important.

I always use `to_owned` to say "that is what I wanted but I need to own it".
I reserve `clone` to "I am duplicating this thing because I will mutate the result to turn it into something else", or similar uses.

But this is just personal style and I can adjust to different styles.

librad/src/meta/user.rs
massimiliano-mantione: You are right, it makes no sense at all!
Sorry, this was part of a "FIXME" that got lost :-/
I was using a placeholder error to make it compile and I never implemented the real error.
Now I pushed a "BuilderError" for this case.

librad/src/git.rs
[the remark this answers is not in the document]
FintanH: I haven't grown an opinion about it yet :sweat_smile: Thought you could shed some light because the documentation of `ToOwned` just says it's "more general than clone", which makes no sense to me.

PR #125, opened 2020-05-08
Parallel Build Checks

### LINKED ISSUE #120: CI: run clippy as concurrent step
`clippy` is getting very annoying due to long recompilation times. Also if it fails, the test suite will not be run, so any issues there will require another build cycle to fix.

Instead, we should just occupy two build agent slots and run `clippy` concurrently to the other jobs.

FintanH:
Fixes #120 

I split the commands out into multiple files. Each one of these, I think, can be run in parallel.
I added a `- wait` for the build and test step because I think this might make sense. We mostly care about building and testing initially, the rest can be done during a cleanup. I think it would also mean we have the build cache set up from the build and test too.

### DIFF
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -1,17 +1,28 @@
 steps:
-  - label: "Build, test, lint, clip, doc"
+  - label: "Build, test, doc"
     commands:
       - "ci/build-test"
       - wait
-      - "ci/advisory"
-      - "ci/clippy"
       - "ci/docs"
-    agents:
+    agents: &build-agent
       production: "true"
       platform: "linux"
-    env:
+    env: &build-docker
       DOCKER_IMAGE: "gcr.io/opensourcecoin/radicle-link-seedling-build@sha256:f53c35f65541d91b7665003a6362be48d6b2687a5a9e46adaadccaf3129da51f"
       DOCKER_FILE: .buildkite/docker/rust-nightly/Dockerfile
+
+  - label: "Lint and clip"
+    commands:
+      - "ci/clippy"
+    agents: *build-agent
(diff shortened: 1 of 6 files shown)


kim:
This doesn't (yet) run the tasks concurrently -- each of them would need to be it's own `step`. I think the `cargo deny` stuff can indeed be run concurrently, because it doesn't do any compilation. `cargo doc` does, afaik, so I'd be careful -- the problem is, as soon as there is any dependency downloading and/or building involved, each of the concurrent `step`s has to populate an independent cache, which may eat the build time speedup otherwise gained by concurrency. Since `clippy` recompiles _anyways_, my thinking was that the build time will overall be reduced, plus we get to see the test results even with liberal formatting.

FintanH:
> This doesn't (yet) run the tasks concurrently -- each of them would need to be it's own step. 

Ah, the documentation wasn't so clear around `commands`. But why would you have a `wait` command then if they can't be run concurrently? That's why I thought they would build at the same time.

So, I suppose we could do 3 steps then: Build & Test, Advisory, and Docs & Clippy? It's right about now that I wish we had `dhall-buildkite` :eyes:

.buildkite/pipeline.yaml
```diff
@@ -1,14 +1,29 @@
 steps:
-  - label: "Build, lint, test"
+  - label: "Build, test, doc"
     commands:
-      - "ci/run"
-    agents:
+      - "ci/build-test"
+      - wait
```
kim, line 5: This is not the [wait step](https://buildkite.com/docs/pipelines/wait-step), but the bash builtin, which waits for child processes.

FintanH: :facepalm:

Issue #121, opened 2020-05-07
Type witnesses for entity verification status


kim:
It should be possible to assert on the type level that an `Entity` value has the desired (success) `VerificationStatus`. For example, initialising a repo from an entity should take a type which witnesses that it does not have a parent (ie. is at revision 1) and is signed by the local key, instead of checking these invariants at runtime.

As one does not simply do type witnesses in Rust, this will probably be newtype wrappers, either returned from the corresponding verification methods, or with "smart" constructors enforcing exactly the preconditions we're after.

massimiliano-mantione:
Fixed by #147 (which includes #148).

PR #126, opened 2020-05-08
Verify & Testify

### LINKED ISSUE #119: Entity: VerificationStatus::Verified is never set
Its meaning is that `Entity::compute_history_status` did not yield an error, and so should be set after the history-walking loop exits.

FintanH:
Fixes #119 

Added `Verified` when we are exiting the loop for `compute_history_status`. This needed to change the test code slightly, so as not to `clone` the `user` so we could mutate it.

Also added some refactoring code as I was passing through.

### DIFF
--- a/librad/src/meta/entity.rs
+++ b/librad/src/meta/entity.rs
@@ -142,46 +142,18 @@ pub enum VerificationStatus {
 }
 
 impl VerificationStatus {
-    pub fn verified(&self) -> bool {
-        if let VerificationStatus::Verified = self {
-            true
+    pub fn verification_failed(&self) -> Option<&Error> {
+        if let VerificationStatus::VerificationFailed(err) = self {
+            Some(err)
         } else {
-            false
+            None
         }
     }
-    pub fn signed(&self) -> bool {
-        if let VerificationStatus::Signed = self {
-            true
+    pub fn history_verification_failed(&self) -> Option<&HistoryVerificationError> {
+        if let VerificationStatus::HistoryVerificationFailed(err) = self {
+            Some(err)
         } else {
-            false
-        }
-    }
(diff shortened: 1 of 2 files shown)


librad/src/meta/entity.rs
```diff
@@ -647,6 +619,7 @@ where
 
             // End at root revision
             if revision == 1 {
+                self.status = VerificationStatus::Verified;
```
FintanH, line 622: So I set the `status` here. But in the test suite, it comes out as `Unknown` :thinking: @massimiliano-mantione: maybe you have an idea of what's going on.

FintanH: Ah it's because of the `clone` here: https://github.com/radicle-dev/radicle-link/pull/126/files#diff-e8128eef3fbc5e6398034e0c89623737R113 :facepalm:

librad/src/meta/entity.rs
```diff
@@ -142,46 +142,18 @@ pub enum VerificationStatus {
 }
 
 impl VerificationStatus {
-    pub fn verified(&self) -> bool {
-        if let VerificationStatus::Verified = self {
-            true
+    pub fn verification_failed(&self) -> Option<&Error> {
```
kim, line 145: Could we preserve the success cases as well (a la `is_ok()`)? Perhaps `VerificationStatus` could actually be `Result<SuccessCases, ErrorCases>`

FintanH: Well, my aim here was to use the success case directly when someone wants to compare or match on. I think https://github.com/radicle-dev/radicle-link/issues/121 could make any of this obsolete too. I imagine we'd have `Result<Entity<Signed, T>, Error>`. Does that sound right to you?

PR #135, opened 2020-05-11
Optimise serde of `Hash` for JSON

### LINKED ISSUE #116: Hash: optimise serde for JSON
Currently, it attempts to store itself as a sequence of bytes, which is good for CBOR, but less good for JSON. 

Instead, store a `multibase`-encoded `multihash` as a JSON string. As we already know, `serde` is inappropriate for our wire encoding, so we will need to move to a better serialisation library there (#102)

kim:
That is, go through multibase(multihash(hash)) string instead of byte array.

Fixes #116

### DIFF
--- a/librad/src/hash.rs
+++ b/librad/src/hash.rs
@@ -23,7 +23,7 @@ use std::{
 
 use multibase::Base;
 use multihash::{Blake2b256, Multihash};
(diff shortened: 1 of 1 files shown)


kim:
Closed via d358e861b7c305723832928bc248a40351b8d201

Issue #115, opened 2020-05-07
EntityData: typed setters


kim:
`EntityData` is supposed to be a "dumb" data container, where validation occurs only when converting to an `Entity`, which is immutable. That is fine, but the implementation turns out to yield a quite impractical API: all fields are of type `String`, while the library user will have appropriately typed values at hand. That is, a serialisation roundtrip is unnecessarily enforced.

Instead, `EntityData` should be strongly typed. `Entity` can then just consume `EntityData`, effectively making it immutable.

Note that this will require appropriately defined `serde` impls for `Hash` and `PublicKey` in the context of JSON.

FintanH:
So I presume this is regarding thing like the [`hash`](https://github.com/radicle-dev/radicle-link/blob/master/librad/src/meta/entity/data.rs#L63-L64) family of fields and [`EntitySignatureData`](https://github.com/radicle-dev/radicle-link/blob/master/librad/src/meta/entity/data.rs#L71)?

But are you also advocating we provide types for [`revision`](https://github.com/radicle-dev/radicle-link/blob/master/librad/src/meta/entity/data.rs#L62) and what not? If so, I feel like we're moving towards validation functions and then compute the hash of the result of them.

kim:
We have:

```rust
pub struct Entity<T> {
    status: VerificationStatus,
    name: String,
    revision: u64,
    rad_version: u8,
    hash: Hash,
    root_hash: Hash,
    parent_hash: Option<Hash>,
    signatures: HashMap<PublicKey, EntitySignature>,
    keys: HashSet<PublicKey>,
    certifiers: HashSet<RadUrn>,
    info: T,
}

pub struct EntityData<T> {
    pub rad_version: u8,
    pub name: Option<String>,
    pub revision: Option<u64>,
    pub hash: Option<String>,
    pub root_hash: Option<String>,
    pub parent_hash: Option<String>,
    pub signatures: Option<HashMap<String, EntitySignatureData>>,
    pub keys: HashSet<String>,
    pub certifiers: HashSet<String>,
    pub info: T,
}
```

Where `Entity::to_builder` gives you `EntityData`. I totally agree with the pattern that there is a mutable type, which gets "frozen" and "thawed" through an immutable one. What I don't find practical is that we loose (almost) all type information along the way.

My proposal is something like:

```rust
#[derive(Serialize, Deserialize)]
pub struct Entity<T> {
    name: String,
    revision: u64,
    rad_version: u8,
[...]

PR #137, opened 2020-05-11
Remove obsolete code

### LINKED ISSUE #118: Remove obsolete code
After #106 goes in, a large chunk of code can be removed:

* [x] mesh example
* [x] the `project` module
* [x] everything related to GPG

Note that this will also break the CLI. As it is for demo purposes only, I suggest to remove it entirely for now, and resurrect it as needs arise and time permits.

kim:
Closes #118


librad/src/git/repo.rs
```diff
@@ -422,6 +426,14 @@ impl Repo {
         })
     }
 
+    /// Obtain a [radicle-surf] `Browser` for this repo, consuming self.
+    ///
+    /// [radicle-surf]: https://github.com/radicle-dev/radicle-surf
+    pub fn browser(self) -> Result<surf::Browser, Error> {
+        let bro = surf::Browser::new(self.repo.into())?;
```
FintanH, line 433: `bro` and `Ok(bro)` - amazing :clap: :smile:

librad/src/keys.rs
```diff
@@ -15,5 +15,274 @@
...
+
+    pub fn public(&self) -> PublicKey {
+        PublicKey(self.0.public_key())
+    }
+
+    pub fn sign(&self, data: &[u8]) -> Signature {
+        Signature(ed25519::sign_detached(data, &self.0))
+    }
+
+    const PKCS_ED25519_OID: &'static [u64] = &[1, 3, 101, 112];
```
FintanH, line 71: What's this? :thinking:

librad/src/keys.rs
```diff
@@ -15,5 +15,274 @@
...
+    }
+}
+
+impl AsRef<[u8]> for SecretKey {
+    fn as_ref(&self) -> &[u8] {
+        self.0.as_ref()
+    }
+}
+
+#[derive(Debug)]
```
FintanH, line 114: Would it be useful to have an `Error` instance?

librad/src/keys.rs
```diff
@@ -15,5 +15,274 @@
...
+        let bytes = hex::decode(s)?;
+        let buffer = if bytes.len() == 64 {
+            let mut buffer = [0u8; 64];
+            for (i, v) in bytes.iter().enumerate() {
+                buffer[i] = *v;
+            }
+            buffer
+        } else {
+            return Err(signature::DecodeError::InvalidLength {
+                actual: bytes.len(),
```
FintanH, line 231: minor suggestion: Have a `let actual = bytes.len();` above to reuse the result in the `if` and here :)

librad/src/keys.rs
```diff
@@ -15,5 +15,274 @@
...
+    }
+
+    pub fn from_bs58(s: &str) -> Option<Self> {
+        let bytes = match bs58::decode(s.as_bytes())
+            .with_alphabet(bs58::alphabet::BITCOIN)
+            .into_vec()
+        {
+            Ok(v) => v,
+            Err(_) => return None,
+        };
```
FintanH, line 244: The indentation threw me off here. I think you could do:
```
let bytes = bs58::decode(s.as_bytes())
     .with_alphabet(bs58::alphabet::BITCOIN)
     .into_vec()
     .ok()?;
```

.buildkite/docker/rust/Dockerfile
```diff
@@ -1,25 +1,46 @@
...
+    apt-get autoremove; \
+    rm -rf /var/lib/apt/lists/*
 
-RUN rustup component add rustfmt \
-    && cargo install sccache cargo-deb \
-    && rm -rf /usr/local/cargo/registry \
-    && rm /usr/local/cargo/.package-cache
+# Rust toolchain
+# Make sure this is in sync with rust-toolchain!
+ENV RUST_VERSION=nightly-2020-05-11 \
```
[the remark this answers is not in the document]
kim: Ya maybe, but I think `COPY` doesn't follow symlinks, and can't escape the "context dir" (ie. this directory).

I'm contemplating a top-level [multistage](https://docs.docker.com/develop/develop-images/multistage-build/) `Dockerfile`, but need to figure out how to teach CI to only use the first stage.

librad/src/keys.rs
kim: It's a magical value encoding the signature algorithm in a PKCS-encoded certificate. I cannot document this, I'm sorry, it is only accessible to initiated members of the free masons guild "Templars of Londres Shacklewell" (TLS).

librad/src/keys.rs
kim: I did not write this. But it should also be obsolete soon (#134 + #121 + #115)

librad/src/keys.rs
kim: This method is obsolete

PR #147, opened 2020-05-15
Entity status with type witnesses

massimiliano-mantione:
This is a first step towards the goal.

The feature is implemented but there are still the following issues:

* The code only uses the "unknown" state: verification calls would need a proper resolver so i would do this in another pull request, after we have resolvers in place.
* I would add some type aliases like `DraftUser`, `SignedUser` and `VerifiedUser` but I wanted to agree on naming first.
* A couple of clippy issues with `new` method names.
* The API does not consume the unverified entity, it still sets its status but it also produces a new (cloned) entity of the proper type; we should talk more about how error handling would work with a consuming API.

### DIFF
--- a/librad/src/meta/entity.rs
+++ b/librad/src/meta/entity.rs
@@ -33,7 +33,6 @@ use serde::{
 use std::{
     collections::{HashMap, HashSet},
     convert::{Into, TryFrom},
-    iter::FromIterator,
     marker::PhantomData,
     str::FromStr,
 };
--- a/librad/src/meta/entity.rs
+++ b/librad/src/meta/entity.rs
@@ -456,11 +456,11 @@ where
     /// - the entity has not been already signed using this same key
     /// - this key is allowed to sign the entity (using `check_key`)
     pub async fn sign(
-        mut self,
+        &mut self,
         key: &SecretKey,
         by: &Signatory,
         resolver: &impl Resolver<User<Draft>>,
-    ) -> Result<Entity<T, Signed>, Error> {
+    ) -> Result<(), Error> {
         let public_key = key.public();
         if self.signatures().contains_key(&public_key) {
             return Err(Error::SignatureAlreadyPresent(public_key.to_owned()));
@@ -471,7 +471,7 @@ where
             sig: self.compute_signature(key)?,
         };
         self.signatures.insert(public_key, signature);
-        Ok(self.with_status::<Signed>())
+        Ok(())
     }
 
     /// Check that an entity signature is valid
@@ -522,8 +522,8 @@ where
     where
         ST: Clone,
     {
-        let mut keys = HashSet::<PublicKey>::from_iter(self.keys().iter().cloned());
-        let mut users = HashSet::<RadUrn>::from_iter(self.certifiers().iter().cloned());
+        let mut keys = self.keys().iter().cloned().collect::<HashSet<_>>();
+        let mut users = self.certifiers().iter().cloned().collect::<HashSet<_>>();
 
         if self.revision == 1 && (self.parent_hash.is_some() || self.root_hash != self.hash) {
             // TODO: define a better error if `self.parent_hash.is_some()`
--- a/librad/src/meta/entity.rs
+++ b/librad/src/meta/entity.rs
@@ -132,10 +132,15 @@ pub enum HistoryVerificationError {
     },
 }
 
+/// Type witness for a fully verified [`Entity`].
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Verified;
+
+/// Type witness for a signed [`Entity`].
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Signed;
+
+/// Type witness for a draft [`Entity`].
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Draft;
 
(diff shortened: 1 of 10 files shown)


librad/src/meta/entity.rs
```diff
@@ -141,6 +141,55 @@ pub enum VerificationStatus {
     Unknown,
 }
 
+pub trait VerificationStatusMarker {
+    fn status() -> VerificationStatus;
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct EntityStatusVerified {
```
FintanH, line 149: This can be defined as:
```rust
pub struct EntityStatusVerified {}
```

librad/src/meta/entity.rs
```diff
@@ -141,6 +141,55 @@ pub enum VerificationStatus {
...
+    marker: (),
+}
+impl EntityStatusVerified {
+    pub(self) fn new() -> Self {
```
FintanH, line 153: Is `pub(self)` legitimate syntax? :thinking: I've never seen that before :o

librad/src/meta/entity.rs
```diff
@@ -141,6 +141,55 @@ pub enum VerificationStatus {
...
+    }
+}
+impl VerificationStatusMarker for EntityStatusVerified {
+    fn status() -> VerificationStatus {
+        VerificationStatus::Verified
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct EntityStatusSigned {
```
FintanH, line 164: Same as `EntityStatusVerified` above

librad/src/meta/entity.rs
```diff
@@ -141,6 +141,55 @@ pub enum VerificationStatus {
...
+    }
+}
+impl VerificationStatusMarker for EntityStatusSigned {
+    fn status() -> VerificationStatus {
+        VerificationStatus::Signed
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct EntityStatusUnknown {
```
FintanH, line 179: Ditto

librad/src/meta/entity.rs
```diff
@@ -205,9 +254,11 @@ pub trait Resolver<T> {
 ///   and certifiers, to prove that the entity evolution is actually under the
 ///   control of its current "owners" (the idea is taken from [TUF](https://theupdateframework.io/)).
 #[derive(Clone, Debug, PartialEq, Eq)]
-pub struct Entity<T> {
+pub struct Entity<T, ST> {
     /// Entity verification status
     status: VerificationStatus,
+    /// Verification status marker type
+    status_marker: ST,
```
FintanH, line 261: So I was expecting to see this as `_status_marker: PhantomData<ST>`. The above status types would then be types that don't have constructors, i.e.:
```rust
pub struct Unknown;
pub struct Signed;
pub struct Verified;
```

Then the return type of `check_signatures` for `Entity` would be: `Entity<T, Signed>`

Does this make sense to you? Maybe this wouldn't work for some reason, but I can't tell without following the types myself.

kim:
Will take a closer look later, but from what I see it’s coming along nicely!

Be careful with `master` churn, though, some of the changed files no longer exist!

kim:
I think this is pretty cool already -- do we want to wait until we have all the pieces for verification together, or do a merge-fest now and integrate after?

librad/src/meta/entity.rs
```diff
@@ -205,9 +191,11 @@ pub trait Resolver<T> {
 ///   and certifiers, to prove that the entity evolution is actually under the
 ///   control of its current "owners" (the idea is taken from [TUF](https://theupdateframework.io/)).
 #[derive(Clone, Debug, PartialEq, Eq)]
-pub struct Entity<T> {
-    /// Entity verification status
-    status: VerificationStatus,
+pub struct Entity<T, ST> {
+    // /// Entity verification status
+    // status: VerificationStatus,
```
FintanH, line 196: Ah, I forgot to delete this in my PR. Could clean it up now :)

librad/src/meta/entity.rs
```diff
@@ -582,87 +593,81 @@ where
...
-        certifier_resolver: &impl Resolver<User>,
-    ) -> Result<(), HistoryVerificationError> {
+    pub async fn check_history_status(
+        self,
+        resolver: &impl Resolver<Entity<T, ST>>,
+        certifier_resolver: &impl Resolver<User<Draft>>,
+    ) -> Result<Entity<T, Verified>, HistoryVerificationError>
+    where
+        ST: Clone,
+        ST: std::fmt::Debug,
```
FintanH, line 610: I added these `Debug` bounds for debugging. Can remove them now :)

librad/src/meta/entity_test.rs
```diff
@@ -325,16 +327,16 @@ async fn test_project_update() {
...
         .last_mut()
         .unwrap()
         .sign(&K1, &Signatory::OwnedKey, &EMPTY_RESOLVER)
         .await
         .unwrap();
-    assert!(matches!(history.check().await, Ok(())));
-    assert_eq!(history.status(), Some(&VerificationStatus::Verified));
+
+    // History with single valid user is valid
+    assert!(matches!(history.check().await, Ok(_)));
```
FintanH, line 339: For future work, I'd prefer we use `assert_eq` and the expected value. These kinds of checks can be too permissive, in my opinion. For example, I witnessed the weird behaviour [above](https://github.com/radicle-dev/radicle-link/pull/147/files#diff-e8128eef3fbc5e6398034e0c89623737R300).

Also, `assert_eq` gives better information about the differences, instead of just having a `false` :shrug:

PR #148, opened 2020-05-15
Phantom Status

FintanH:
I took the liberty of testing out using `PhantomData` on `Entity`. I think the changes are pretty cool since you can't push an `Entity<T, Unknown>` onto `UserHistory` -- yay type safety!

I removed the `VerificationStatus` from `Entity` because now the types tell you when things were successful, otherwise, you got an error.

I hope it's ok that I went and did this :bow: My curiosity was piqued :grin:

### DIFF
--- a/librad/src/meta/entity.rs
+++ b/librad/src/meta/entity.rs
@@ -459,11 +459,11 @@ where
     /// - the entity has not been already signed using this same key
     /// - this key is allowed to sign the entity (using `check_key`)
     pub async fn sign(
-        mut self,
+        &mut self,
         key: &Key,
         by: &Signatory,
         resolver: &impl Resolver<User<Draft>>,
-    ) -> Result<Entity<T, Signed>, Error> {
+    ) -> Result<(), Error> {
         let public_key = key.public();
         if self.signatures().contains_key(&public_key) {
             return Err(Error::SignatureAlreadyPresent(public_key.to_owned()));
@@ -474,7 +474,7 @@ where
             sig: self.compute_signature(key)?,
         };
         self.signatures.insert(public_key, signature);
-        Ok(self.with_status::<Signed>())
+        Ok(())
     }
 
     /// Check that an entity signature is valid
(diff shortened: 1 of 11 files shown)


librad/src/meta/entity.rs
```diff
@@ -502,11 +459,11 @@ where
...
     /// - this key is allowed to sign the entity (using `check_key`)
     pub async fn sign(
-        &mut self,
+        mut self,
         key: &Key,
         by: &Signatory,
-        resolver: &impl Resolver<User<EntityStatusUnknown>>,
-    ) -> Result<(), Error> {
+        resolver: &impl Resolver<User<Unknown>>,
+    ) -> Result<Entity<T, Signed>, Error> {
```
massimiliano-mantione, line 466: This should not be a signed entity because invoking `sign` signs with just one key.
There's no guarantee that the entity has been signed with *all* keys.
And returning an error here makes no sense: the entity has been signed with the provided key, the operation is successful, it's just that the entity is not necessarily _completely_ signed.

librad/src/meta/entity.rs
```diff
@@ -526,33 +483,32 @@ where
...
-    fn with_status<NewSt>(&self, status_marker: NewSt) -> Entity<T, NewSt>
-    where
-        NewSt: VerificationStatusMarker,
-    {
+    fn with_status<NewSt>(self) -> Entity<T, NewSt> {
         Entity::<T, NewSt> {
-            status: NewSt::status(),
-            status_marker,
-            name: self.name.clone(),
+            status_marker: PhantomData,
```
massimiliano-mantione, line 501: This is what I was missing! I just did not realize that I could use `PhantomData` as a "value" and the compiler would infer the correct type.
And this was the only reason why I did not use `PhantomData ` and I essentially reimplemented it.

librad/src/meta/entity.rs
```diff
@@ -145,50 +146,12 @@ pub trait VerificationStatusMarker {
...
-impl VerificationStatusMarker for EntityStatusUnknown {
-    fn status() -> VerificationStatus {
-        VerificationStatus::Unknown
-    }
-}
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Verified;
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Signed;
+#[derive(Debug, Clone, PartialEq, Eq)]
```
massimiliano-mantione, line 153: Since we are already renaming types, maybe we could rename `Unknown` as `Draft`?

librad/src/meta/entity.rs
FintanH: Ah great point! :smile: Thanks for catching that

librad/src/meta/entity.rs
FintanH: Ya, it's a bit of a weird one to get your head around immediately :grin:

PR #149, opened 2020-05-15
Monorepo

### LINKED ISSUE #144: Implement monorepo RFC
#127 

Entails changing the `git::repo` and `git::transport` and `git::server`. Prerequisite for #95, #124, #123, #138, #122

kim:
Closes #144

### DIFF
--- a/librad/src/git/ext.rs
+++ b/librad/src/git/ext.rs
@@ -1,0 +1,179 @@
+// This file is part of radicle-link
+// <https://github.com/radicle-dev/radicle-link>
+//
+// Copyright (C) 2019-2020 The Radicle Team <dev@radicle.xyz>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 or
+// later as published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+//! Extensions and wrappers for `git2` types
+
+use std::{fmt, ops::Deref};
+
+use serde::{de::Visitor, Deserialize, Deserializer, Serialize, Serializer};
+
+/// Serializable [`git2::Oid`]
+#[derive(Debug)]
+pub struct Oid(pub git2::Oid);
+
+impl Serialize for Oid {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        self.0.to_string().serialize(serializer)
+    }
+}
+
+impl<'de> Deserialize<'de> for Oid {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        struct OidVisitor;
+
+        impl<'de> Visitor<'de> for OidVisitor {
+            type Value = Oid;
+
+            fn expecting(&self, f: &mut fmt::Formatter) -> fmt::Result {
+                write!(f, "a hexidecimal git2::Oid")
+            }
+
+            fn visit_str<E>(self, s: &str) -> Result<Self::Value, E>
+            where
+                E: serde::de::Error,
+            {
+                s.parse().map(Oid).map_err(serde::de::Error::custom)
+            }
+        }
+
+        deserializer.deserialize_str(OidVisitor)
+    }
+}
+
+impl Deref for Oid {
+    type Target = git2::Oid;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+pub fn is_not_found_err(e: &git2::Error) -> bool {
+    e.code() == git2::ErrorCode::NotFound
+}
+
+pub trait Git2ErrorExt<T> {
+    fn map_not_found<E, F>(self, f: F) -> Result<T, E>
+    where
+        E: From<git2::Error>,
+        F: FnOnce() -> Result<T, E>;
+}
+
+impl<T> Git2ErrorExt<T> for Result<T, git2::Error> {
+    fn map_not_found<E, F>(self, f: F) -> Result<T, E>
+    where
+        E: From<git2::Error>,
+        F: FnOnce() -> Result<T, E>,
+    {
+        self.or_else(|e| {
+            if is_not_found_err(&e) {
+                f()
+            } else {
+                Err(e.into())
+            }
+        })
+    }
+}
+
+/// Iterator chaining multiple [`git2::References`]
+#[must_use = "iterators are lazy and do nothing unless consumed"]
+pub struct References<'a> {
+    inner: Vec<git2::References<'a>>,
+}
+
+impl<'a> References<'a> {
+    pub fn new(refs: impl IntoIterator<Item = git2::References<'a>>) -> Self {
+        Self {
+            inner: refs.into_iter().collect(),
+        }
+    }
+
+    pub fn from_globs(
+        repo: &'a git2::Repository,
+        globs: impl IntoIterator<Item = impl AsRef<str>>,
+    ) -> Result<Self, git2::Error> {
+        let globs = globs.into_iter();
+        let mut iters = globs
+            .size_hint()
+            .1
+            .map(Vec::with_capacity)
+            .unwrap_or_else(Vec::new);
+        for glob in globs {
+            let iter = repo.references_glob(glob.as_ref())?;
+            iters.push(iter);
+        }
+
+        Ok(Self::new(iters))
+    }
+
+    pub fn names<'b>(&'b mut self) -> ReferenceNames<'a, 'b> {
+        ReferenceNames {
+            inner: self.inner.iter_mut().map(|refs| refs.names()).collect(),
+        }
+    }
+
+    pub fn peeled(self) -> impl Iterator<Item = (String, git2::Oid)> + 'a {
+        self.filter_map(|reference| {
+            reference.ok().and_then(|head| {
+                head.name().and_then(|name| {
+                    head.target()
+                        .map(|target| (name.to_owned(), target.to_owned()))
+                })
+            })
+        })
+    }
+}
+
+impl<'a> Iterator for References<'a> {
+    type Item = Result<git2::Reference<'a>, git2::Error>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.inner.pop().and_then(|mut iter| match iter.next() {
+            None => self.next(),
+            Some(item) => {
+                self.inner.push(iter);
+                Some(item)
+            },
+        })
+    }
+}
+
+/// Iterator chaining multiple [`git2::ReferenceNames`]
+#[must_use = "iterators are lazy and do nothing unless consumed"]
+pub struct ReferenceNames<'repo, 'references> {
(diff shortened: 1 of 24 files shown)


librad/src/git/header.rs
```diff
@@ -0,0 +1,187 @@
...
+                GitService::UploadPackLs => &"UploadPackLs",
+                GitService::UploadPack => &"UploadPack",
+                GitService::ReceivePackLs => &"ReceivePackLs",
+                GitService::ReceivePack => &"ReceivePack",
+            })
+            .finish()
+    }
+}
+
+impl PartialEq for Service {
```
FintanH, line 147: They don't have `PartialEq` impl :scream: I think we need to start making PRs to `git2` :grimacing:

kim: Ya, but for more serious reasons... :disappointed:

kim:
There are a number of issues remaining. The most serious ones currently on my mind are:

1. It turns out that `libgit2` doesn't want us to mess with fetching: it doesn't allow us to fetch refspecs which the remote didn't advertise. 

    This is, ahem, protocol-compliant, but `git` actually has knobs which allow exactly this. For now, this means that we would advertise all refs with their `refs/namespaces/*` prefix (there are other knobs which allow to limit the output). This is not entirely satisfying.

2. Determining if we have a certain `(RadUrn, Oid)` is rather expensive, as we need to figure out if the URN's `Path` belongs to us or a remote we're tracking, **and** we need to walk back the branches if `Oid` is not the current tip. The fix is probably to stick another field in the gossip payload which signifies the creator of that `Oid`, as opposed to the originator of the message (the latter being mutated along the way in order to build the broadcast tree).

kim:
Fixed 2. above, so I think this is good to go in order to unblock the dependent tasks.

librad/src/git/url.rs
```diff
@@ -15,6 +15,8 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+#![allow(unused)]
```
FintanH, line 18: :thinking: what's up with this?

librad/src/git/types.rs
```diff
@@ -0,0 +1,269 @@
...
+            ..self.clone()
+        }
+    }
+}
+
+#[derive(Clone)]
+pub struct Refspec {
+    pub remote: Reference,
+    pub local: Reference,
+    pub force: bool,
```
FintanH, line 127: I think I'd prefer an `enum` instead of a `bool`. What do you think?

librad/src/git/types.rs
```diff
@@ -0,0 +1,269 @@
...
+        remote_peer: &'a PeerId,
+        rad_refs_of: impl Fn(PeerId) -> Result<Refs, E>,
+        certifiers_of: impl Fn(&PeerId) -> Result<HashSet<RadUrn>, E>,
+    ) -> Result<impl Iterator<Item = Self> + 'a, E> {
+        // FIXME: do this in constant memory
+        let mut refspecs = Vec::new();
+
+        for tracked_peer in tracked_peers {
+            // Heads
+            {
```
FintanH, line 176: This is visually pleasing. Thanks :)

librad/src/git/storage.rs
```diff
@@ -0,0 +1,371 @@
...
+            .or_else(|e| {
+                if is_not_found_err(&e) {
+                    Ok(false)
+                } else {
+                    Err(e.into())
+                }
+            })
+    }
+
+    pub(crate) fn has_urn(&self, urn: &RadUrn) -> Result<bool, Error> {
```
FintanH, line 166: Could clean up `has_urn` and `has_ref` since they're essentially the same other than how to make the `str`.

librad/src/git/ext.rs
```diff
@@ -0,0 +1,156 @@
...
+        repo: &'a git2::Repository,
+        globs: impl IntoIterator<Item = impl AsRef<str>>,
+    ) -> Result<Self, git2::Error> {
+        let globs = globs.into_iter();
+        let mut iters = globs
+            .size_hint()
+            .1
+            .map(Vec::with_capacity)
+            .unwrap_or_else(Vec::new);
+        for glob in globs {
```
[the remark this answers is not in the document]
kim: I'm never sure of short-circuiting `Result`s, so prefer the imperative style mostly. We also _have_ to allocate the `Vec` and move it to the iterator because of lifetimes.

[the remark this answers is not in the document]
FintanH: Cool, thanks for the explanation :)

librad/src/net/protocol.rs
```diff
@@ -302,8 +302,8 @@ where
         mut incoming: BoxStream<'_, quic::Result<quic::Stream>>,
         hello: impl Into<Option<gossip::Rpc<SocketAddr, A>>>,
     ) {
-        tracing::info!("New outgoing connection");
         let remote_id = conn.remote_peer_id().clone();
+        tracing::info!(remote_id = %remote_id, "New outgoing connection");
```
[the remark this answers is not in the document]
kim: It wasn't -- I think it's because the span is at trace level, but this at info. So, if you run with `RUST_LOG=info`, the span context will not be logged.

librad/src/paths.rs
```diff
@@ -27,8 +27,7 @@ use directories::ProjectDirs;
 #[derive(Clone)]
 pub struct Paths {
     keys_dir: PathBuf,
-    projects_dir: PathBuf,
```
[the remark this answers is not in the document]
kim: `projects_dir` used to contain the git repos, and `profiles_dir` user profile templates for the CLI.

librad/src/git/url.rs
kim: I made this module `pub(crate)`, but wanted to keep the conversion functions for future use.

librad/src/net/protocol.rs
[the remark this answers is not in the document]
FintanH: Ah, I thought trace _implied_ info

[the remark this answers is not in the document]
FintanH: Oh no, wrong way around :sweat_smile:

librad/src/paths.rs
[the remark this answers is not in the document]
FintanH: Cool, thanks

librad/src/git/types.rs
kim: You have a strange aversion against Mr. Boole 😛 Why would an `enum` be preferable?

librad/src/git/storage.rs
kim: Ya maybe... since we don't know the `RefsCategory`, we'd need an `impl FromStr for Reference`. But then, this may also be too strict.

A useful thing would be some combinator for this `if is_not_found_err` business, but I couldn't figure out what the type of it should be 🤔

librad/src/git/types.rs
FintanH: Because I know I'll come along and get it the wrong way around one day :sweat_smile: Whereas if I have a type to work with I won't (less likely) get them the wrong way around. I've got bad boolean sight :grin:

librad/src/net/protocol.rs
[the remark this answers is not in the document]
kim: Ya it's a very empirical business to get this right... :confused:

Maybe it would be cool to have a formatter which indents the span context, like

```
stuff{fields}
    other_stuff{more}
        even_more_stuff{lots}
```

librad/src/git/types.rs
kim: But you'd always construct it using

```rust
Refspec {
    remote,
    local,
    force: jawohl
}
```

so I'm not sure

librad/src/git/storage.rs
FintanH: Ya, I saw that around the place :thinking: Maybe something like:
```rust
fn on_not_found<T, F>(e: git2::Error, f: F) -> Result<T, Error> where FnOnce() -> T {
  if is_not_found(e) {
    Ok(f())
  } else {
    Err(e.into())
  }
}
```

librad/src/git/types.rs
FintanH: Ya, and I suppose it's only used in the `Display` impl anyway. No biggy.

librad/src/net/protocol.rs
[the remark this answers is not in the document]
FintanH: That would be slick indeed :eyes:

librad/src/git/storage.rs
kim: Ya.. although that's not a lot nicer, because we can't partially apply functions in Rrrrrust. Ideally, the receiver should be `Result<T, git2::Error>`, so you can use it like `git2_thing.map_not_found(|| bla())`.

kim: Ohhh:

```rust
pub trait Git2ErrorExt<T> {
    fn map_not_found<E: From<git2::Error>>(self, f: impl FnOnce() -> Result<T, E>) -> Result<T, E>;
}

impl<T> Git2ErrorExt<T> for Result<T, git2::Error> {
    fn map_not_found<E: From<git2::Error>>(self, f: impl FnOnce() -> Result<T, E>) -> Result<T, E> {
        self.or_else(|e| {
            if is_not_found_err(&e) {
                f()
            } else {
                Err(e.into())
            }
        })
    }
}
```

FintanH: :kissing_closed_eyes::ok_hand:

librad/src/git/storage.rs
```diff
@@ -0,0 +1,371 @@
...
+    pub(crate) fn has_commit(&self, urn: &RadUrn, oid: git2::Oid) -> Result<bool, Error> {
+        let span = tracing::warn_span!("Storage::has_commit", urn = %urn, oid = %oid);
+        let _guard = span.enter();
+
+        if oid.is_zero() {
+            return Ok(false);
+        }
+
+        let git = self.lock();
+        let commit = git.find_commit(oid);
```
massimiliano-mantione, line 125: Stupid readability nitpick: why not directly `match git.find_commit(oid)`?

kim: `borrowck` tells me that `git may not live long enough` (weirdly, because `git` is let-bound), and suggest to `let x = match ...; x` :/ Not sure which is more readable...

librad/src/git/storage.rs
kim: 2aba2b816aea79d29adefe94cff03760a1971b7b

"Wat?" moment: I had to change this to not use `impl Trait`, but proper trait bounds, because:

If type inference fails, one is not allowed to use the turbofish when there is an `impl Trait` involved in _one_ of the arguments (I actually needed to annotate the type of `E`, but no luck).

librad/src/net/peer.rs
```diff
@@ -76,17 +70,37 @@ pub struct BindError {
...
+        PeerId::from(&self.key)
     }
 
     pub fn public_key(&self) -> PublicKey {
         self.key.public()
     }
 
+    // FIXME: this should not be here, but we can't otherwise do entity signing
+    // in tests
+    pub fn key(&self) -> &SecretKey {
```
massimiliano-mantione, line 100: Can't we put it under `#[cfg(test)]`?

kim: Unfortunately no: integration tests (those in the `tests/`) don't run with with `cfg(test)` enabled apparently.

I though that perhaps we will eventually have a method `Peer::sign<T>(meta: &mut Entity<T>) -> ??`, but we need to equip the peer with a resolver for this.

librad/src/git/storage.rs
massimiliano-mantione: Nah, this assignment is simpler in the end.

PR #151, opened 2020-05-20
Peer events

kim:
Introduce an upstream event channel for `net::Peer`.

Allows to subscribe to "background" activity triggered by gossip. Note that running the tests multithreaded is still flaky, so there might be some other issue.

On top of #149


kim:
Yeah, so, I think the config is not flushed to disk properly :/

PR #152, opened 2020-05-20
Bump and fix surf revision

FintanH:
Depends on https://github.com/radicle-dev/radicle-surf/pull/112

Bumps the surf revision so that proxy can make use of the latest changes. Also link can now use the `RepositoryRef` version of the `Browser`.

### DIFF
--- a/librad/src/git/repo.rs
+++ b/librad/src/git/repo.rs
@@ -44,6 +44,7 @@ use crate::{
     peer::PeerId,
     uri::{self, RadUrl, RadUrn},
 };
+use radicle_surf::vcs::git as surf;
 
 pub use storage::Tracked;
 
@@ -77,6 +78,9 @@ pub enum Error {
 
     #[error(transparent)]
     Git(#[from] git2::Error),
+
+    #[error(transparent)]
+    Surf(#[from] surf::error::Error),
 }
 
 pub struct Repo {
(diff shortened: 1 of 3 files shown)


librad/src/git/repo.rs
```diff
@@ -429,8 +435,8 @@ impl Repo {
     /// Obtain a [radicle-surf] `Browser` for this repo, consuming self.
     ///
     /// [radicle-surf]: https://github.com/radicle-dev/radicle-surf
-    pub fn browser(self) -> Result<surf::Browser, Error> {
-        let bro = surf::Browser::new(self.repo.into())?;
+    pub fn browser(&'_ self) -> Result<surf::Browser<'_>, Error> {
```
xla, line 438: :scream_cat:

FintanH:
I'm going to hold off on this until I'm sure that I can do the needed surf stuff in proxy

librad/src/git/repo.rs
kim, line 438: This is correct now, but with #149 we won't be able to return a `Browser`. The reason is that we do not want to hand out a RAII lock to the library user -- while the lock is held, no progress can be made reacting to network events.

Instead, we want something like:

```rust
pub fn readonly(&self) -> Result<surf::Repository, Error> {
    let path = // ...
    surf::Repository::open(path, self.namespace())
}
```

kim: Depends on radicle-dev/radicle-surf#101, obviously

FintanH: Ah interesting :thinking: Can follow up with that after again once https://github.com/radicle-dev/radicle-surf/issues/101 is also in

FintanH:
@kim Does this change make sense? Grabbing the `Browser` from `Locked` instead of `Repo`.

PR #155, opened 2020-05-27
Unlock

kim:
Alright, I promise to hold off on big-ass changes like this for a while, but I felt like this one has to go in in order to enable upstream integration :)

So, this whole locking business turned out to be rather peculiar, and so I went and studied `libgit2` for a bit to see how it deals with concurrent access to a repository. The outcome is that it does account for it (and if it didn't, we'd be screwed anyway). _But_ there might be slight inconsistencies when accessing the repo with two different handles -- for example, `libgit2` goes to great lengths to detect odb changes, but _I think_ less so with the refsdb (ie. you create a branch with one handle, and _may_ not immediately see it from the other). I think this is perfectly acceptable.

Hence, this change does the following:

* The networking stack gets it's own `git2::Repository` behind a mutex (it has to `Sync` because of futures)
* After boostrapping a network `Peer` object, you obtain a `PeerApi`, which is the main (and sole) entry point for `radicle-link` applications. You can get access to a `Storage`, which is a separate instance of a `git2::Repository`, but _not_ behind a lock. You can "clone" it, which involves obtaining a new handle.
* All git operations are now implemented in terms of `Storage`, with `Repo` just being a namespace-aware wrapper.

### DIFF
--- a/librad/tests/propagation_basic.rs
+++ b/librad/tests/propagation_basic.rs
@@ -15,15 +15,16 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
-use std::future::Future;
+#![feature(async_closure)]
 
-use futures::future::Either;
+use std::time::Duration;
+
+use futures::{future, stream::StreamExt};
 
 use librad::{
-    internal::sync::Monitor,
     meta::{entity::Signatory, project::ProjectInfo},
-    net::peer::BoundPeer,
-    peer::PeerId,
+    net::peer::{FetchInfo, Gossip, PeerEvent, Rev},
+    uri::{self, RadUrn},
 };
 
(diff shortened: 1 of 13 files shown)


librad/tests/propagation_basic.rs
```diff
@@ -36,104 +37,124 @@ use fixtures::{Alice, Radicle};
...
-        }),
-    )
+            .unwrap();
+        }
+
+        // Add a commit on peer1
+        let commit_id = {
+            let repo = peer1_storage.open_repo(radicle.urn()).unwrap();
+            let empty_tree = {
+                let mut index = repo.index().unwrap();
```
kim, line 104: Note that this one is pretty unsound -- the index is not meant for concurrent access, let alone on a bare repo, afaik. In reality, these commits will be added in the working copy, though.

PR #163, opened 2020-06-04
Simplified entity sign methods (fixes #161).

### LINKED ISSUE #161: User::create can come with signing
It seems like [User::create](https://github.com/radicle-dev/radicle-link/blob/master/librad/src/meta/user.rs#L162) could come signed with its own `SecretKey`. This would make it easier to bootstrap a user without needing a resolver.

massimiliano-mantione:
See #161...

### DIFF
--- a/librad/src/meta/entity.rs
+++ b/librad/src/meta/entity.rs
@@ -420,6 +420,50 @@ where
         Ok(key.sign(&self.canonical_data()?))
     }
 
+    /// Given a private key owned by the entity, sign the current entity
+    ///
+    /// The following checks are performed:
+    ///
+    /// - the entity has not been already signed using this same key
+    /// - this key is owned by the current entity
+    pub fn sign_owned(&mut self, key: &SecretKey) -> Result<(), Error> {
+        let public_key = key.public();
+        if self.signatures().contains_key(&public_key) {
+            return Err(Error::SignatureAlreadyPresent(public_key));
+        }
+        if !self.has_key(&public_key) {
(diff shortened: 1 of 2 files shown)


librad/src/meta/entity.rs
```diff
@@ -420,6 +420,50 @@ where
...
+    ///
+    /// - the entity has not been already signed using this same key
+    /// - this key is owned by the provided user
+    pub fn sign_by_user(&mut self, key: &SecretKey, user: &User<Verified>) -> Result<(), Error> {
+        let public_key = key.public();
+        if self.signatures().contains_key(&public_key) {
+            return Err(Error::SignatureAlreadyPresent(public_key));
+        }
+        if !user.has_key(&public_key) {
+            return Err(Error::UserKeyNotPresent(user.urn(), public_key));
```
FintanH, line 457: I would say this error is _sort_ of misleading. It's more the case that the `SecretKey`/`PublicKey` is not owned by this `User`. What do you think?

massimiliano-mantione: For now I was just consistent with what the original, generic `sign` method does: if you provide a signatory that is a `Signatory::User` and the user does not have the key, that is the error.
So, if we change it, we should change it in both places and keep it consistent.
Changing it means just renaming it, and I am definitely open to better error names...

FintanH: Let's keep it for now :+1:

FintanH:
I'll let you do the honours :bow:

PR #164, opened 2020-06-04
Idempotent Track

FintanH:
While integrating `librad` into `proxy`, I was running into an issue where tracking was attempted to be done twice on an entity. Presumably, it's because we insert the user first, then create a project with that user. But not 100% sure about this.

Either way, it seems useful for tracking to idempotent and not scream when we try to track the same peer twice.

### DIFF
--- a/librad/src/git/storage.rs
+++ b/librad/src/git/storage.rs
@@ -358,17 +358,10 @@ impl Storage {
             "Storage::track"
         );
 
-        match self.backend.remote(&remote_name, &url) {
-            Ok(_) => Ok(()),
-            Err(err) => {
-                if err.code() == git2::ErrorCode::Exists {
-                    tracing::warn!(urn = %urn, peer = %peer, "already tracking remote");
-                    Ok(())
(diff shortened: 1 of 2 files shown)


xla:
Ref radicle-dev/radicle-upstream#414

librad/src/git/storage.rs
```diff
@@ -358,8 +358,17 @@ impl Storage {
             "Storage::track"
         );
 
-        let _ = self.backend.remote(&remote_name, &url)?;
-        Ok(())
+        match self.backend.remote(&remote_name, &url) {
```
kim, line 361:
```suggestion
        self.backend
            .remote(&remote_name, &url)
            .map(|_| ())
            .map_not_found(|| Ok(()))
```

FintanH: Noice

FintanH: Actually it'll be `already_exists`, so I'm adding that to the extensions

PR #174, opened 2020-06-11
Namespace Browser

FintanH:
Browse the namespace for the given `RadUrn` and `revision`.

The namespace value is the `id` of the urn.


librad/src/git/repo.rs
```diff
@@ -64,8 +64,10 @@ impl<'a> Repo<'a> {
     }
 
     /// Obtain a read-only view of this repo
-    pub fn browser(&'_ self) -> Result<surf::Browser<'_>, Error> {
-        self.storage.browser(&self.urn).map_err(Error::from)
+    pub fn browser(&'_ self, revision: &str) -> Result<surf::Browser<'_>, Error> {
```
kim, line 67: What's a "revision" exactly?

FintanH: That would be a ref for setting the history. I was thinking it could just be the `default_branch` of the project, but we don't have resolvers yet.

This plays into the conversation we had yesterday about git's magic string resolving for references.

kim: Ya ok, not great, but perhaps we can live with it until more things are in place

FintanH: Ya, I'd prefer to get the default branch

PR #175, opened 2020-06-12
Generalised combinator for conditional branching on `Result` error types

kim:


librad/src/internal/result.rs
```diff
@@ -0,0 +1,52 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+pub trait ResultExt<T, E> {
+    /// Calls `f` if the result is [`Err`], **and** the predicate `pred` on the
+    /// error value returns true. Otherwise returns the [`Ok`] value of
+    /// `self`. Note that `f` may change the error type, so as long as the
+    /// target type can be converted from the original one.
```
kim, line 22: ... which makes type inference fall on its nose a lot of the times lolrust

PR #181, opened 2020-06-16
Handle `rad/self`

kim:


librad/src/git/storage.rs
```diff
@@ -522,94 +619,206 @@ impl Storage<WithSigner> {
...
+                        Ok::<_, Error>(Reference::rad_id(id.urn().id))
+                    },
+
+                    RadSelfSpec::Urn(self_urn) => {
+                        let meta: User<Draft> = self.metadata(&self_urn)?;
+                        Config::try_from(&self.backend)?.guard_user_valid(&meta)?;
+                        Ok(Reference::rad_id(self_urn.id))
+                    },
+
+                    RadSelfSpec::None => unreachable!(),
```
FintanH, line 808: I think it's possible to avoid this if you use `return` in the `None` case, and then do `let target = match spec { /* etc. */ }` followed by the `self.backend` code. It'd only be an issue if the first case was moved below, so no big deal :shrug:

kim:
What this does:

* When creating a repo:
    * One of the entity signatures must match the public part of your device-local key
    * A symref `rad/self` is created which points to either `rad/id` (if the entity is "self-signed"), or the top-level entity found via its URN (that is, the `rad/id` ref there). Of course, the latter must exist
    * We also create all the `rad/ids/*` symrefs as per the RFC (bailing out if the respective top-level entities don't exist), but there's currently no validation rule which ensures they cover the whole set of certifiers down the road

* When cloning a repo, there's some (mostly stub) assertions about the remote peer's `rad/id`. If they pass, we just use that as our own `rad/id`

* When fetching a repo, there's a bunch more refspecs to consider

* One can configure the local state to use a particular `User` as a default `rad/self` reference. It is currently required that this user is `Verified`, which I believe we can't actually construct yet lol. Once we can, there needs to be a runtime check for the verification status when this default user is actually used -- it might not currently be valid.

* **Setting** `rad/self` on a repo one didn't create is a separate (and optional) operation -- for no other reason than either way being an arbitrary design decision, and me mandating earlier that it SHALL be possible to track a repo without revealing one's "identity".

librad/src/git/storage.rs
kim: `return` what? The result of the pattern match is bound to a variable

librad/src/git/repo.rs
```diff
@@ -98,7 +98,10 @@ impl<'a> Repo<'a, WithSigner> {
     /// Fetch new refs and objects for this repo from [`PeerId`]
     pub fn fetch(&self, from: &PeerId) -> Result<(), Error> {
         self.storage
-            .fetch_repo(&self.urn, from)
+            .fetch_repo(RadUrl {
```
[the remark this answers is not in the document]
kim: Not really, `RadUrl` has one additional piece of information over `RadUrn`, so it's not an isomorphism, so `From`, `Into`, et. al. don't make sense. There is `RadUrn::into_rad_url`, but I came to dislike this a bit over just the constructors, because of named fields.

librad/src/git/storage.rs
FintanH: So I meant here: https://github.com/radicle-dev/radicle-link/pull/181/files#diff-b55217e1bbc5075454fe7f8b69de56fcR789-R790 you could call `return `Ok(())` and `return reference.delete().map_err(Error::from)`. Then float the `let target` as well of the rest of logic outside the outer `match`. I believe the `return` will still work for the type checker, as long as the rest of the arms of the `match` have the same types.

That way we'll have exhaustive pattern matching, and no `unreachable`.

FintanH:
> One can configure the local state to use a particular User as a default rad/self reference. It is currently required that this user is Verified, which I believe we can't actually construct yet lol. 

We can if we verify a single signed user with a resolver that resolves to itself :stuck_out_tongue_closed_eyes:

librad/src/git/storage.rs
```diff
@@ -493,15 +537,68 @@ impl Storage<WithSigner> {
...
+        let rad_id = Reference::rad_id(meta.urn().id);
+        let rad_self = Reference::rad_self(meta.urn().id, None);
+        let rad_self_target = match &self_sig.by {
+            Signatory::OwnedKey => rad_id.clone(),
+            Signatory::User(urn) => Reference::rad_id(urn.id.clone()),
+        };
+
+        // Invariants
+        {
+            // Check if we could create `rad/self`, pointing to a valid target
```
FintanH, line 560: So I think this comment may be confusing me slightly. The word "could" implies that we have created `rad/self` before this point. But is it actually the case that we're checking if we "can" create a `rad/self` symlink below?

librad/src/git/storage.rs
```diff
@@ -522,94 +619,206 @@ impl Storage<WithSigner> {
...
+                            urn.id.clone(),
+                            &RadUrn::new(
+                                certifier_hash.clone(),
+                                uri::Protocol::Git,
+                                uri::Path::empty(),
+                            ),
+                        );
+                        let certifier_id = Reference::rad_id(certifier_hash);
+                        self.backend
+                            .reference_symbolic(
```
FintanH, line 742: I'm quite fond of these typed `Reference`s. I wonder if we could extend that API a bit with `SymReferences` which could take two `Reference`s? 

It might also be useful to have a method on them, say `create(self, repo: git2::Repostiory) -> Result<(), Error>` that handles the `to_string()` calling.

What do you think?

librad/src/git/storage.rs
kim: I don't get it, but may you like this better 7deb942

librad/src/git/storage.rs
kim: Idk, "could" is the subjunctive, no? We _could_ create them _if_ `commit_initial_meta` was successful.

librad/src/git/storage.rs
kim: Ya, patches welcome :) It's probably that one would want a constructor with named fields, as the options are not only boolean-blind, but also what's pointing to what blind.

librad/src/git/storage.rs
FintanH: Ya, but in this case, it's probably ambiguous. Because "could" is also the past of "can" so I was reading this as "Check if we were able to create rad/self".

I appreciate you changing it :)

librad/src/git/storage.rs
FintanH: :kissing_closed_eyes::ok_hand:

PR #186, opened 2020-06-17
Get ANY Entity

FintanH:
A lot of the time, the proxy wants a specific kind entity. This makes it easier so that we can fetch that kind of entity or error if it's not the kind we thought it would be.


FintanH:
> What I think is great about this change is coerce, map and try_map .

:heart:

 > Would challenge that the tiny project and user methods really bring value instead of driving single protocol consumer concerns into this library. This could easily be done in upstream itself with the help of the newly introduced helpers.

So just to rehash our conversation over a call: I think the combinators allow the users of this library to extend it (thank you Functors and Poor-Man's Traversables). I think, however, that it's useful to provide the the `project` and `user` functions regardless to save our users from writing boiler-plate. I think it's common that libraries provide the combinators as well as the final presentation for the cases the authors envision :)

librad/src/git/storage.rs
```diff
@@ -546,6 +563,55 @@ impl Storage {
...
+    }
+
+    /// Get the [`Entity`] found at the provided [`RadUrn`]. Then try to map the
+    /// [`EntityInfo`] into the target `T`.
+    ///
+    /// # Errors
+    ///   * If the entity resolution fails.
+    ///   * If the mapping fails, an [`Error::IncorrectEntityInfo`] will be
+    ///     returned.
+    pub fn entity_info_or<F, T>(
```
kim, line 597: I understand your desire to make the error message informative, but I don't think it's worth it -- the entity "kind" is not exhaustive. Iow, why not simply try to deserialize the blob contents into `T`, and if that fails return either the raw bytes or the `serde_json::Value` in the error payload?

FintanH: Ya, fair point. I'll make the change.

FintanH: Makes much more sense :ok_hand: The irony being that this branch is called specific-entity.

PR #189, opened 2020-06-19
Emit protocol Listening event

xla:
Emit an explicit event similar to the tracing for consumers to reason about the successful startup of the peer.

### DIFF
--- a/librad/src/net/protocol.rs
+++ b/librad/src/net/protocol.rs
@@ -151,7 +151,11 @@ where
         tracing::info!(msg = "Listening", local.addr = ?endpoint.local_addr());
 
         self.subscribers
-            .emit(ProtocolEvent::Listening(endpoint.local_addr().unwrap()))
+            .emit(ProtocolEvent::Listening(
+                endpoint
+                    .local_addr()
+                    .expect("unable to get local endpoint addr"),
+            ))
             .await;
 
         futures::stream::select(incoming, futures::stream::select(bootstrap, gossip_events))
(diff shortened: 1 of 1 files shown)


librad/src/net/protocol.rs
```diff
@@ -149,6 +150,10 @@ where
 
         tracing::info!(msg = "Listening", local.addr = ?endpoint.local_addr());
 
+        self.subscribers
+            .emit(ProtocolEvent::Listening(endpoint.local_addr().unwrap()))
```
FintanH, line 154: Maybe an `expect` instead of an `unwrap` would be better :)

librad/src/net/protocol.rs
```diff
@@ -149,6 +150,14 @@ where
 
         tracing::info!(msg = "Listening", local.addr = ?endpoint.local_addr());
 
+        self.subscribers
+            .emit(ProtocolEvent::Listening(
+                endpoint
```
kim, line 155: Could you maybe bind this at the start of the method, and stick it into the `trace_span!` (like `trace_span!("Protocol::run", local_addr = %local_addr)`). The `info!` can then be removed.

I figure this would much improve log output when there are more than one peers in a single process, as you can tell which one emitted the log line.

librad/src/net/protocol.rs
```diff
@@ -147,7 +151,11 @@ where
             .await
             .map(|event| Ok(Run::Gossip { event }));
 
-        tracing::info!(msg = "Listening", local.addr = ?endpoint.local_addr());
+        tracing::info!(msg = "Listening", local.addr = %local_addr);
```
FintanH, line 154: You don't need to add the `local_addr` here because the span captures it above

PR #190, opened 2020-06-19
List Entities

### LINKED ISSUE #142: Listing projects
`project::list` was removed in #137. Pending #123 and #127, it should be resurrected.

FintanH:
Fixes #142 

Allows us to list entities and apply a filter if we want to target one kind specifically.

### DIFF
--- a/librad/src/git/storage.rs
+++ b/librad/src/git/storage.rs
@@ -40,7 +40,14 @@ use crate::{
         result::ResultExt,
     },
     keys::SecretKey,
-    meta::entity::{self, data::EntityInfoExt, Draft, Entity, Signatory},
+    meta::entity::{
+        self,
+        data::{EntityInfo, EntityInfoExt},
+        Draft,
+        Entity,
+        GenericDraftEntity,
+        Signatory,
+    },
     paths::Paths,
     peer::PeerId,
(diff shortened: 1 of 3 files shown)


librad/src/git/storage.rs
```diff
@@ -554,11 +554,35 @@ impl Storage {
...
+            )?)
         })
     }
 
+    /// Get all the [`Entity`] data in this `Storage`.
+    ///
+    /// The caller has the choice to filter on the [`EntityInfo`], which is
+    /// useful when the you want a list of a specific kind of `Entity`.
+    pub fn entities(&self) -> Result<Vec<GenericDraftEntity>, Error> {
+        let references = self.backend.references_glob("refs/namespaces/**/rad/id")?;
```
kim, line 568: Why did you choose to glob on `**`? Iiuc, this would also match `rad/id` branches of remotes. Those may or may not be verified, and may or may not agree. In the best case, the output would include duplicate `GenericDraftEntity` values.

FintanH: Ah you're right. I messed up the globbing in my head. What I want is `refs/namespace/*/rad/id` to get all the first level entities. I can submit a patch. Sorry for the haste.

PR #198, opened 2020-06-22
Pass secret key to storage only where needed

### LINKED ISSUE #196: Pass SecretKey to `git::Storage` only where needed
First step towards utilising radicle-dev/radicle-keystore#1

kim:
Fixes #196

---

First step towards using radicle-dev/radicle-keystore#1, instead of passing around the secret key. On top of #197

### DIFF
--- a/librad/src/git/storage.rs
+++ b/librad/src/git/storage.rs
@@ -19,6 +19,7 @@ use std::{
     borrow::Borrow,
     collections::{BTreeMap, HashMap, HashSet},
     io,
+    marker::PhantomData,
     ops::{Deref, Range},
     path::Path,
 };
@@ -50,7 +51,7 @@ use crate::{
     keys::SecretKey,
     meta::entity::{self, data::EntityInfoExt, Draft, Entity, GenericDraftEntity, Signatory},
     paths::Paths,
-    peer::PeerId,
+    peer::{self, PeerId},
     uri::{self, RadUrl, RadUrn},
 };
 
@@ -69,6 +70,15 @@ pub enum Error {
 
     #[error("Metadata is not signed")]
     UnsignedMetadata,
+
+    #[error("Signer key does not match key used at initialisation")]
+    SignerKeyMismatch,
+
+    #[error("Can't refer to the local key for this operation")]
+    SelfReferential,
+
+    #[error(transparent)]
+    PeerId(#[from] peer::conversion::Error),
 
     #[error(transparent)]
     Urn(#[from] uri::rad_urn::ParseError),
@@ -95,13 +105,17 @@ pub enum Error {
     Io(#[from] io::Error),
 }
 
-pub struct Storage {
+pub type NoSigner = PhantomData<!>;
+pub type WithSigner = SecretKey;
+
+pub struct Storage<S> {
     pub(super) backend: git2::Repository,
-    pub(crate) key: SecretKey,
+    peer_id: PeerId,
+    signer: S,
 }
 
 // FIXME(kim): we really don't want to export this
-impl Deref for Storage {
+impl<S> Deref for Storage<S> {
     type Target = git2::Repository;
 
     fn deref(&self) -> &Self::Target {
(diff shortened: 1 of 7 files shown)


kim:
Ooof, no, this doesn't actually work -- the `Signer`'s `PublicKey` and the `PeerId` must match. Lemme just go for the `Signer` trait directly.

kim:
@FintanH @massimiliano-mantione Could you have a careful look, please?

The idea is the following: we need access to a `Storage` without knowing the local secret key, kind of like a "read only mode". Once we initialised a `Storage` however, all subsequent operations which **do** require access to the secret key (resp. a `Signer`) **must** use the same key.

So, I introduced a generic type on `Storage`, which determines the type of the `signer` _value_ -- which is either an actual signer, or nothing at all. Then, I split the operations into three different inherent impls: ones which require a signer, ones which (strictly) don't, and ones which can be polymorphic. On initialisation, a signer must be provided, and we store the public part in the git config -- if we later construct a `Storage` with a signer, but it uses a different key, that's a runtime error.

Note that I held back on actually using the `Signer` trait just now, as that would mean to deal with `async` in a few more places.

librad/src/git/storage.rs
```diff
@@ -676,6 +764,7 @@ impl Storage {
                     Signatory::OwnedKey => (peer_id, None),
                 }
             })
+            .filter(|(peer, _)| peer != self.peer_id())
```
kim, line 767: Fixes #199

librad/src/git/storage.rs
```diff
@@ -95,62 +105,383 @@ pub enum Error {
...
     }
 
+    pub fn peer_id(&self) -> &PeerId {
+        &self.peer_id
+    }
+
+    pub fn open_repo(&self, urn: RadUrn) -> Result<Repo<S>, Error> {
+        {
+            let id_ref = Reference::rad_id(urn.id.clone());
+            if !self.has_ref(&id_ref)? {
```
FintanH, line 145: Could this be `has_urn` instead?

kim:
> How do we sign any entity?

Yeah, that's one of the reasons we need to talk about the `PeerApi` surface -- currently, you'd need to pass the signer to `PeerConfig`, but _also_ keep a copy of it to do the signing. The `PeerApi` could wrap the signing methods and use the signer it knows about internally. Of course, this breaks postfix notation, but would preserve modularity _within_ `librad`.

librad/src/git/storage.rs
kim: Yap

FintanH:
> Yeah, that's one of the reasons we need to talk about the PeerApi surface -- currently, you'd need to pass the signer to PeerConfig, but also keep a copy of it to do the signing. The PeerApi could wrap the signing methods and use the signer it knows about internally. Of course, this breaks postfix notation, but would preserve modularity within librad.

Well call `create_repo` is inside the `WithSigner` version of `Storage` so we could tuck away the signing of the passed in `Entity` in there. The _other_ issue is: where do we get our own `PublicKey` from for adding it to an `Entity` that we create? Should that stay in the `PeerApi`?

kim:
> create_repo is inside the WithSigner version of Storage so we could tuck away the signing of the passed in Entity in there.

That's what I mean by "modularity": the storage should receive an `Entity<T, Verified>`, and not be concerned about this signing business.

> The other issue is: where do we get our own PublicKey from for adding it to an Entity that we create?

BYO :)

It's up to the application to deal with the key storage, password prompting, agent management etc. `librad` should just be given a thing that tells its what the public key is, and which can sign a bunch of bytes. You _also_ need one of those to boot up the `Peer`, which makes this look a bit awkward.

FintanH:
> That's what I mean by "modularity": the storage should receive an Entity<T, Verified>, and not be concerned about this signing business.

Ya, good point. I guess this will all become clearer when we discuss what needs a signed/verified entity and how that signing is coordinated. At the moment the concerns seem conflated in a way, to me at least :)

librad/src/lib.rs
```diff
@@ -15,8 +15,9 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
-#![feature(str_strip)]
 #![feature(bool_to_option)]
+#![feature(never_type)]
```
xla, line 19: :exploding_head:

massimiliano-mantione:
> On initialisation, a signer must be provided, and we store the public part in the git config -- if we later construct a `Storage` with a signer, but it uses a different key, that's a runtime error.

Just to be sure, the "migration path" to change the PeerID of a device is to change that public key in the git config, right?

The scenario I have in mind is: I have a notebook and I have a large monorepo on it (I have been using Radicle a lot with lots of projects).

Then I decide that the PeerID (key) I use on this notebook must be changed.
But I am not going to change notebook, not even to reinstall the OS or repartition the disk.

My expectation is that I should be able to reuse the git storage, re-verifying everything (every signature) but not necessarily re-downloading everything from somewhere.
Should Upstream provide support for this?

massimiliano-mantione:
> That's what I mean by "modularity": the storage should receive an `Entity<T, Verified>`, and not be concerned about this signing business.

:100: :+1:

kim:
>  the "migration path" to change the PeerID of a device

While working on this, I realised that it is somewhat complicated: there is data in the monorepo which was signed by that key, and lives under some namespace hierarchy which we consider the "owned" part of the repo. Also, peer will tuck these parts under the `PeerId` (`remotes/<PeerId>/*`). So, if you'd like to rotate your device key, but keep the repo, you'd need to "quarantine" those refs and start the "owned" part from scratch. Essentially like cloning everything from a remote peer, and then setting up any `rad/id`, `rad/self`, etc again. Preserving your user identity requires access to the old key, ofc.

PR #223, opened 2020-06-27
Bumps surf version

sarahscott:


FintanH:
@kim: Agreed. I was starting to get annoyed by this dance too. I'll prep this change :+1:

PR #224, opened 2020-06-27
Remove Surf

FintanH:
This may be _controversial_, but I was thinking we could just remove surf altogether. `Storage` has an `AsRef<git::Repository>` impl so we can create a `Browser` by getting the ref, and then build the `Namespace` from the `RadUrn`'s `id`.

i.e.
```rust
let namespace = Namespace::from(urn.id.to_string());
let browser = Broswer::new_with_namespace(storage.as_ref(), &namespace, default_branch)?;
```

This means that we'll never get any mismatch between the `Browser` `librad` will given back and the `Browser` by the version provided by crates in upstream code.

If this doesn't suit, I can opt for the version bound route instead.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -872,7 +872,6 @@ dependencies = [
  "proptest",
  "quinn",
  "radicle-keystore",
- "radicle-surf",
  "rand 0.7.3",
  "rand_pcg 0.2.1",
  "rcgen",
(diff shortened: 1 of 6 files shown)


kim:
The Deref / AsRef is about to be removed — it took me a while to understand
that auto-dereferencing makes this an inheritance / multiple-inheritance
mechanism. Being able to dereference to a raw git handle is exactly **not**
what we want.

We don’t have a network boundary, unfortunately, which would define the
contract for accessing the backend storage. Defining this boundary by means
of libraries is not obvious. For now, I would prefer to create a fresh
browser from the `Paths` where needed, and otherwise cut all ties to the
`librad` crate.

It should also be clear by now that higher level coco-concepts like code
browsing, issues, revisions, etc are outside the domain of the radicle-link
protocol. So, if there’s any benefit from a monorepo, that repo should be
radicle-upstream.

FintanH:
Ya that makes sense @kim. I wasn't sure if the `AsRef` was gonna go. Thanks for the `Paths` suggestion, didn't occur to me initially but makes sense :+1:

PR #227, opened 2020-07-01
Add get for rad/self

FintanH:
Allow us to retrieve the rad/self for some RadUrn (and PeerId) in our Storage.
For Repo it delegates to the Storage passing its own RadUrn.


FintanH:
> Wondering if we could reduce duplication of this and metadata...

Ya, I was thinking similar. I'll give it a crack now actually. Just wanted something working first :+1:

PR #229, opened 2020-07-02
Fix RadUrn

FintanH:
The `urn` method on `Entity` was using `hash` instead of `root_hash`. This means that are `RadUrn`'s weren't stable and the repository was not in the correct format. It would result in a lone `namespace/<id>` with just `rad/refs` while the entity's `rad/id` would live in another `namespace/<id>`.

This PR fixes this issue and adds a test case to ensure that we have each of those refs under the `namespace` for each repo we create.

### DIFF
--- a/librad/src/git/storage.rs
+++ b/librad/src/git/storage.rs
@@ -806,12 +806,15 @@ impl Storage<WithSigner> {
                     },
                 }?;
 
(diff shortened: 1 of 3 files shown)


massimiliano-mantione:
Good catch!
FWIW this is also fixed in #218 but of course it should be pushed before that one lands.

PR #232, opened 2020-07-03
Introduce a local transport for interacting with the storage backend

kim:
This simply connects to stdio of forked `git-upload-pack` / `git-receive-pack`
processes, which are started with a configuration suitable so the caller can
only access the current namespace. Remote tracking branches are hidden for
`receive-pack`, such that pushing to them is prohibited. The transport is set up
with a signer (ie. secret key), such that `Storage::update_ref` can be called
transparently after a successful push.

In short, interacting with a logical repo requires simply registering the
transport, and setting up a remote with a URL of the form `rad://<id-part-of-URN>`.

Note that the transport is set up in a way so it can be reused for a git remote
helper, allowing CLI access to the storage backend. The helper code will be
submitted in a subsequent patch.

We can now also remove the `Deref<Target = git2::Repository>` from `Storage`,
prohibiting access to the raw git handle.


librad/src/git/local/transport.rs
```diff
@@ -0,0 +1,388 @@
...
+        #[cfg(unix)]
+        use std::os::unix::io::{AsRawFd, FromRawFd};
+        #[cfg(windows)]
+        use std::os::windows::io::{AsRawFd, FromRawFd};
+
+        Self {
+            child_stdout: Stdio::from_raw_fd(io::stdout().as_raw_fd()),
+            child_stdin: Stdio::from_raw_fd(io::stdin().as_raw_fd()),
+        }
+    }
```
kim, line 202: @geigerzaehler :point_up_2:

kim:
You might be wondering why the transport has to be _constructed_ with the signing key, when there is [`git2::RemoteCallbacks::credentials`](https://docs.rs/git2/0.13.6/git2/struct.RemoteCallbacks.html#method.credentials). I did, too, but I can't see a way to actually _trigger_ this callback from a custom transport.

librad/src/git/local/url.rs
```diff
@@ -0,0 +1,107 @@
...
+use thiserror::Error;
+
+use crate::{
+    hash::{self, Hash},
+    uri::{self, RadUrn},
+};
+
+#[derive(Clone)]
+pub struct LocalUrl {
+    repo: Hash,
```
massimiliano-mantione, line 32: Is `repo` the namespace? If yes, maybe we could name it `namespace`.

librad/src/git/local/transport.rs
```diff
@@ -0,0 +1,388 @@
...
+pub fn register(settings: Settings) {
+    static INIT: Once = Once::new();
+
+    LocalTransportFactory::new().configure(settings);
+    unsafe {
+        INIT.call_once(move || {
+            git2::transport::register(local::URL_SCHEME, move |remote| {
+                Transport::smart(&remote, true, LocalTransportFactory::new())
+            })
+            .unwrap()
```
massimiliano-mantione, line 59: Why is it safe to `unwrap` here?

librad/src/git/local/transport.rs
```diff
@@ -0,0 +1,388 @@
...
+impl Connected {
+    pub fn wait(&mut self) -> Result<(), Error> {
+        let status = self.process.wait()?;
+        if status.success() {
+            match self.on_success.take() {
+                None => Ok(()),
+                Some(f) => f(),
+            }
+        } else {
+            Ok(())
```
massimiliano-mantione, line 149: Why `Ok(())` if the child process failed?

librad/src/git/local/transport.rs
```diff
@@ -0,0 +1,388 @@
...
+        }
+    }
+}
+
+// Belts + suspenders
+impl Drop for Connected {
+    fn drop(&mut self) {
+        let _ = self
+            .wait()
```
massimiliano-mantione, line 158: I see the "Belts + suspenders" comment, I guess it means "be absolutely sure we do not ignore a `Connected` instance without `wait`ing it.

However this `Drop` implementation seems like papering over incorrect `Connected` creations which should be detected by the `#[must_use = "`wait` must be called"]` attribute.

`Drop` is not fallible but is it normal to drop a running `Connected` instance?
Wouldn't it be better to abort the child process and fail loudly?

librad/src/git/local/transport.rs
kim: The only error this can possibly return is `libgit2`'s allocator failing to allocate memory. I would presume that this is a legitimate reason to panic without prior notice.

librad/src/git/local/url.rs
kim: `namespace` is just how we happen to store it, `repo` is what it is logically

librad/src/git/local/transport.rs
kim: 605ec4cba2d2866ae4e8c026bc79e795e42c3cfc

librad/src/git/local/transport.rs
kim: `must_use` is a lint _warning_ by default, and I'm not sure `#![deny(unused_must_use)]` would have any effect on the library _user_. The issue is that _not_ `wait(2)`-ing on a child process may leave the process as a zombie on some systems, so we must ensure that happens.

Terminating the child if it hasn't already is potentially a good idea, but I'm not sure `git-receive-pack` is `SIGKILL`-safe.

librad/src/git/local/transport.rs
massimiliano-mantione: Ooops, right. It cannot fail in other ways because it is invoked under `ONCE`, and registering it twice is the only reason why the underlying native call would fail (apart from allocations).
I did not notice.

librad/src/git/local/transport.rs
kim: I think this is the best we can do: cead0919061051bfd8afea43ae2b97b8fcf9b9a8

massimiliano-mantione: Indeed!

librad/src/git/local/url.rs
```diff
@@ -0,0 +1,107 @@
...
+#[derive(Debug, Error)]
+#[non_exhaustive]
+pub enum ParseError {
+    #[error("Invalid scheme: {0}")]
+    InvalidScheme(String),
+
+    #[error("Cannot-be-a-base URL")]
+    CannotBeABase,
+
+    #[error("Malformed URL")]
```
FintanH, line 56: Why not transparent here?

librad/src/git/local/url.rs
```diff
@@ -0,0 +1,107 @@
...
+        if url.scheme() != super::URL_SCHEME {
+            return Err(Self::Err::InvalidScheme(url.scheme().to_owned()));
+        }
+        if url.cannot_be_a_base() {
+            return Err(Self::Err::CannotBeABase);
+        }
+
+        let repo = url
+            .host_str()
+            .expect("we checked for cannot-be-a-base. qed")
```
FintanH, line 77: Might be safer if we do:
```rust
let repo = if url.cannot_be_base() {
  return Err(Self::Err::CannotBeABase);
} else {
  /* this logic */
};
```

That way it's explicit we checked and we're in the `else` branch. If someone comes to change this code later, they'll have to think why this was an `if/else`.

librad/src/git/storage.rs
```diff
@@ -153,21 +153,6 @@ pub struct Storage<S> {
     signer: S,
 }
 
-// FIXME(kim): we really don't want to export this
-impl<S> Deref for Storage<S> {
```
FintanH, line 157: :raised_hands:

librad/src/git/local/transport.rs
```diff
@@ -0,0 +1,426 @@
...
+                header,
+                inner: stdout,
+            },
+            write: stdin,
+        })
+    }
+
+    pub fn connect(
+        &mut self,
+        url: LocalUrl,
```
FintanH, line 287: Should we use the `Into<RadUrn>` pattern here to allow us to pass in `RadUrn` as well? Or was there a specific reason we need `LocalUrl` here?

librad/src/git/local/transport.rs
```diff
@@ -0,0 +1,426 @@
...
+}
+
+pub fn register(settings: Settings) {
+    static INIT: Once = Once::new();
+
+    LocalTransportFactory::new().configure(settings);
+    unsafe {
+        INIT.call_once(move || {
+            git2::transport::register(local::URL_SCHEME, move |remote| {
+                Transport::smart(&remote, true, LocalTransportFactory::new())
```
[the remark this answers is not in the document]
kim: To appease the borrow checker (try for yourself, it just doesn't like to move an "outer variable").

librad/src/git/local/transport.rs
```diff
@@ -0,0 +1,426 @@
...
+
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error("No such URN: {0}")]
+    NoSuchUrn(RadUrn),
+
+    #[error(transparent)]
+    Storage(#[from] storage::Error),
+
+    #[error("Child exited unsuccessfully")]
```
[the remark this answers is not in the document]
kim: I considered it, but you can't call a method from the `thiserror` macros. I don't think it's worth moving out the error code, because whether it's `42` or `255` is most likely not going to elucidate much.

librad/src/git/local/transport.rs
kim: Sure you can, but is there a specific reason to make it more polymorphic?

The remote helper gets invoked with `<name of remote> <url of remote>`, eg. `git remote-rad origin rad://deadf4laf31`. The URL is a `LocalUrl`. Similarly, the transport `action` is also called with a `LocalUrl`.

librad/src/git/local/url.rs
kim: Because I didn't think it's too interesting -- the URL is supposed to be constructed programmatically via `LocalUrl`. Feel free to change.

librad/src/git/local/url.rs
kim: Sure ya... I used to dislike this pattern of early returns in Rust, too, but then I learned that type inference often fails when you do `let foo = { /* ... */ }?;`, which I find even more annoying. Feel free to change.

librad/src/git/local/transport.rs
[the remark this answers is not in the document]
FintanH: Right, I get ya

PR #233, opened 2020-07-03
Git remote helper

kim:
Builds on top of the local transport, allowing interaction with the storage
backend via the git CLI. Noteworthy is the use of git's credential helper
infrastructure to prompt for the keystore passphrase.

Note that it is not yet clear how to set HEAD (the default branch) for the
logical repository, rendering the clone without a checked-out working copy.
Configuring the remote tracking branches as actual remotes in the work tree is
not part of this patch either.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1413,10 +1413,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "radicle-git-helpers"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "futures",
+ "git2",
+ "libgit2-sys",
+ "librad",
+ "librad-test",
+ "radicle-keystore",
+ "tempfile",
+]
+
+[[package]]
 name = "radicle-keystore"
 version = "0.1.0"
-source = "git+https://github.com/radicle-dev/radicle-keystore?rev=a04f4dbaec4d05be2c7a516a36b5bbec02ee32aa#a04f4dbaec4d05be2c7a516a36b5bbec02ee32aa"
+source = "git+https://github.com/radicle-dev/radicle-keystore?rev=27c9a7b8ce5b87d2be3703f7a3290fa7a0552fc6#27c9a7b8ce5b87d2be3703f7a3290fa7a0552fc6"
 dependencies = [
+ "async-trait",
+ "futures",
  "lazy_static",
  "rpassword",
  "secstr",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-members = [ "librad", "librad-test", "radicle-tracker" ]
+members = [ "librad", "librad-test", "git-helpers", "radicle-tracker" ]
 
 [patch.crates-io]
 # Ed25519 support for rustls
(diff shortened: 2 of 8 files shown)


rudolfs:
<details>
<summary>Is it supposed to panic when run without any args?</summary>

```sh
radicle-link/git-helpers » cargo run                                                                                                                                                                                                                                    ‹wrk/4›
    Finished dev [unoptimized + debuginfo] target(s) in 0.14s
     Running `/Users/rudolfs/work/radicle-link/target/debug/git-remote-rad`
thread 'main' panicked at 'index out of bounds: the len is 0 but the index is 0', git-helpers/src/bin/remote/main.rs:43:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
</details>

kim:
> Is it supposed to ask for a password on `cargo test`? 

Good catch, setting up a decidedly interactive thing for non-interactive use is peculiar :)

> Is it supposed to panic when run without any args?

Error messages can be improved (I'll leave this to someone else), but you're not supposed to run the helper directly -- it's invoked by `git pull|push` under the hood.

kim:
Nb. there is not much we can do to customise the prompt of `git-credential` -- "Password for '<URL>'" is not too helpful when it is actually the passphrase for the keystore it is asking for. The wonderful and mannifold ways of the `git` suite allow for several options:

* Set `GIT_ASKPASS` or `core.askPass`

  This just executes whatever the value of those variables is, and writes the prompt it would normally display to it's stdin. 

* Make our own credential helper

  This is probably the most flexible solution, as the helper can delegate to agents / credential caches as it sees fit, and can be configured as desired. On the downside, it also _has_ to be configured, ie. the user has to run `git config`.

* Just open `/dev/tty`

  Instead of the git credential subsystem, the remote helper could also just do what the builtin git prompt does, ie. prompt on `/dev/tty` instead of fd 0. That is obviously the least flexible solution, as it prohibits non-interactive use, respectively would rely on not-well-known conventions if it were to delegate to agent / cache processes.

kim:
@FintanH can you rephrase the question?

FintanH:
> can you rephrase the question?

Sure, my question boils down to: Where will this executable run from?

kim:
> Where will this executable run from?

It needs to be in `$PATH` or `$GIT_EXEC_PATH` -- `git` is really a collection of executables (it `execve(2)`s all the time, even "internally"), each of which can be replaced. A remote helper works by `git` looking at the URL scheme of the `url` of a remote's config, and looking for an executable in those paths named `git-remote-SCHEME`. Look into `/usr/lib/git-core` -- you'll find remote helpers for `http`, `https`, `ftp`, and even `mediawiki`. Our scheme is `rad://`, so the executable is called `git-remote-rad`.

So, whenever you run `git push|pull|fetch|...`, `git-remote-rad` is run and, basically, runs a git server connected to `stdio` on demand.

kim:
Perhaps this is a little obscured in the test, because it uses a "typed" URL:

* To set up an existing repo for pushing/pulling to the link monorail, you'd run `git remote add whatever rad://<id of urn>`, and then use git like you always did.
* To check out an existing repo from the monorail into a new working repo, you'd run `git clone rad://<id of urn>` and it'd DWYM. Except I don't have a good solution for `HEAD` yet, as mentioned above.

FintanH:
Right, sorry you mentioned this before so thanks for the refresher :grin:

PR #236, opened 2020-07-03
Fire gossip announcement from git remote callback in e2e test

kim:
For demonstration purposes, mainly.

Unfortunately, I can't see a way to intercept `git-receive-pack` in such a way
that we could get at the pushed refs -- otherwise, the local transport could
handle to gossiping.

Also, async/await raises its ugly head again.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -57,9 +57,9 @@ checksum = "e742194e0f43fc932bcb801708c2b279d3ec8f527e3acda05a6a9f342c5ef764"
 dependencies = [
  "argh_shared",
  "heck",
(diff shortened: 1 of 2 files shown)


kim:
Could be made a tad bit more general, if it wasn't for the `libgit2` API being a bit awkward for setting up remotes. `Remote::push` with an empty refspec will consider the `push` refspec (not the `fetch` one) -- but to set one up you need to `Repository::remote`, `Repository::remote_add_push`, `Repository::find_remote`, because -- the almighty knows why -- we can't modify a `Remote`.

PR #237, opened 2020-07-06
Entity signatures refactor

massimiliano-mantione:
With this refactor entity signatures are not stored in a separated property.

Instead, the two  sets of owned keys and certifier URNs become maps that contain `Option`s that in turn contain the signatures.

This makes various checks simpler:

* it is not possible to sign the entity with a not owned key
* it is not possible to certify the entity with an unlisted certifier
* the async `sign` method disappears and is replaced by two synchronous methods that do not require resolvers:
  - `sign_owned` to sign with an owned key
  - `sign_by_user` to sign with a key belonging to a certifier
* checking that the entity is fully signed becomes a matter of iterating over those two maps, verifying that they do not contain `None` entries, and verifying every signature

Another change is that we now sign the entity hash and not the full content.
This speeds up verification because we verify the hash correctness only once (in `from_data`, just after deserializing), and we can trust it after that (it is immutable).
When verifying each signature we just use the hash (which we already have) instead of having to recompute the "canonical data" all the time.

This refactor is essentially the same thing as the 1st commit of PR #218 which @FintanH has already reviewed.
It has been extracted and rebased on top of master.

It is proposed here because we are considering whether moving signatures away from the entity blob and into the commit message (as commit trailers) or if we want to keep using this approach.

### DIFF
--- a/librad/src/git/storage/config.rs
+++ b/librad/src/git/storage/config.rs
@@ -25,10 +25,7 @@ use crate::{
     git::ext::is_not_found_err,
     internal::result::ResultExt,
     keys::SecretKey,
-    meta::{
-        entity::{Signatory, Verified},
-        user::User,
-    },
+    meta::{entity::Verified, user::User},
     peer::{self, PeerId},
     uri::{self, RadUrn},
 };
@@ -168,15 +165,15 @@ impl Config {
     }
 
     /// Validation rules as described for [`Config::set_user`]
-    pub fn guard_user_valid<S>(&self, user: &User<S>) -> Result<(), Error> {
+    pub fn guard_user_valid<S>(&self, user: &User<S>) -> Result<(), Error>
+    where
+        S: Clone,
+    {
         let peer_id = self.peer_id()?;
-        user.signatures()
+        user.keys()
             .get(peer_id.as_public_key())
-            .ok_or(Error::NotSignedBySelf)
-            .and_then(|sig| match sig.by {
-                Signatory::User(_) => Err(Error::OwnedKeyRequired),
-                Signatory::OwnedKey => Ok(()),
-            })
+            .ok_or(Error::NotSignedBySelf)?;
+        Ok(())
     }
 
     pub fn user(&self) -> Result<RadUrn, Error> {
@@ -242,6 +239,9 @@ mod tests {
         ))
     }
 
+    // FIXME: This test is impossible to write
+    // (the error condition is impossible with this refactor)
+    /*
     #[test]
     fn test_guard_user_unsigned() {
         let key = SecretKey::new();
@@ -253,9 +253,14 @@ mod tests {
             Err(Error::NotSignedBySelf)
         ))
     }
+    */
 
+    // FIXME: This test is impossible to write
+    // (the error condition is impossible with this refactor)
+    /*
     #[async_test]
     async fn test_guard_user_not_self_signed() {
+
         let key = SecretKey::new();
         let config = setup(&key);
 
@@ -263,12 +268,7 @@ mod tests {
         {
             let bob = User::<Draft>::create("bob".to_owned(), key.public()).unwrap();
             alice
-                .sign(
-                    &key,
-                    &Signatory::User(bob.urn()),
-                    &ConstResolver::new(bob.clone()),
-                )
-                .await
+                .sign_by_user(&key, &bob.clone().as_verified())
                 .unwrap();
         }
 
@@ -277,6 +277,7 @@ mod tests {
             Err(Error::OwnedKeyRequired)
         ))
     }
+    */
 
     #[async_test]
     async fn test_guard_user_valid() {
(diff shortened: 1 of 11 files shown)


librad/src/meta/entity.rs
```diff
@@ -381,108 +520,76 @@ where
...
-        let signature = EntitySignature {
-            by: Signatory::OwnedKey,
-            sig: self.compute_signature(key)?,
-        };
-        self.signatures.insert(public_key, signature);
+        self.keys
+            .get_mut(&public_key)
+            .ok_or_else(|| Error::KeyNotPresent(public_key.clone()))
+            .map(|opt| match opt {
+                Some(_) => Err(Error::SignatureAlreadyPresent(public_key)),
```
FintanH, line 587: I think since this is recoverable, it should just be returned as `Ok(Some(SignatureAlreadyPresent))`. The signature being there is not a fatal error to the function. Saw the idea here: http://sled.rs/errors.html, which made a lot of sense and this exact case came to mind.

librad/src/meta/entity.rs
```diff
@@ -493,82 +600,48 @@ where
...
             keys: self.keys,
             certifiers: self.certifiers,
-            signatures: self.signatures,
             info: self.info,
         }
     }
 
-    /// Compute the status of this entity (only this revision is checked)
+    /// FIXME: For tests, remove after we have a real verifier
+    pub fn as_verified(&self) -> Entity<T, Verified> {
```
FintanH, line 640: Add `#[cfg(test)]` if it's for tests.

librad/src/meta/entity.rs
```diff
@@ -581,35 +654,41 @@ where
...
-        } else {
-            Err(Error::SignatureMissing)
+
+        for (urn, s) in self.certifiers.iter() {
+            match s {
+                Some(sig) => {
+                    if !sig.verify(self.hash()) {
+                        return Err(Error::SignatureVerificationFailed(sig.key.to_owned()));
+                    }
+                    resolver
```
FintanH, line 680: It terrifies me that we could remove this and the code would still compile without any worry. There's a disconnect between the `urn` in the `sig` and the verification.

librad/src/meta/entity.rs
```diff
@@ -124,15 +130,70 @@ pub enum HistoryVerificationError {
...
         error: UpdateVerificationError,
     },
 }
 
+/// Type representing an entity revision
+pub type EntityRevision = u64;
+
+/// Timestamp for entities and signatures, as milliseconds from Unix epoch
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Deserialize, Serialize)]
+pub struct EntityTimestamp(i64);
```
FintanH, line 150: So currently the timestamp is just aesthetic, right? It's not used for any business logic right?

FintanH: It might be useful to separate it into its own module so as not to clutter up this one.

librad/src/meta/entity.rs
```diff
@@ -145,31 +206,113 @@ pub struct Signed;
...
+    }
+
+    pub(crate) fn verify(&self, key: &PublicKey, hash: &Hash, revision: EntityRevision) -> bool {
+        verify_signature(&self.sig, key, hash, revision, self.timestamp)
+    }
+}
+
+/// A signature for an `Entity`, when signed by a certifier
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+pub struct CertifierSignature {
```
FintanH, line 273: I'm not understanding the difference between `SelfSignature` and `CertifierSignature`. They look like they're built from the same data but `SelfSignature` doesn't hold onto the details. Why is there even a difference?

librad/src/meta/entity.rs
```diff
@@ -145,31 +206,113 @@ pub struct Signed;
...
+    pub revision: EntityRevision,
+    /// Key used by the certifier
+    pub key: PublicKey,
+    /// The signature data
+    pub sig: Signature,
+}
+
+impl CertifierSignature {
+    fn build(
+        key: &SecretKey,
```
FintanH, line 286: How would we even know a certifier's `SecretKey`?

librad/src/meta/entity.rs
```diff
@@ -330,7 +465,16 @@ where
...
+        self.to_data()
+            .clear_hash()
+            .clear_signatures()
+            .reset_timestamp()
+    }
+
+    /// Helper to build a new entity cloning the current one
+    /// (signatures are cleared because they would be invalid anyway)
+    pub fn prepare_next_revision(&self) -> EntityData<T> {
+        self.to_builder().set_parent(&self)
```
FintanH, line 477: So if I used `to_builder` without `set_parent` that would be an error, right?

librad/src/git/storage/config.rs
```diff
@@ -242,6 +239,9 @@ mod tests {
         ))
     }
 
+    // FIXME: This test is impossible to write
```
FintanH, line 242: I'm confused about this. What's going on here?

PR #241, opened 2020-07-08
Add refs_heads functions

### LINKED ISSUE #239: Get refs/heads of Peer
We have no way of asking for `remotes/<tracked_peer>/refs/heads/*` from the `Storage`/`Repo` API. We need this for listing a remote's branches, and maybe our own too :)

proposal add:
```rust
pub fn heads_of(urn: &RadUrn, peer: P) -> Result<???, Error>
```

Not sure what the `???` would be yet, maybe just the name of the reference.

FintanH:
Adding functions to retrieve the `refs/heads/*` references that would
live under a RadUrn for our local peer view and a remote peer view.

Fixes #239

### DIFF
--- a/librad/src/git/storage.rs
+++ b/librad/src/git/storage.rs
@@ -41,7 +41,7 @@ use crate::{
         p2p::url::{GitUrl, GitUrlRef},
         refs::{self, Refs},
         repo::Repo,
-        types::Reference,
+        types::{Reference, RefsCategory},
     },
     hash::Hash,
     internal::{
@@ -383,6 +383,32 @@ impl<S: Clone> Storage<S> {
         }?;
 
         Ok(Refs::from(signed))
+    }
+
+    /// Get the names of the references that live under
+    /// `namespaces/<urn-id>/refs/heads/*`.
+    pub fn refs_heads(&self, urn: &RadUrn) -> Result<Vec<String>, Error> {
+        self.refs_heads_of(urn, None)
+    }
+
+    /// Get the names of the references that live under
+    /// `namspaces/<urn-id>/refs/remotes/<peer-id>refs/heads/*`.
+    pub fn refs_heads_of<P>(&self, urn: &RadUrn, peer: P) -> Result<Vec<String>, Error>
+    where
+        P: Into<Option<PeerId>>,
+    {
+        let heads = Reference {
+            namespace: urn.id.clone(),
+            remote: peer.into(),
+            category: RefsCategory::Heads,
+            name: "*".to_string(),
+        };
+
+        let mut names = vec![];
+        for name in References::from_globs(&self.backend, &[heads.to_string()])?.names() {
+            names.push(name?.to_string());
+        }
+        Ok(names)
     }
 
     /// The set of all certifiers of the given identity, transitively
--- a/librad/src/git/storage.rs
+++ b/librad/src/git/storage.rs
@@ -41,7 +41,7 @@ use crate::{
         p2p::url::{GitUrl, GitUrlRef},
         refs::{self, Refs},
         repo::Repo,
-        types::{Reference, RefsCategory},
+        types::Reference,
     },
     hash::Hash,
     internal::{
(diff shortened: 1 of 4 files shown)


librad/src/git/storage.rs
```diff
@@ -385,6 +385,32 @@ impl<S: Clone> Storage<S> {
...
+        P: Into<Option<PeerId>>,
+    {
+        let heads = Reference {
+            namespace: urn.id.clone(),
+            remote: peer.into(),
+            category: RefsCategory::Heads,
+            name: "*".to_string(),
+        };
+
+        let mut names = vec![];
```
kim, line 407: Why allocate a `Vec` here? You can return the iterator if you do:

```rust
pub fn refs_heads_of<'a, P>(&'a self, urn: &RadUrn, peer: P) -> Result<impl Iterator<Item = Result<String, Error> + 'a, Error>
```

FintanH: I was running into lifetime issues but I think you're right it could work if I call `iter`

FintanH: Arrghhh, ya I'm still running into an issue that the `refs` that we get from `References::from_globs` will be dropped and its lifetime is tied to `ReferenceNames`.

But I suppose we can return `References` instead :)

kim: Look at `all_metadata` -- the opaque iterator needs to have the same lifetime as `&self` (via `+ 'a`). But yeah, a concrete type would also be fine.

FintanH: I think that would work if I wanted to return `Iterator<Item = Result<Reference<'a>, Error>>` but because I call `names` on the temporary value, returned by `References::from_globs`, it says that this will be dropped. I had the `+ 'a` on the `Iterator`, but it didn't seem to help.

kim:
I must insist that it would be preferable to define a single method returning a `References` iterator, which takes a `Reference`. We can control the visibility of the `Reference` constructors if we want to restrict which of them the library user is allowed to see. Reasoning is this:

* Don't pollute the surface API unnecessarily (who says someone doesn't need `refs/tags` or `refs/notes/` tomorrow?)
* Potentially easier testing (if `Reference` gains a `#[cfg(test)]` `create` method, it can be a single rountrip test covering all possible inputs)
* Unlike other methods on `Storage`, this doesn't deal with domain objects, so there's not argument to be had that we might want to add more internal transformations in the future

FintanH:
@kim: ah sorry! I misunderstood the request. That makes sense 👌 the only thing I'd like to possibly change is have a way of distinguishing between a Reference, that points to a single ref (if it exists), and a Reference that returns 0 or more refs (like ids/* and heads/*). Maybe a phantom witness or something. I'll get around to fixing this today. Sorry again for the misunderstanding 😅

kim:
> a way of distinguishing between a Reference, that points to a single ref (if it exists), and a Reference that returns 0 or more refs

Yes indeed, that would be awesome

FintanH:
This should be better @kim :ok_hand: I'm not 100% happy with the `Any` type witness but I used it to get around the shared vector when doing the fetch of `Refspecs`.

kim:
Nice!

> I'm not 100% happy with the Any type witness

May I suggest to use the "poor ~~rustacean's~~ man's GADTs" pattern?:

```rust
pub enum SomeReference {
    Single(Reference<Single>),
    Multiple(Reference<Multiple>),
}
```

Instead of throwing away the type information, the consumer could still reify the (effectful) methods at runtime.

FintanH:
> May I suggest to use the "poor ~rustacean's~ man's GADTs" pattern?:

Genius :ok_hand:

librad/src/git/types.rs
```diff
@@ -44,15 +45,47 @@ impl Display for RefsCategory {
...
+    /// The remote peer the reference might live under.
     pub remote: Option<PeerId>,
+    /// The category of the reference.
     pub category: RefsCategory,
+    /// The suffix path of the reference, e.g. `self`, `id`, `banana/pineapple`.
     pub name: String, // TODO: apply validation like `uri::Path`
+
+    /// Type witness for the cardinality this reference could point to, i.e. if
+    /// we are pointing to exactly one, zero or more, or both.
+    marker: PhantomData<N>,
```
kim, line 85: Interestingly, I remember clippy did want me to prefix such fields with an underscore... maybe because it thinks the method name `new` is special, and if the field is only used there, it must be unused?

FintanH:
> If it gets too verbose, we can add impl From<Reference<Single>> for SomeReference and impl From<Reference<Multiple>> for SomeReference

Ah, good point. I could feel that that's what I needed but my neuron didn't connect.

librad/src/git/types.rs
FintanH: Clippy :shrug:

PR #242, opened 2020-07-09
Upgrade to rustls 0.18

kim:
Pending release of quinn. Also, `cargo update`.


massimiliano-mantione:
To me it looks reasonable. Why is this a draft? :-)

kim:
Because we have to wait for `quinn` (the QUIC library) to make a new release which bumps their dependency on `rustls`. They don't re-export it, but we must ensure the same version of the crate is used here and by quinn.

PR #244, opened 2020-07-10
Introduced typed API for creating git symbolic refs (fixes #204).

### LINKED ISSUE #204: Extend Reference API
I was pointing out here: https://github.com/radicle-dev/radicle-link/pull/181#discussion_r444263609 that we could extend the notion of `Reference`s to also have `SymbolicReference`s. This could take named parameters to disambiguate the direction, and also we could avoid some boolean blindness :sunglasses:

massimiliano-mantione:
This is the most rust-idiomatic way of doing this I could think of.

Invocations could save one line if the `force` method also directly did the `build` and returned the Result but IMHO leaving it explicit is clearer (and less surprising).

### DIFF
--- a/librad/src/git/types.rs
+++ b/librad/src/git/types.rs
@@ -159,46 +159,74 @@ pub struct SymbolicReference<'a> {
 
 pub struct SymbolicReferenceSource<'a> {
     repo: &'a git2::Repository,
-    source: &'a Reference,
+    source: String,
 }
 
 pub struct SymbolicReferenceTarget<'a> {
     repo: &'a git2::Repository,
-    source: &'a Reference,
-    target: &'a Reference,
+    source: String,
+    target: String,
 }
 
 pub struct SymbolicReferenceForce<'a> {
     repo: &'a git2::Repository,
-    source: &'a Reference,
-    target: &'a Reference,
+    source: String,
+    target: String,
     force_value: bool,
 }
 
+/// Builder to create a symbolic reference on the given git repo
+///
+/// Look at the subsequent `SymbolicReferenceSource`, `SymbolicReferenceTarget`
+/// and `SymbolicReferenceForce` types to see the full API.
 impl<'a> SymbolicReference<'a> {
+    /// Create the builder
     pub fn new(repo: &'a git2::Repository) -> Self {
         Self { repo }
     }
 
+    /// Specify the symbolic reference source
     pub fn source(self, source: &'a Reference) -> SymbolicReferenceSource {
         SymbolicReferenceSource {
             repo: self.repo,
-            source,
+            source: source.to_string(),
         }
     }
 }
 
+/// Intermediate symbolic reference builder
 impl<'a> SymbolicReferenceSource<'a> {
-    pub fn target(self, target: &'a Reference) -> SymbolicReferenceTarget {
-        SymbolicReferenceTarget {
+    /// Specify the symbolic reference target
+    /// (checks that the ref exists and fails otherwise)
+    pub fn target(self, target: &'a Reference) -> Result<SymbolicReferenceTarget, git2::Error> {
+        target.find(self.repo)?;
+        Ok(SymbolicReferenceTarget {
             repo: self.repo,
             source: self.source,
-            target,
-        }
+            target: target.to_string(),
+        })
+    }
+
+    /// Specify the symbolic reference target using a `git2::Reference` that we
+    /// trust to exist (fails if its `name` is not a valid string)
+    pub fn target_ref(
+        self,
+        target: &'a git2::Reference,
+    ) -> Result<SymbolicReferenceTarget<'a>, git2::Error> {
+        Ok(SymbolicReferenceTarget {
+            repo: self.repo,
+            source: self.source,
+            target: target
+                .name()
+                .ok_or_else(|| git2::Error::from_str("Invalid target reference name"))?
+                .to_owned(),
+        })
     }
 }
 
+/// Intermediate symbolic reference builder
 impl<'a> SymbolicReferenceTarget<'a> {
+    /// Sets the `force` argument to use when creating the reference
     pub fn force(self, force: bool) -> SymbolicReferenceForce<'a> {
         SymbolicReferenceForce {
             repo: self.repo,
@@ -209,13 +237,16 @@ impl<'a> SymbolicReferenceTarget<'a> {
     }
 }
 
+/// Final symbolic reference builder
 impl<'a> SymbolicReferenceForce<'a> {
-    pub fn build(self) -> Result<git2::Reference<'a>, git2::Error> {
-        let name = self.source.to_string();
-        let target = self.target.to_string();
-        let message = format!("creating symref {} -> {}", name, target);
-        self.repo
-            .reference_symbolic(&name, &target, self.force_value, &message)
+    /// Creates the symbolic reference
+    pub fn create(self) -> Result<git2::Reference<'a>, git2::Error> {
+        self.repo.reference_symbolic(
+            &self.source,
+            &self.target,
+            self.force_value,
+            &format!("creating symref {} -> {}", self.source, self.target),
+        )
     }
 }
 
--- a/librad/src/git/types.rs
+++ b/librad/src/git/types.rs
@@ -217,7 +217,7 @@ impl<'a> SymbolicReference<'a> {
 /// Intermediate symbolic reference builder
 impl<'a> SymbolicReferenceSource<'a> {
     /// Specify the symbolic reference target
-    /// (checks that the ref exists and fails otherwise)
+    /// (`create` will fail if the reference does not exist)
     pub fn target(self, target: &'a Reference) -> SymbolicReferenceTarget {
         SymbolicReferenceTarget {
             repo: self.repo,
@@ -227,7 +227,8 @@ impl<'a> SymbolicReferenceSource<'a> {
     }
 
     /// Specify the symbolic reference target using a `git2::Reference` that we
-    /// trust to exist (fails if its `name` is not a valid string)
+    /// trust to exist
+    /// ('create' will fail if its `name` is not a valid string)
     pub fn target_ref(self, target: &'a git2::Reference) -> SymbolicReferenceTarget<'a> {
         SymbolicReferenceTarget {
             repo: self.repo,
--- a/librad/src/git/types.rs
+++ b/librad/src/git/types.rs
@@ -153,6 +153,72 @@ impl<'a> Into<ext::blob::Branch<'a>> for &'a Reference {
     }
 }
 
+pub struct SymbolicReference<'a> {
+    repo: &'a git2::Repository,
+}
+
+pub struct SymbolicReferenceSource<'a> {
+    repo: &'a git2::Repository,
+    source: &'a Reference,
+}
+
+pub struct SymbolicReferenceTarget<'a> {
+    repo: &'a git2::Repository,
+    source: &'a Reference,
+    target: &'a Reference,
+}
+
+pub struct SymbolicReferenceForce<'a> {
+    repo: &'a git2::Repository,
+    source: &'a Reference,
+    target: &'a Reference,
+    force_value: bool,
+}
+
+impl<'a> SymbolicReference<'a> {
+    pub fn new(repo: &'a git2::Repository) -> Self {
+        Self { repo }
+    }
+
+    pub fn source(self, source: &'a Reference) -> SymbolicReferenceSource {
+        SymbolicReferenceSource {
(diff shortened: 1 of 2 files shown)


massimiliano-mantione:
> I think this would be vastly simpler if we went with @kim's recommendation combined with one struct:

The problem with using `git2::Reference` for the source is that it implies that the source already exists, which is not what we want (I did it by mistake, tests started failing :facepalm:).

And the problem with using `git2::Reference` in general is that `reference_symbolic` takes two `&str`s, not the raw c strings, but a `git2::Reference` is not guaranteed to give you a valid `name` that you can then pass to `reference_symbolic` :-/

So what I did is just storing a `String` and requiring the user to provide:

* either a radicle typed references, and then check that it exists with `find`
* or a `git2::Reference` (if they did `find` the check themselves), but then I must check that it actually can provide a `name`

The result is #244 a7f724ff587789b0c8eb1949db7f416ae99fbd73

A single struct with public fields directly creatable as a literal at this point IMHO would be worse: the caller would not be able to specify the target either as a `git2::Reference` that is supposed to exist or as a radicle typed `Reference` whose existence must be checked (transparently).

FintanH:
> The problem with using git2::Reference for the source is that it implies that the source already exists, which is not what we want (I did it by mistake, tests started failing facepalm).

Ah good point. So I'm already getting them mixed up. `source` is the one that gets created or `target`? Is there an issue with requiring that the one that exists is `git2::Reference`? i.e. if it's `target` then:
```rust
/// The reference that already exists.
target: git2::Reference,
/// The one we are going to creating a symbolic link for.
source: Reference
```

> And the problem with using `git2::Reference` in general is that reference_symbolic takes two &strs, not the raw c strings, but a `git2::Reference` is not guaranteed to give you a valid name that you can then pass to reference_symbolic :-/

Hmm that's a good point. I hate this `Option<&str>` stuff. We could possibly add an error case where we couldn't get the ut8 name back (should be a rare occasion we ever see this).

> So what I did is just storing a String and requiring the user to provide:

I'd rather we stored the `types::Reference` than the `String`, if anything.

> A single struct with public fields directly creatable as a literal at this point IMHO would be worse: the caller would not be able to specify the target either as a git2::Reference that is supposed to exist or as a radicle typed Reference whose existence must be checked (transparently).

You could do these checks in `create`, where the effects are happening.

massimiliano-mantione:
> You could do these checks in `create`, where the effects are happening.

If the caller provided a radicle `Reference` the check is invoking `find` to see if it actually exists.
If the caller provided a `git2::Reference` there's no check (we presume it does exist), we just have this annoying thing that `name` gives an `Option<String>.

Moving the checks in `create` means remembering what must be checked (storing something that states it in the struct).
I can store a plain `String` (what's needed by `gt2::reference_symbolic`) because the check has already been performed, whatever it was.

massimiliano-mantione:
> I'd rather we stored the `types::Reference` than the `String`, if anything.

I understand, it embodies the semantics that the reference is guaranteed to exist.

But it also forces the "`name` is valid" check, and if the input were a typed `Reference` we'd already know that the string was valid.
With this approach we never run useless checks (and conversions: I suppose that `name` checks by attempting the conversion from the raw C string to the rust UTF8 `String`, which I do not want to run if it's not needed).

kim:
I see the difficulty with deferring actual creation of the symrefs until the `rad/id` ref actually exists, while at the same time guaranteeing that all the prerequisites are met. This could be solved by allowing to store a thunk in either side of the symref, which resolves to `Result<git2::Reference, git2::Error>`. Think:

```rust
pub enum SymrefPointer<'a, F>
where
    F: FnOnce() -> Result<git2::Reference<'a>, git2::Error>> 
{
    Immediate(git2::Reference<'a>),
    Deferred(F),
}
```

It's up to the caller to deal with any "transactional" semantics. In `create_repo`, we would build up the list of refs we require before committing the initial metadata as `Immediate`, and, if applicable, defer the self target until the commit has succeeded. If the `try_for_each` aborts with an error (eg. because said commit was unsuccessful), we nuke the entire namespace.

It would also be acceptable to punt on this, but then all we need is a datatype holding two `Reference`s and a `bool`, on which we can invoke `create<'a>(repo: &'a git2::Repository) -> Result<git2::Reference<'a>, git2::Error>`.

FintanH:
> Moving the checks in create means remembering what must be checked (storing something that states it in the struct).

I don't understand this point :thinking: What needs to be remembered?

What's wrong with something like this?:
```rust
struct SymbolicReference {
  source: Reference,
  target: Reference,
  force: bool,
}

    pub fn create(self, repo: &Repository<'a>) -> Result<git2::Reference<'a>, git2::Error> {
        self.source.find(repo).and_then(|exists| {
                self.repo.reference_symbolic(
                    &self.source.to_string(),
                    &self.target.to_string(),
                    self.force_value,
                    &format!("creating symref {} -> {}", self.source, self.target),
                )
            })
    }
```

massimiliano-mantione:
> ```rust
> struct SymbolicReference {
>   source: Reference,
>   target: Reference,
>   force: bool,
> }
> ```

This is storing radicle `Reference`s, which are not guaranteed to exist.
If the caller already had a `git2::Reference`, which is guaranteed to exist, this calls `find` anyway for not reason.

This is why @kim was proposing a thunk: the thunk would encode the knowledge about which check must be performed.

FintanH:
Ah, I didn't see that you meant if we _already_ have a `git2::Reference`. Is this an actual case in our code currently or just a pre-optimisation?

massimiliano-mantione:
> Ah, I didn't see that you meant if we _already_ have a `git2::Reference`. Is this an actual case in our code currently or just a pre-optimisation?

The current code never uses `git2::reference` but I added the case because @kim said that we should use `git2::reference` to ensure that the reference exists so I guessed it would be a case that must be handled properly.

I now pushed a version that is equivalent to @kim 's thunk idea but with the closures "unrolled" and implemented manually because storing actual closures like that, with variable sizes and all, and not going `dyn FnOnce` looked needlessly complex...

librad/src/git/types.rs
```diff
@@ -153,6 +153,118 @@ impl<'a> Into<ext::blob::Branch<'a>> for &'a Reference {
     }
 }
 
+pub struct SymbolicReference<'a> {
+    repo: &'a git2::Repository,
```
FintanH, line 157: I don't think this should take the `Repository`. We only need to know about a `Repository` when we create the symref. We should separate these concerns.

librad/src/git/types.rs
```diff
@@ -153,6 +153,118 @@ impl<'a> Into<ext::blob::Branch<'a>> for &'a Reference {
...
+/// Look at the subsequent `SymbolicReferenceSource`, `SymbolicReferenceTarget`
+/// and `SymbolicReferenceForce` types to see the full API.
+impl<'a> SymbolicReference<'a> {
+    /// Create the builder
+    pub fn new(repo: &'a git2::Repository) -> Self {
+        Self { repo }
+    }
+
+    /// Specify the symbolic reference source
+    pub fn source(self, source: &'a Reference) -> SymbolicReferenceSource {
```
FintanH, line 209: I think this building up of the structure is complex. I don't think we need 4 or 5 different `struct`s. We only need one that takes the arguments. Then `create` runs any of the checks. Whether we keep the `enum` for the two kinds of references or not.

librad/src/git/types.rs
```diff
@@ -153,6 +153,118 @@ impl<'a> Into<ext::blob::Branch<'a>> for &'a Reference {
...
+pub struct SymbolicReference<'a> {
+    repo: &'a git2::Repository,
+}
+
+pub struct SymbolicReferenceSource<'a> {
+    repo: &'a git2::Repository,
+    source: String,
+}
+
+enum SymbolicReferenceTargetArgument<'a> {
```
FintanH, line 165: I think this is optimising for a case we don't even have yet. So my feeling is that it's not necessary for now and we solve for the simplest case of have two `librad::Reference`s.

massimiliano-mantione:
> I don't think this building up of the structure is that useful. I'd rather one structure and one constructor.

I think the structure you want would be `The struct you are describing is `SymbolicReferenceForce` (with a different name), but...

> The constructor could even be a "smart constructor", i.e. the fields are hidden and the API (this one constructor) tells us we have created something that should be able to turn into a symbolic ref.

...this single constructor would have the problem of having two named arguments of the same type (source and target, both `Reference`), and a boolean (`force`), taking us full circle, back to square one.

I remember this all started because we did not want to have a single function with confusing arguments (you said that one never remembers which is which, and what does the bool mean).
A design with this single function was refused because it would have been ambiguous (rust does not have named arguments), while building the structure (with a literal) forces the caller to trite the property names and the ambiguity goes away.
Relying to the IDE to show the argument names was also refused: the request was that the caller has to write them so they would be explicitly shown in the code (like having named arguments).

But if now thew proposal is to have a _constructor_ for this class the invocation ambiguity would still be there!

I see that having a single struct makes for a simpler implementation.
It is basically the current `SymbolicReferenceForce`, renamed as `SymbolicReference`, and with both `source` and `target` as `types::Reference`.
Getting there would just require deleting some code from the current proposal.

Then, if we wanted to support calls that provide `git2::Reference` (which also seemed a requirement from previous comments) this would need one more rewrite...

kim:
Yeah the builder style is okay, although `pub` fields would also be (and less code). Let's not get fancy about the existence checks (apologies for suggesting to get clever here) -- I don't think this is currently implemented as desired, as the caller would want to control _both_ sides of the link. Passing the repository reference in `create` would be preferable, as we can pass around the value freely, and reason about where effects happen.

massimiliano-mantione:
> I don't think this is currently implemented as desired, as the caller would want to control _both_ sides of the link.

What does this mean?

The target is under control, and there are checks to make sure that it actually exists (no dangling links).

The source (the name of the ref to be created) is also specified (could not be otherwise).
The source could or could not already exist, and the `force` argument specifies whether overwriting it or returning an error if it already exists.

The only not covered combination is "return an error if the source does not already exist", but do we need it?

Or, likely, I am missing something.

kim:
`clone_repo` needs to maintain the invariant that the `refs/ids/*` exist for all certifiers (which are the source side of the symref). If `Fetcher::prefetch` collects and returns the refs it updated, we can keep it simple here (ie. not do any existence checks).

massimiliano-mantione:
There are two separated checks here:

> `clone_repo` needs to maintain the invariant that the `refs/ids/*` exist for all certifiers (which are the source side of the symref).

This IMHO is desirable but out of the scope of this brooming API PR.

The remote itself could keep an incomplete set of `refs/ids/*` (for buggy or malicious reasons).
After the fetch the local view of that remote `refs/ids/*` would simply reflect the remote set of refs.
Verifying the completeness of that set should be part of the verification code (to decide if that remote is trustworthy).
IMHO this completeness verification it is not relevant to an API to create a symbolic reference.

> If `Fetcher::prefetch` collects and returns the refs it updated, we can keep it simple here (ie. not do any existence checks).

This is probably doable using the `update_tips` callback in the fetch options.
As noted above there would be no guarantee that the resulting set is the complete set of required certifiers, it would just be the set of updated refs that are guaranteed to exist.

The `update_tips` callback provides the name of the updated reference (as `&str`), not a `git2::Reference`.
So, if I understand it correctly, the idea is to allow this "create symbolic ref" API to specify a target that is known to exist (to skip the existence check) using a `String` and not a `git2::Reference` (which _is_ guaranteed to exist).

Note that there is no easy way to parse a `String` (a full reference name, or path) into a `types::Reference`.
Keeping the builder pattern is would be easy to add an `existing_target(name: String)` method that specifies a source that is known to exist using its full name, and skips the existence check at `create`.

Is this what we want?
If yes, I can also make `Fetcher::prefetch` return the set of updated refs (as a set of strings) and then use this new `existing_target(name: String)` method to skip the checks.

kim:
> This IMHO is desirable but out of the scope of this brooming API PR.

Yes

> This is probably doable using the update_tips callback in the fetch options.

Yes

> The update_tips callback provides the name of the updated reference (as &str), not a git2::Reference.
So, if I understand it correctly, the idea is to allow this "create symbolic ref" API to specify a target that is known to exist (to skip the existence check) using a String and not a git2::Reference (which is guaranteed to exist).

I'm saying that I'd prefer the simpler over the safer API, given that the use sites do maintain their invariants. I'd also prefer a "dumb" datatype without any state (ie. reference to `git2::Repository`), so it can be used for entirely "symbolic" computations (ie. rendering a bunch of strings, for a config f.ex.). Adding a side-effecting method to actually _create_ the symref in a given repo, where the caller can specify whether or not to ensure that target and/or source already exist would be nice, but we don't actually have a use for it currently.

PR #246, opened 2020-07-14
Add 'refs' component to tracked 'remotes' path

FintanH:
According to the spec of the monorepo the path should take the form
`namespaces/$IDENTITY/refs/remotes/$TRACKED_PEER_ID/refs/...`. We were
missing the 'refs' portion which was causing some subtle issues. This
adds the 'refs' back and fixes the propagation test that was also
missing this portion.

### DIFF
--- a/librad/src/git/types.rs
+++ b/librad/src/git/types.rs
@@ -91,7 +91,7 @@ impl<N> Display for Reference<N> {
 
         match &self.remote {
             None => write!(f, "{}/{}", self.category, self.name),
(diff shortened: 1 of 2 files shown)


kim:
What exactly were the subtle issues this caused?

FintanH:
> What exactly were the subtle issues this caused?

I'll have to look closer. I was setting up some manual references in proxy in lieu of not having networking set up. So probably PBCAK.

I'm going to close this in favour of this https://github.com/radicle-dev/radicle-link/pull/247.

Issue #212, opened 2020-06-24
Event Streams: Define container format


kim:
Currently thought of as being mainly some "envelope" containing the signature of
the payload. Ie. potentially the same as #97

PR #248, opened 2020-07-16
Spec: Identities

kim:
[Rendered](https://github.com/radicle-dev/radicle-link/blob/spec/identities/docs/spec/identities.md)

### DIFF
--- a/docs/spec/identities.md
+++ b/docs/spec/identities.md
@@ -1,0 +1,427 @@
+---
+title: "Radicle Link -- Section: Identities"
+author:
+- Kim Altintop[^0]
+status: draft
+date: 2020-07-15
+
+numbersections: true
+link-citations: true
+bibliography: references.bib
+reference-links: true
+...
+[^0]: [kim@monadic.xyz](mailto:kim@monadic.xyz)
+
+# Identities
+
+In order to collaborate on repositories within a consensus-free network, we must
+be able to refer to them using a stable identifier. Note that this identifier is
+a **statement of intent**: a repository can be described as a collection of
+ever-moving leaves of a DAG whose root element is the empty object. Therefore,
+the _content_ of a repository is not enough to describe it -- while two views on
+the repository may share objects, they may diverge substantially otherwise. Both
+views may however state their intent to eventually converge to the same state.
+
+While in principle a random identifier with sufficient entropy would suffice for
+the purpose, this would put the burden of deciding which repository views are
+_legit_ (for some definition of legit) entirely on the user. Instead, our
+approach is to establish an ownership proof, tied to the network identity of a
+peer (`PeerId`), or a set of peers, such that repository views can be replicated
+according to the trust relationships between peers ("tracking").
+
+Our model is based on The Update Framework (TUF)[@tuf], conceived as a means of
+securely distributing software packages.
+
+## Data Model
+
+The identity of a repository is established using a document of the form:
+
+```rust
+struct Doc<T, D> {
+    replaces: Option<Revision>,
+    payload: T,
+    delegations: D,
+}
+```
+
+where:
+
+* `Revision` is a cryptographic hash of a document's contents, such that this
+  document is content-addressable by this hash within the storage system.
+* `replaces` refers to the previous revision of the document, or none if it is
+  the first revision
+* `payload` is an extensible, forwards- and backwards-compatible datatype
+  containing application-defined metadata about the repository. The protocol
+  interprets _some_ of the properties, as described in [TODO].
+* `delegations` contains the public keys of key owners who are authorised to
+  issue and approve new revisions of the document. The delegation format depends
+  on the type of identity being established, as detailed below.
+
+The `Doc` MUST be serialised in canonical form, e.g. [Canonical JSON].
+
+The authenticity of the `Doc` is captured by the following type:
+
+```rust
+struct Identity<T, D> {
+    id: ContentId,
+    root: Revision,
+    revision: Revision,
+    doc: Doc<T, D>,
+    signatures: HashMap<PublicKey, Signature>,
+}
+```
+
+where:
+
+* `id` is the content-addressable hash of the `Identity` itself
+* `root` is the initial revision of the identity `Doc`
+* `revision` is the current revision of the identity `Doc`
+* `doc` is the `Doc` described by `revision`
+* `signatures` contains signatures over the document history, indexed by the public
+  keys used. A signature is made over the cryptographic hash of the concatenation
+  of the `Revisions` chain, from most recent to the `root`.
+
+An `Identity` describes the attestation of a `Doc`s validity. Attestation chains
+MUST be stored as a hash-linked history (each `Identity` refers to its parent)
+-- we omit the parent `id` as a field here, as it is not used in the following
+sections.
+
+The `root` of a verified `Identity` is the stable identifier of the repository.
+
+> **TODO** Insert specification of URNs here
+
+## Verification
+
+Document revisions, as well as authenticity attestations form a hash-linked
+chain leading to the initial revision of the document. In order to verify the
+authenticity of a given identity attestation (`Identity`), the chain of updates
+must be contiguous and subject to the following verification rules.
+
+We can distinguish four levels of validity:
+
+1. **Untrusted**
+
+    The identity document is well-formed, and points to a root object which is
+    retrievable from local storage.
+
+2. **Signed**
+
+    The identity carrier passes 1., and is signed by at least one key specified
+    in the delegations of the document.
+
+3. **Quorum**
+
+   The identity carrier passes 2., and is signed by a quorum of the keys
+   specified in the [delegations](#delegations) of the document (`Q > D/2`).
+
+4. **Verified**
+
+    The identity carrier passes 3., and:
+
+    * The document does not refer to a previous revision, and no previous
+      revision is known
+    * **Or**, the set of signatures forms a quorum of the [delegations](#delegations)
+      of the **previous** revision.
+
+    It is an error if:
+
+    * No previous revision is given, but a parent in the identity attestation
+      chain is found
+    * A previous revision is given, but the identity attestation chain does not
+      yield a parent
+    * A previous revision is given, but it is not the same the parent
+      attestation refers to
+    * The current and parent attestations refer to different `root`s
+
+The verification process can now be described recursively:
+
+```rust
+/* State transitions, definitions elided for brevity */
+fn signed(i: Untrusted<Identity>) -> Result<Signed<Identity>, Error>;
+fn quorum(i: Signed<Identity>) -> Result<Quorum<Identity>, Error>;
+fn verified(i: Quorum<Identity>, parent: Option<Verified<Identity>>) -> Result<Verified<Identity>, Error>;
+
+fn verify(
+    head: Untrusted<Identity>,
+    mut parents: impl Iterator<Item = Untrusted<Identity>>,
+) -> Result<Verified<Identity>, Error> {
+    let head = quorum(signed(head)?)?;
+    let parent = match parents.next() {
+        None => Ok(None),
+        Some(parent) => verify(parent, parents),
+    }?;
+
+    verified(head, parent)
+}
+```
+
+> **Implementation Notes**:
+>
+> * The recursive definition is given for brevity, implementations may prefer to
+>   walk the history in reverse order.
+> * As an optimisation, implementations SHOULD store the result of a successful
+>   verification in persistent storage, such that verification of updates can
+>   start from the last-good state instead of the history root. Observe,
+>   however, that two different `Identity` objects (as identified by their
+>   `id`) may attest the same `Doc` revision -- the persistent state SHOULD
+>   therefore allow retrieving the set of attesting `Identity`s by `Doc`
+>   revision.
+> * The `quorum` predicate above SHOULD actually skip over this parent instead
+>   of aborting the procedure -- it is possible that a later `Identity` reaches
+>   the quorum.
+
+## Delegations
+
+Radicle Link distinguishes two types of identities: personal and project. The
+first describes an actor in the system, while the second describes a (software)
+project on which one or more actors collaborate. Apart from their payload types
+`T`, they differ in their delegations type `D`:
+
+Personal identities can only delegate to anonymous keys, while project
+identities MAY attach a personal identity  to a key delegation.
+
+More formally:
+
+```rust
+type User<T> = Identity<T, HashSet<PublicKey>>;
+
+enum ProjectDelegation<U> {
+    Key(PublicKey),
+    User(User<U>),
+}
+
+type Project<T, U> = Identity<T, ProjectDelegation<U>>;
+```
+
+Per identity document, the `PublicKeys` delegated to MUST form a set. It is, for
+example, an error if a `PublicKey` appears in both the document and a `User`
+delegations, or within two different `User` delegations.
+
+If a `Project` delegates to a `User`, it MUST do so by including a specific
+revision of the `User` document. The replication rules ensure that the
+respective `User` histories are replicated as sibling histories [insert link].
+
+If one of these sibling histories is found to not include the delegated-to
+revision in its ancestry path, it is said to have **forked**, and signatures
+made by one of the `User`s keys after the point of inclusion in the `Project` no
+longer count towards the quorum rules. Note that this is only recoverable if a
+quorum of valid keys remains on the project, otherwise a [Key Recovery](#key-recovery)
+procedure must be invoked on the `Project`.
+
+If a sibling history is NOT **forked** (i.e. it includes the attested revision
+in its ancestry path), AND it is **verified**, its key delegations are
+considered authoritative for all `Project` attestations between the point the
+`User` was delegated to, and the currently known-good head.
+
+> **Implementation Note**: If `User` keys are not reused (i.e. revoked and later
+> re-introduced) since the attestation point (which MUST be verified), it is
+> sufficient to check for key revocations using the latest known heads of the
+> respective histories.
+
+When calculating the **Quorum** or **Verifed** threshold, multiple signatures
+made by the set of keys of a `User` SHALL be counted as only one vote towards
+the quorum. This prevents unilateral decisions made by a single `User`. We
+consider this simple scheme sufficient for the purpose, but more sophisticated
+delegations may be supported in the future, such as customising the quorum
+threshold, or key roles.
+
+Note that a key revocation event in a sibling `User` history may render the
+project unusable if the remaining keys cannot form a quorum. Also note that the
+`Project` SHOULD renew the attestation from time to time.
+
+The `delegations` of a `Project` are also referred to as the project's
+**maintainers**.
+
+## Key Recovery
+
+> **TODO** Define how to delegate trust to a set of keys `K \ D` for the purpose
+> of recovering from key revocations in case the quorum rules can no longer be
+> met with the remaining keys. (i.e. "Social Recovery")
+
+## Effect on Replication
+
+Peers MUST NOT replicate repositories whose identities they are unable to
+verify. If an update to the identity cannot be verified, they SHOULD keep the
+data they already have, but refuse to update it. Peers SHOULD keep the identity
+histories regardless of their verification status.
+
+When **cloning** a repository, the attestation history according to the remote
+peer is fetched before all other repository contents, and the verification
+procedure is run on it. If this does not yield a verified status, the clone is
+aborted. The resulting repository state MUST include the attestation histories
+of at least a quorum of the delegates as per the remote peer's view of the
+identity document. In `git`, the claim that this will be the case can be
+determined before fetching the repository contents by examining the advertised
+remote refs. If these preconditions are not met, the clone is aborted, and
+already fetched data is pruned.
+
+When **fetching** (i.e. updating) a repository, a predetermined set of verified
+delegates is already known -- their views of the attestation history SHALL be
+fetched first, along with any advertised histories not yet present in the local
+repository (as those might be owned by future delegates). The fetching end MUST
+now find a valid `Doc` revision near the tips of the available delegate
+histories, and pick the most recent one. If two valid identities are found of
+which neither is in the ancestry path of the other, the attestation chain is
+said to have **forked**. In lieu of a consensus system, it is undecidable which
+side of the fork to commit to, and so replication of the attested repository
+SHALL refuse any further updates.
+
+> **TODO** Fork detection (i.e. pairwise compare if in same ancestry path, either
+> select most recent or flag as forked)
+
+## Git Encoding
+
+In the `git` implementation, a `Doc` corresponds to a `blob` object, stored as
+the single entry of a `tree` object, such that its name (acc. to the `tree`) is
+equal to the `blob` hash of the _initial_ version of the `Doc`, serialised in
+canonical form. That is:
+
+    let name = git hash-object -t blob doc.canonical_form()
+
+An `Identity` corresponds to a `commit` object.
+
+We map the fields as follows:
+
+```rust
+/* Simplified git object model */
+struct Commit {
+    id: Oid,
+    tree: Tree,
+    message: String,
+}
+
+struct Tree {
+    id: Oid,
+    entries: Vec<TreeEntry>,
+}
+
+struct TreeEntry {
+    id: Oid,
+    name: String,
+    object: BlobOrTree,
+}
+
+struct Blob {
+    id: Oid,
+    content: Vec<u8>,
+}
+
+/* Mapping (trivial type conversions elided) */
+let commit = /* .. */;
+let identity = Identity {
+    id: commit.id,
+    root: commit.tree.entries[0].name,
+    revision: tree.id,
+    doc: deserialize(first_blob(commit.tree).content),
+    signatures: from_trailers(commit.message),
+};
+```
+
+Where:
+
+* `first_blob` finds the first `TreeEntry` which is of type `blob`.
+* `deserialize` is implemented by a standard JSON parser. `User` delegations
+  from a `Project` are specified in the `Project`'s `Doc` as URNs, which are
+  resolved by parsing a `blob` object of the same name as the URN's `id` field
+  below the `tree` entry of type directory named `delegations`.
+* `from_trailers` interprets the commit message as per [git-interpret-trailers],
+  and extracts the signatures from trailers with the token `x-rad-signature`.
+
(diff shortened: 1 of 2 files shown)


docs/spec/identities.md
```diff
@@ -0,0 +1,438 @@
...
+We can distinguish four levels of validity:
+
+1. **Untrusted**
+
+    The identity document is well-formed, and points to a root object which is
+    retrievable from local storage.
+
+2. **Signed**
+
+    The identity carrier passes 1., and is signed by at least one key specified
```
massimiliano-mantione, line 108: We could omit the `Signed` state: the subsequent `Quorum` state subsumes it, and `Signed` seems to have no use on its own.

docs/spec/identities.md
```diff
@@ -0,0 +1,438 @@
...
+    in the delegations of the document.
+
+3. **Quorum**
+
+   The identity carrier passes 2., and is signed by a quorum of the keys
```
massimiliano-mantione, line 113: If we require just a quorum of signatures, how do we attest the fact that the user owns the newly introduced keys?
(or, in the case of a project, how do we attest that new maintainers do agree to being maintainers if they do not sign the identity)

Probably the rule should require:

- all signatures if the set of delegations changed
- a quorum of signatures if the set of delegations did not change

docs/spec/identities.md
```diff
@@ -0,0 +1,438 @@
...
+   specified in the delegations of the document (`doc.delegations.len() / 2`).
```
massimiliano-mantione, line 114: Should the quorum be `(x / 2) + 1` to ensure a majority?

docs/spec/identities.md
```diff
@@ -0,0 +1,438 @@
...
+
+```rust
+type User<T> = Identity<T, HashSet<PublicKey>>;
+type Project<T> = Identity<T, HashMap<PublicKey, Option<Urn>>>;
+```
+
+> **TODO** Consider a container for project delegations which does not require
+> to repeat the URN for every key
+
+The set of `PublicKey`s delegated to MUST be unique per identity document.
```
massimiliano-mantione, line 193: How can one produce a new revision of an entity with different metadata (the T payload) if one does not want to change the set of delegated keys?
This should be a common operation but it seems forbidden by this rule.

docs/spec/identities.md
```diff
@@ -0,0 +1,438 @@
...
+
+The purpose of the project-user link is to allow applications to display
```
massimiliano-mantione, line 195: I thought the purpose of the maintainer list was to establish who is trusted to operate on the project itself (not just on the project identity document but on the actual project).
Like, who can sign tags, or new releases... in other words, who is a trusted maintainer.
The ability to display additional metadata should be a side effect of knowing the set of maintainers.

docs/spec/identities.md
```diff
@@ -0,0 +1,438 @@
...
+`T`, they differ in their delegations type `D`:
+
+Personal identities can only delegate to anonymous keys, while project
+identities MAY attach a personal identity (by it's URN) to a key delegation.
+
+More formally:
+
+```rust
+type User<T> = Identity<T, HashSet<PublicKey>>;
+type Project<T> = Identity<T, HashMap<PublicKey, Option<Urn>>>;
```
massimiliano-mantione, line 187: This can be controversial.

Linking keys instead of URNs has the advantage of making the delegated keys explicit, but it has the disadvantage of forcing project updates whenever a delegated user revokes a key.

It also has the disadvantage that if a user creates a new device key (and adds it to their key set in their identity) that device key cannot be used to emit events in a project feed, even if the user were a project maintainer (through a key listed in the project delegates set).

In principle a maintainer should be a _user_ and not a key.

kim:
Sorry, I confused myself while getting in a hurry, and actually lost the key idea: so, the key observation is that we must casually relate project and user histories, but that is not possible without consensus UNLESS updates to a user would also include ALL projects the user participated in (and, strictly speaking, all the signatures she ever made). Which is undesirable for several reasons.

So, if the project must pin to a specific key set, it can include the user doc at a specific revision _as a blob_ named after that revision in a subdirectory. We can now treat the user history as canonical if it is both valid, and contains the blob name in its ancestry path.

docs/spec/identities.md
kim: I don't think this matters -- if a quorum agrees that they trust the new key, the key owner doesn't have to agree to that.

massimiliano-mantione:
> [...]
> Which is undesirable for several reasons.
 
This is clear.

> So, if the project must pin to a specific key set, it can include the user doc at a specific revision _as a blob_ named after that revision in a subdirectory.

Why including _as a blob_ and not just referring to the user identity attestation as a parent commit?

If one wants to verify the project entity one must also verify its maintainers, and to verify them one needs their full identity history.
So linking to that history as a parent will not cause more data to be fetched than what must be fetched anyway.

docs/spec/identities.md
kim: Why should this be a common operation? The **whole point** of this is to secure who can modify the metadata. If you want to separate those two, you're designing a PKI.

kim:
> Why including as a blob and not just referring to the user identity attestation as a parent commit?

Because both are equivalent, but the graph traversal becomes harder to implement using parents. There is virtue in simplifying the implementation burden.

docs/spec/identities.md
kim: There is also no harm in separating both transitions, and it immediately carries useful error information.

docs/spec/identities.md
massimiliano-mantione: Fair enough, this attests _trust_ of a key and not _ownership_ of a key, but it should suffice :smiley:

massimiliano-mantione:
> Because both are equivalent, but the graph traversal becomes harder to implement using parents. There is virtue in simplifying the implementation burden.

Would it be equivalent to include a property in the metadata stating "this is the user attestation" and containing the OID of the attestation commit?
Same info as the parent link but without the physical link, same info as the tree blob but without the tree entry at all.

docs/spec/identities.md
kim: > Modifying level 2 (like, changing the user email or name in the profile) should not require changing the set of keys at level 1.

It is completely irrelevant whether or not the keys change when the metadata changes. The point is that it is those keys which are allowed to change it, and so a new revision must be created in any case. What is the problem you're trying to solve?

kim:
> Would it be equivalent to include a property in the metadata stating "this is the user attestation" and containing the OID of the attestation commit?

Except that the availability guarantee for the user commit is weaker. What is the problem you're trying to solve? Saving one tree entry?

massimiliano-mantione:
> Except that the availability guarantee for the user commit is weaker.

True but if the whole attestation history is not available we have bigger problems so the availability of that single document should be irrelevant.

In practice the OID of the attestation should be enough to do a cache lookup and see if we already consider that attestation trusted, if the lookup fails we'd need the full attestation history anyway.

And the OID of the attestation should be the "official external reference" to it, so I wanted to use that and not the internal document blob pointer.

> What is the problem you're trying to solve? Saving one tree entry?

First of all checking that I understand how it should work.
Knowing that you consider them equivalent is a good step in confirming this understanding.

Then, yes, I consider a tree entry more expensive than a plain metadata property but this is not a problem in itself, the solutions are logically equivalent, both would work, that's fine.

docs/spec/identities.md
massimiliano-mantione: I was trying to understand this sentence:

```
The set of `PublicKey`s delegated to MUST be unique per identity document.
```

Does it mean that there cannot be two entity revisions sharing the same set of keys?
Or did I parse it incorrectly, and "identity document" means the stable URN? (the 1st revision of the feed of entity documents)

BTW, two different projects *could* delegate to the exact set of maintainers, therefore to the exact set of keys.
This is what I am not getting.

Anyway, this explanation:

> It is completely irrelevant whether or not the keys change when the metadata changes.

clarifies the point for me, and on this we agree.
It is that spec line that I am not understanding.

docs/spec/identities.md
```diff
@@ -0,0 +1,438 @@
...
+project on which one or more actors collaborate. Apart from their payload types
+`T`, they differ in their delegations type `D`:
+
+Personal identities can only delegate to anonymous keys, while project
+identities MAY attach a personal identity (by its URN) to a key delegation.
+
+More formally:
+
+```rust
+type User<T> = Identity<T, HashSet<PublicKey>>;
```
FintanH, line 186: Are there any rules for the addition/removal of these delegations? What I'm wondering is if it's possible if an actor could remove the `PublicKey`s and insert their own. Alternatively, they add enough keys to obtain a quorum.

Does this constitute as a fork? Are these scenarios even possible or maybe something here already prevents this?

docs/spec/identities.md
FintanH: > The set of `PublicKey`s delegated to MUST be unique per identity document.

I think all this is saying is that we can't have duplicated `PublicKey`s in the set of keys. This is clear when we read the pseudo-code `HashSet<PublicKey>`, but the RFC should apply to any language implementation.

massimiliano-mantione: > I think all this is saying is that we can't have duplicated `PublicKey`s in the set of keys.

Ah, that's ok.

But the wording is really misleading for me.
I mean, I cannot imagine a "set" with not unique elements, regardless of the programming language.
So I parsed it as "the whole set must be unique among different documents", and could not understand that at all :-/

Could we word it like this?
```
The set of `PublicKey`s delegated to in an identity MUST not contain duplicates.
```

docs/spec/identities.md
massimiliano-mantione: > Alternatively, they add enough keys to obtain a quorum.

The point is that we also check for a quorum of signatures from the _previous_ revision of the identity.
So it is impossible to create a new revision without controlling a quorum of the keys of the current one.

docs/spec/identities.md
```diff
@@ -0,0 +1,438 @@
...
+securely distributing software packages.
+
+## Data Model
+
+The identity of a repository is established using a document of the form:
+
+```rust
+struct Doc<T, D> {
+    replaces: Option<Revision>,
+    payload: T,
```
massimiliano-mantione, line 41: I'd really like to see a storage model where the `payload: T,` property of the "document" is split out in a separated blob.

The rationale is that radicle link is not concerned with that payload.

Putting it inside the document blob forces at least a json-compatible format on it, but most of all it forces radicle link to deal with that payload when deserializing the document metadata that it cares about, when that payload should be totally ignored (and possibly opaque).

Putting it in a separated blob would allow the application layer to evolve the `T` type without changing the radicle link code.

docs/spec/identities.md
```diff
@@ -0,0 +1,438 @@
...
+
+### Workflow
+
+Our construction of the `Identity` from a git commit allows for multiple `id`s
+to describe the same revision of the document (and thus be equally valid). This
+means that the respective delegates' histories may diverge in their _commit_
+histories, but still converge to an agreement on the validity of the attested
+document revision.
+
+While this allows for arbitrarily complex workflows (and history topologies), we
```
massimiliano-mantione, line 316: Would it be possible to specify a working workflow, and possibly mandate it?
Extensibility is good but leaving too much freedom in this direction can mean asking for trouble in the long run.

Dealing with the logical (or physical) merge of the commit histories of different devices is not trivial and we should provide a detailed spec of how this should work.

docs/spec/identities.md
```diff
@@ -0,0 +1,438 @@
...
+}
+```
+
+where:
+
+* `id` is the content-addressable hash of the `Identity` itself
+* `root` is the initial revision of the identity `Doc`
+* `revision` is the current revision of the identity `Doc`
+* `doc` is the `Doc` described by `revision`
+* `signatures` contains signatures over the `revision`, indexed by the public
```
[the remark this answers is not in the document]
kim: No no, just `sign(revision)`. The public key must be included in the encoding of the signature, because otherwise we can't verify it (`Signature::verify(&self, data: &[u8], pk: &PublicKey)`)

docs/spec/identities.md
```diff
@@ -0,0 +1,438 @@
...
+  keys used
+
+An `Identity` describes the attestation of a `Doc`s validity. Attestation chains
+MUST be stored as a hash-linked history (each `Identity` refers to its parent)
+-- we omit the parent `id` as a field here, as it is not used in the following
```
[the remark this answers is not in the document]
kim: The trick is that the chain of `Doc` revisions is separate from the chain of `Identity`s. That is, two different `commit`s can be valid for the same `Doc` revision, because the `tree` hash is the same.

docs/spec/identities.md
```diff
@@ -0,0 +1,438 @@
...
+struct Blob {
+    id: Oid,
+    content: Vec<u8>,
+}
+
+/* Mapping (trivial type conversions elided) */
+let commit = /* .. */;
+let identity = Identity {
+    id: commit.id,
+    root: commit.tree.entries[0].name,
```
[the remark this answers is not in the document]
kim: The name of the _initial_ blob is it's own hash (`Oid`). This way, we kind of force everyone to actually run the verification, instead of trusting a `root` field somewhere.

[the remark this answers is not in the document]
FintanH: Sorry, what I'm saying is that `commit.tree.entries[0].name` points to a `String` not an `Revision`/`Oid`, which seems like a mistake to me.

Is what you're saying the `name` points to the `Blob` and we want to use _its_ `Oid`?

docs/spec/identities.md
FintanH: Ah yes, good call :) Makes sense to me now :ok_hand:

docs/spec/identities.md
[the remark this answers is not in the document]
FintanH: So, if I'm understanding correctly, the "parent id" here is referring to the `Identity`'s parent id and not the `Doc`s?

docs/spec/identities.md
kim: I disagree: the extensibility and compatibility requirements exist in any case -- every application that wishes to put extra metadata there MUST be able to deal with payloads it doesn't know about. Instead of requiring every application to deal with a serialisation scheme themselves, we can devise a parser which simply preserves unknown portions of the payload as untyped `Value`s, forcing applications to always treat things they expect as optional fields. I don't see an advantage of a separate file for this.

FintanH: I believe the idea is that `payload: T` _is the blob_ contents. The fact the `T` is generic means the application layer can evolve the `T` however they want because they determine the `Deserialize` impls. Radicle link shouldn't care what's stored in there as long as it can de/serialise it.

docs/spec/identities.md
kim: I can detail it more, but it is no more complex than walking the history in topological order and skipping over commits which do not satisfy the **Quorum** rule. Recall that we are verifying the histories of the different `PeerId`s one-by-one, and have established a linear ancestry of _document_ revisions.

docs/spec/identities.md
massimiliano-mantione: > I don't see an advantage of a separate file for this.

Radicle link is only concerned with the `replaces` `delegations` properties from the `Doc`, which are tiny compared to `payload: T`.

The `replaces` and `delecations` properties are radicle link's business, their eventual evolution is radicle link's concern, but the `payload` is not so.

Splitting out the payload allows Radicle link to have:

* no need to parse it
* no need to even load it in memory
* no concern about its format at all, including
  - no need no need to play deserialization tricks to parse it in a free form way if its type is unknown, or
  - no need to know its type at all, or
  - no need to be updated every time Upstream evolves the `T` type if we implement deserialization in such a way that the payload structure must be known (which is the trivial way to implement it)

I would count these as advantages.

FintanH: I'm not sure your counter-proposal is clear to me. Can you expand on what splitting out the payload would look like?

docs/spec/identities.md
massimiliano-mantione: So, if I got it right, the idea is to have actual merges between the different devices' branches.

If we have two devices D1 and D2 with keys K1 and K2, a user could

- define the identity on D1, sign it on K1, now it is part of D1's branch history but without a quorum
- merge it from D1 into D2's branch, still no quorum
- sign it there with key K2, now there is a valid entity with a quorum
- merge it back into D1 so that also D1 can see the entity with the quorum of signatures

The histories of D1 and D2 would contain various merge commits.
When walking them backwards care should be taken to pick the right parent if one wants to stay on the same device branch and not switch into the other.

It should not be that hard because commits in D1 are signed with K1 (and K2 on D2) but this logic must be implemented.

Is this the workflow we are aiming to?

And BTW, how do we create a second device for a user U to add a new key?

User U has device D1 with key K1.
The second device is supposed to have K2 as key but K2 is still not part of the user identity U (it cannot be: the user identity U is on D1 but D1 cannot sign it with K2).

Device D2 is supposed to have U as "self identity" but user U does not have the device key K2, so we have the impasse of a device with a key, but its "self" identity does not have that key as trusted (yet), so it cannot communicate with others and be verified (can it?).

It is these kind of passages that we should specify.

docs/spec/identities.md
massimiliano-mantione: Let recap the two data structures in the current proposal and how they map to git contents (as I understood it):


The "document":
```
struct Doc<T, D> {
    /// Optional OID of the "parent" revision document
    replaces: Option<Revision>,
    /// Opaque application specific data
    payload: T,
    /// Roughly, the set of trusted keys (related to users in the case of projects)
    delegations: D,
}
```

The "identity"
```
struct Identity<T, D> {
    /// The OID of the identity revision commit
    id: ContentId,
    /// The OID of the identity revision document
    /// (pointed to by the tree)
    root: Revision,
    /// The OID of the tree referred by the identity commit
    revision: Revision,
    /// The document is referred to by the tree entry with name "root"
    /// (it also has OID "root")
    doc: Doc<T, D>,
    /// Commit trailers
    signatures: HashMap<PublicKey, Signature>,
}

```

So an identity is a commit pointing to a tree and containing signatures as commit trailers.

My simpler proposal would be to take the `replaces` and `delegations` properties away from the `Doc` and to put them:

1. either into another blob pointed to by the tree,
2. or inside other commit trailers or anyway some kind of encoding in the commit message text, in a proper canonical format.

Choice between 1 and 2 is an implementation detail, I'd have my preference but ultimately I care more about the separation than the actual encoding.

Going for 2 (so I don't have to define more tree entries and their naming) the identity encoding would be:

```
struct Identity<T, D> {
    /// The OID of the identity revision commit
    id: ContentId,
    /// Optional OID of the "parent" revision document, extracted from the commit message
    replaces: Option<Revision>,
    /// Roughly, the set of trusted keys (extracted from the commit message)
    delegations: D,
    /// The OID of the identity revision payload document
    /// (pointed to by the tree)
    root: Revision,
    /// The OID of the tree referred by the identity commit
    revision: Revision,
    /// The payload document is referred to by the tree entry with name "root"
    /// (it also has OID "root")
    /// (I put it here but it should be skipped and provided by a method)
    payload: Vec<u8>,
    /// Commit trailers
    signatures: HashMap<PublicKey, Signature>,
}
```

Struct `Doc` would disappear because its contents have been split between:

* the `Itentity::replaces` and `Identity::delegations` properties, and
* the application payload that is ignored by radicle link.

Depending on how we would handle merges in the workflow there could be a simpler encoding but this would be the basic step.

FintanH: Well, one issue I have with that is that not having a generic `payload` means we're constraining the type too much. We always have to interact with a `Vec<u8>`. If we pick a generic parameter, however, it gives us a degree of freedom for implementations and testing.

I'm also finding it hard to see your advantages here. In the above, all you've done is moved the `Doc` fields up. Thus, these are the same definitions (minus the generics).

massimiliano-mantione: > We always have to interact with a `Vec<u8>`.

This is exactly the point of my proposal: not having a type at the radicle link level.

The point is who is the "we".
"We" as "radicle link" will just ignore that blob, we are is just meant to give Upstream (or whatever app layer we'll serve) access to the payload.

Then "we" as "Upstream" will deal with the blob, exactly as radicle link deals with other blobs at its own level.
Having envelops providing opaque payloads to upper layers is how most protocols are structured.

massimiliano-mantione: One further explanation: soon we'll have to define how #212 works.
And there the idea is precisely to use some kind of envelop carrying opaque payloads.

There the need is obvious because we expect to have a lot of different event types: tying radicle link to those types, or even just making it aware of them, cannot possibly work.

What I am saying is that we should apply the same principle here.

FintanH: Apologies that I'm taking some time to see your vision :) I think I'm starting to see what you mean, however...

>  "We" as "radicle link" will just ignore that blob, we are is just meant to give Upstream (or whatever app layer we'll serve) access to the payload.

How and where does this happen? Link still has to store the blob of data and provide the hash for looking it up. Where does this go? 

> There the need is obvious because we expect to have a lot of different event types: tying radicle link to those types, or even just making it aware of them, cannot possibly work.

Can you explain how the current proposal is "tying radicle link to those types"? If the contents are generic in `T` we're not tying it to anything, at least that's my perspective.

docs/spec/identities.md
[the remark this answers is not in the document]
kim: Yes. The `id` is only introduced to clarify the identity of the `Identity` object (hah), and point out that two different `id`s can be equally valid of a given `revision` of a `Doc`.

docs/spec/identities.md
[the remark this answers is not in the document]
kim: I wrote "trivial type conversions elided" because I assumed that the previous definition of `let name` clarifies that the name is actually an `Oid` (the hash of the initial version of the blob). Perhaps at this point I should reconsider the choice of using rust syntax -- specifying this more precisely would get very verbose.

docs/spec/identities.md
kim: Massi, this is an unnecessary indirection, and your motivations don't justify it. Every application wishing to specialise `T` to custom payloads will have to deal with the fact that their payload may not actually be present, or that it was created by an earlier or later version of their software. We can make everyone's lives easier by using an extensible payload format (such as JSON-LD), which makes no guarantees, but preserves unknown schema parts in untyped form.

docs/spec/identities.md
kim: I have no idea why you think this is complicated, and I don't even understand your questions. The `tree` stays the same for every revision, it is only the signatures in a given commit which determine whether that revision is accepted -- if you know enough signatures over that tree, you can include all of them in a new commit to pass the rules. The identity (commit hash) of that commit doesn't matter, and there can exist more than one across the different maintainer branches. Iow, whether you merge or not is irrelevant, it is also irrelevant which parent comes first if you do.

I added a simple example as a recommendation, but as I wrote it can be made almost arbitrarily complicated.

docs/spec/identities.md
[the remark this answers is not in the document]
FintanH: I was able to get to the point that it was the blob's initial `Oid`, but thought it might have been a clerical error. Should be fine though :+1:

docs/spec/identities.md
```diff
@@ -0,0 +1,427 @@
...
+
+* Implementations SHOULD limit the download of untrusted identity histories to a
+  reasonable size in bytes.
+
+
+[git-interpret-trailers]: https://git-scm.com/docs/git-interpret-trailers
+[git-notes]: https://git-scm.com/docs/git-notes
+[radicle-registry]: https://github.com/radicle-dev/radicle-registry
+[subtree]: https://git-scm.com/docs/git-merge#Documentation/git-merge.txt-subtree
+[canonical JSON]: https://wiki.laptop.org/go/Canonical_JSON
```
FintanH, line 427:
```suggestion
[canonical JSON]: http://wiki.laptop.org/go/Canonical_JSON
```

Turns out the site is not `https`. I got a 404 when I clicked it

docs/spec/identities.md
```diff
@@ -0,0 +1,427 @@
...
+    root: Revision,
+    revision: Revision,
+    doc: Doc<T, D>,
+    signatures: HashMap<PublicKey, Signature>,
+}
+```
+
+where:
+
+* `id` is the content-addressable hash of the `Identity` itself
```
cloudhead, line 76: What is the significance of this?

FintanH: I believe the idea is that we should be able to fetch data by its hash, whether it's a `Doc` or and `Identity`. So the `id` here is the fetchable hash for `Identity`. In the **Git** section you can see that the `id` is the SHA identifier of the commit.

cloudhead: Okay, so doesn't really play a role in the protocol?

FintanH: I think it means if we picked another backing store, let's say sqlite, we would store these data inside using their hashes as the key.

That's what I gather, at least.

kim: There can be more than one `Identity` (= commit) which verifies a given `Doc` history (= tree + parent trees), so we SHOULD be able to retrieve the ones we already validated (by doc revision). This is significant in a system where consensus is a social construct — although it makes things less confusing if you assign a “merge manager” and keep both histories mostly aligned (as detailed in the “Workflow” section).

NunoAlexandre: How do we validate that a cached validation wasn't fabricated?

FintanH: I think if we turned the question on its head we might get an answer: Why would you cache a non-verified identity?

NunoAlexandre:
@kim, zero intention to bikeshed here: what is the reason to call these `Identities` instead of `Entities`? As someone new to the project, `Entities` for me _now_ stands for key pairs and the whole people/machine/keys problem domain.

I understand that we are concerned to how we _identify_ users and projects, but call `users` and `projects` as `Identities` seems confusing to me, where `Entities` would be clearer and present less overlap with other parts of the Radicle ecosystem.

Thoughts?

kim:
> what is the reason to call these Identities instead of Entities

Because identity describes exactly what this is about. An entity could be anything -- a stool, a pencil, a database table. If anything, we could be talking about the _subject_ of an identity, but that would be nothing more than the discriminator field when parsing a document into an enum of either user or project, and so is not relevant for this spec.

docs/spec/identities.md
kim: Pre-validated identities are obviously _not_ shared on the network. If you don't trust your own storage backend, you might consider to sign the validated-markers, but I'm not sure that'd be worth the trouble.

NunoAlexandre: I was thinking more of a bad actor violating the cache to bypass some validation. But it's clear now.

PR #254, opened 2020-07-23
Extract rad remote helper into lib

rudolfs:
For the [project checkout feature](radicle-dev/radicle-upstream#601) we'll need to package the rad-remote-helper together with Upstream.
For this I looked into the following options:

1. Forcing Upstream to build the `git-helpers` binary from its `radicle-link` dependency, so far seems to be impossible in Cargo:
    - https://github.com/rust-lang/cargo/issues/872
    - https://github.com/rust-lang/cargo/issues/748

2. Change Upstream proxy into a top-level Rust workspace and include `radicle-link` as a sub-component: not possible because `radicle-link` is already organized as a workspace and nested workspaces are not allowed:
    - https://github.com/nox/rust-rfcs/blob/master/text/1525-cargo-workspace.md#validating-a-workspace
    - https://stackoverflow.com/questions/49849878/how-to-deal-with-multiple-nested-workspace-roots

3. Make upstream build and package scripts do a `cargo install --git https://github.com/radicle-dev/radicle-link.git --rev 269a7ededafd7373ebfb03210770d0b3778002f`: blows up build times both for devs and on CI because build artefacts are not shared between proxy and the external cargo install.

4. Extracting the `git-remote-rad` functionality into a lib, so we can create a separate binary within Upstream that will be packaged along with proxy.

This PR implements option nr. 4.


git-helpers/src/lib.rs
```diff
@@ -18,3 +18,81 @@
 extern crate radicle_keystore as keystore;
 
 pub mod credential;
```
cloudhead, line 20: Instead of putting this code here, I'd suggest:
* Adding `pub mod remote_helper` here
* Creating a `remote_helper.rs` file and moving this code into that file
* Renaming the main function to `run`
* Invoking the function from main with `remote_helper::run()`

rudolfs: https://github.com/radicle-dev/radicle-link/pull/254/commits/1547c52c86e3d746a2403df07195f15406b0bb3b

PR #255, opened 2020-07-24
Signers Everywhere

### LINKED ISSUE #216: Use `Signer` trait everywhere
Nb. that the trait is `async_trait`, so that might change some signatures

FintanH:
Closes #216

### DIFF
--- a/librad/src/git/local/transport.rs
+++ b/librad/src/git/local/transport.rs
@@ -30,7 +30,7 @@ use crate::{
     git::{
         ext::{into_git_err, RECEIVE_PACK_HEADER, UPLOAD_PACK_HEADER},
         local::{self, url::LocalUrl},
-        storage::{self, Storage, WithSigner},
+        storage::{self, Signer, Storage},
     },
     keys,
     paths::Paths,
@@ -45,7 +45,7 @@ pub struct Settings<S> {
 
 pub fn register<S>(settings: Settings<S>)
 where
-    S: WithSigner,
+    S: Signer,
     S::Error: keys::SignError,
 {
     static INIT: Once = Once::new();
@@ -75,7 +75,7 @@ impl<S> LocalTransportFactory<S> {
 
 impl<S> SmartSubtransport for LocalTransportFactory<S>
 where
-    S: WithSigner,
+    S: Signer,
     S::Error: keys::SignError,
 {
     fn action(
@@ -235,7 +235,7 @@ pub struct LocalTransport<S> {
 
 impl<S> LocalTransport<S>
 where
-    S: WithSigner,
+    S: Signer,
     S::Error: keys::SignError,
 {
     pub fn new(settings: Settings<S>) -> Result<Self, Error> {
--- a/librad/src/git/local/transport.rs
+++ b/librad/src/git/local/transport.rs
@@ -30,7 +30,7 @@ use crate::{
     git::{
         ext::{into_git_err, RECEIVE_PACK_HEADER, UPLOAD_PACK_HEADER},
         local::{self, url::LocalUrl},
-        storage::{self, Storage, WithSigner},
+        storage::{self, Storage},
     },
     keys::SecretKey,
     paths::Paths,
@@ -231,7 +231,7 @@ impl Localio {
 
 #[derive(Clone)]
 pub struct LocalTransport {
-    storage: Arc<Mutex<Storage<WithSigner>>>,
+    storage: Arc<Mutex<Storage<SecretKey>>>,
 }
 
 impl LocalTransport {
--- a/librad/src/git/local/transport.rs
+++ b/librad/src/git/local/transport.rs
@@ -30,46 +30,36 @@ use crate::{
     git::{
         ext::{into_git_err, RECEIVE_PACK_HEADER, UPLOAD_PACK_HEADER},
         local::{self, url::LocalUrl},
-        storage::{self, Signer, Storage},
+        storage::{self, Storage},
     },
-    keys,
     paths::Paths,
+    signer::BoxedSigner,
     uri::RadUrn,
 };
+
+lazy_static! {
+    static ref SETTINGS: Arc<RwLock<Option<Settings>>> = Arc::new(RwLock::new(None));
+}
 
 /// The settings for configuring a [`LocalTransportFactory`] and in turn a
 /// [`LocalTransport`].
 #[derive(Clone)]
-pub struct Settings<S> {
+pub struct Settings {
     pub paths: Paths,
-    pub signer: S,
-}
-
-/// During testing we may want to be able to change settings globally. This type
-/// captures our settings in a [`RwLock`] so it can be accessed safely across
-/// multiple threads.
-pub type GlobalSettings<S> = Arc<RwLock<Option<Settings<S>>>>;
-
-impl<S> Settings<S> {
-    pub fn global(self) -> GlobalSettings<S> {
-        Arc::new(RwLock::new(Some(self)))
-    }
+    pub signer: BoxedSigner,
 }
 
 /// Register the local transport method to `git` so we can use our own custom
 /// transport methods. See [`LocalTransportFactory`] and its `SmartSubtransport`
 /// implementation for more details.
-pub fn register<S>(settings: GlobalSettings<S>)
-where
-    S: Signer,
-    S::Error: keys::SignError,
-{
+pub fn register(settings: Settings) {
     static INIT: Once = Once::new();
 
+    LocalTransportFactory::new().configure(settings);
     unsafe {
         INIT.call_once(move || {
             git2::transport::register(local::URL_SCHEME, move |remote| {
-                Transport::smart(&remote, true, LocalTransportFactory::new(settings.clone()))
+                Transport::smart(&remote, true, LocalTransportFactory::new())
             })
             .unwrap()
         });
@@ -77,21 +67,23 @@ where
 }
 
 #[derive(Clone)]
-struct LocalTransportFactory<S> {
-    settings: Arc<RwLock<Option<Settings<S>>>>,
+struct LocalTransportFactory {
+    settings: Arc<RwLock<Option<Settings>>>,
 }
 
-impl<S> LocalTransportFactory<S> {
-    fn new(settings: GlobalSettings<S>) -> Self {
-        Self { settings }
+impl LocalTransportFactory {
+    fn new() -> Self {
+        Self {
+            settings: SETTINGS.clone(),
+        }
+    }
+
+    fn configure(&self, settings: Settings) {
+        *self.settings.write().unwrap() = Some(settings)
     }
 }
 
-impl<S> SmartSubtransport for LocalTransportFactory<S>
-where
-    S: Signer,
-    S::Error: keys::SignError,
-{
+impl SmartSubtransport for LocalTransportFactory {
     fn action(
         &self,
         url: &str,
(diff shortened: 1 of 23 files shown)


librad/src/git/local/transport.rs
```diff
@@ -30,55 +30,54 @@ use crate::{
...
+        storage::{self, Storage, WithSigner},
     },
-    keys::SecretKey,
+    keys,
     paths::Paths,
     uri::RadUrn,
 };
 
-lazy_static! {
-    static ref SETTINGS: Arc<RwLock<Option<Settings>>> = Arc::new(RwLock::new(None));
```
FintanH, line 41: @kim Not sure if this was a safe change, but it seemed to work without everything exploding. Also tried calling `register` twice in the `propagation_test` and it seemed to be fine.

I remember you saying it was necessary for the transport registration, but can't remember why this setup exactly.

See what you think when you get back.

librad/src/git/storage.rs
```diff
@@ -145,7 +150,9 @@ pub enum RadSelfSpec {
 }
 
 pub type NoSigner = PhantomData<!>;
-pub type WithSigner = SecretKey;
+
+pub trait WithSigner: sign::Signer + Send + Sync + Clone + 'static {}
```
geigerzaehler, line 154: The name seems a bit strange. Why not just `Signer`?

librad/src/git/storage.rs
```diff
@@ -145,7 +150,9 @@ pub enum RadSelfSpec {
...
+impl WithSigner for SecretKey {}
```
geigerzaehler, line 155: To make the `WithSigner` trait usable out of the box with other `keystore` Signer implementations I’d suggest the following
```suggestion
impl<T: sign::Signer + Send + Sync + Clone + 'static> WithSigner for T {}
```

librad/src/keys.rs
```diff
@@ -51,6 +51,13 @@ fn ensure_initialised() {
     }
 }
 
+pub trait AsPKCS8 {
+    fn as_pkcs8(&self) -> Vec<u8>;
+}
+
+pub trait SignError: error::Error + Send + Sync + 'static {}
+impl SignError for Infallible {}
```
geigerzaehler, line 59: See comment above about blanket implementation.

librad/src/keys.rs
```diff
@@ -156,6 +175,27 @@ impl SecretKeyExt for SecretKey {
     fn metadata(&self) -> Self::Metadata {}
 }
 
+#[async_trait]
+impl sign::Signer for SecretKey {
```
geigerzaehler, line 179: The `keystore` crate already provides an implementation of `Signer` for `sodiumoxide` keys albeit a bit differently. We should either remove that implementation or use it here.

librad/src/git/refs.rs
```diff
@@ -99,6 +101,21 @@ where
...
+
+    #[derive(Debug, Error)]
+    pub enum Error {
+        #[error(transparent)]
+        Sign(#[from] Box<dyn error::Error + Send + Sync + 'static>),
+        #[error(transparent)]
+        Cjson(#[from] CjsonError),
+    }
+
+    impl keys::SignError for Error {}
```
geigerzaehler, line 116: Is this required?

librad/src/git/storage.rs
FintanH: Ya, I was sticking with naming. I think before it made sense because the mnemonic `Storage<WithSigner>` made sense.

FintanH: But it makes sense to change it :)

librad/src/git/storage.rs
FintanH: Ah yes! That was the missing piece I forgot. Thank you :grin:

librad/src/keys.rs
FintanH: I'm still going to have to `impl sign::Signer for SecretKey` because of the newtype here. Is this to ensure that we use the `Signer` implementation of the `sodiumoxide` keys?

librad/src/git/local/transport.rs
kim: Ya we need the global variable to be able to reconfigure at runtime, eg with different paths (for testing).

librad/src/keys.rs
geigerzaehler: You’re right, we need to implement `Signer` anyways. But ideally for the implementation we would just delegate to what `keystore` gives. This would require us to adjust the implementation in `keystore`. I think we can leave this for some other time.

FintanH: I did it anyways :) https://github.com/radicle-dev/radicle-keystore/pull/4

librad/src/git/local/transport.rs
FintanH: My thoughts are that a global variable would be the same as exposing the Arc + RwLock (changes here https://github.com/radicle-dev/radicle-link/pull/255/commits/07afd1ff6033e8e0b54e1112b778905ccf11a61c). The only thing is that it's exposed to the caller. But I'm not sure how else to get around this :/

kim: It’s not the same — registering the transport with libgit2 happens only once, and is stored in a static in C. So when the transport gets invoked, it’s on the struct that is stored at that location — creating and modifying a new LocalTransport has no effect.

FintanH: Well bollix... I tried all the `Box<dyn Signer>` magic I, and others, could think of without making the compiler :weary: I'm not sure how to get a statically know `Signer` without choosing a concrete type. The only concrete type being the `SecretKey`.

kim: How so? A Box<dyn > works fine here: https://github.com/radicle-dev/radicle-link/blob/master/librad/src/git/p2p/transport.rs#L68

FintanH: I can write up a commit to show you tomorrow.

But it boils down to needing `Clone` for `Storage` (because of the `TryCow` stuff). And then our `dyn Signer<Error = Box<dyn StdError>` isn't cloneable.

kim: Ah I see. rustc clearly wins the award for most misleading type errors.

Storage doesn’t care which signer its type parameter is instantiated to, only if there is one (hence the name WithSigner, could also be SomeSigner or AnySigner). If I’m not mistaken, you should be able to instantiate S to either Box<dyn Signer + Clone>, or a newtype around that. A Signer should always be Clone, so perhaps we can add that as a bound on the trait itself.

FintanH: Ah heh, right the second pain in the ass is that you can't do Box<dyn Signer + Clone> because it gives out about not being allowed to have more than one non-auto trait.

FintanH: Seems like we're pretty fucked here. So I started fleshing out the `Box` approach again. To recap we have the following blanket trait:
```rust
pub trait Signer: sign::Signer + Send + Sync + Clone + 'static {}
```

So I went back to `local/transport.rs` and changed `LocalTransport` to:
```rust
type BoxedSigner = Box<dyn Signer<Error = Box<dyn std::error::Error + Send + Sync + 'static>>>;

#[derive(Clone)]
pub struct LocalTransport {
    storage: Arc<Mutex<Storage<BoxedSigner>>>,
}
```

To which I get the following error:
```
error[E0038]: the trait `git::storage::Signer` cannot be made into an object
   --> librad/src/git/local/transport.rs:236:5
    |
236 |     storage: Arc<Mutex<Storage<BoxedSigner>>>,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `git::storage::Signer` cannot be made into an object
    | 
   ::: librad/src/git/storage.rs:154:48
    |
154 | pub trait Signer: sign::Signer + Send + Sync + Clone + 'static {}
    |           ------                               ----- ...because it requires `Self: Sized`
    |           |
    |           this trait cannot be made into an object...

error: aborting due to previous error

For more information about this error, try `rustc --explain E0038`.
```

I look into the error code from `rustc --explain E0038` and this is the first snippet:
```
Trait objects like `Box<Trait>` can only be constructed when certain
requirements are satisfied by the trait in question.

Trait objects are a form of dynamic dispatch and use a dynamically sized type
for the inner type. So, for a given trait `Trait`, when `Trait` is treated as a
type, as in `Box<Trait>`, the inner type is 'unsized'. In such cases the boxed
pointer is a 'fat pointer' that contains an extra pointer to a table of methods
(among other things) for dynamic dispatch. This design mandates some
restrictions on the types of traits that are allowed to be used in trait
objects, which are collectively termed as 'object safety' rules.

Attempting to create a trait object for a non object-safe trait will trigger
this error.

There are various rules:

### The trait cannot require `Self: Sized`

When `Trait` is treated as a type, the type does not implement the special
`Sized` trait, because the type does not have a known size at compile time and
can only be accessed behind a pointer. Thus, if we have a trait like the
following:

```
trait Foo where Self: Sized {

}
```

We cannot create an object of type `Box<Foo>` or `&Foo` since in this case
`Self` would not be `Sized`.

Generally, `Self: Sized` is used to indicate that the trait should not be used
as a trait object. If the trait comes from your own crate, consider removing
this restriction.
```

Right, I don't have a sized on `Signer`, but guess who does?! [`Clone`](https://doc.rust-lang.org/src/core/clone.rs.html#110-134) does.

FintanH:
Putting this back into Draft until we come to a conclusion on https://github.com/radicle-dev/radicle-link/pull/255#discussion_r460120412

FintanH:
Note to self: I'll need to change the `git-helpers` once we resolve the `LocalTransport` issue.

librad/src/git/local/transport.rs
kim: Ach ach, das kompilern von die kostet-nix abstrakten ist nicht vor die gefingerpoken und mittengrabben! Gefahr von das hidden cost!

I’m not sure why you need to tell it at all what the associated error type is, but we can change `keystore::Signer` as we like, no? Just slap `Sized + Send + Sync` there (it will have to have those bounds anyway), and refine to `Clone` here. Does that not work?

FintanH: > I’m not sure why you need to tell it at all what the associated error type is

Der Compiler fragt danach

> but we can change keystore::Signer as we like, no? Just slap Sized + Send + Sync there (it will have to have those bounds anyway)

Do you mean in `radicle-keystore`? I dunno how that helps? 

The above explanation, as far as I'm understanding, says that any trait with `Self: Sized` can't be packed as a `Box<dyn Trait>`. We want a `Box` because of this global variable. We want a `Clone` because of the `Store` And `Clone` wants a `Self: Sized`. If we pass these conditions into the Galois SMT solver I think it won't be solveable :upside_down_face:

kim: Sometimes it seems they’re just making shit up, and then forget why they ended up there.

I believe it might work to use `ToOwned` instead of `Clone` for `S`.

FintanH: `ToOwned` wasn't going to work either because then we would need to say what `Owned` would be.

The latest [commit](https://github.com/radicle-dev/radicle-link/pull/255/commits/a081b64d7cfca03626344bb98e1bcd56335f5e1a) works with a double-heelflip-body-varial-to-revert, but it depends on this change https://github.com/radicle-dev/radicle-keystore/pull/5.

Please let this saga be over...

FintanH: Turns out DynClone was enough.

NOW, let this saga be over.

xla:
> Note to self: I'll need to change the `git-helpers` once we resolve the `LocalTransport` issue

Should we put this in a follow-up issue?

xla:
> Should we put this in a follow-up issue?

It looks like it's already addressed.

PR #258, opened 2020-07-27
Check Has is what we Want

### LINKED ISSUE #252: Has does not equality check
The `Has` function does not actually check the equality of what we `Want`ed:
https://github.com/radicle-dev/radicle-link/blob/master/librad/src/net/protocol.rs#L192-L200

FintanH:
Fixes #252 

When we query the network for some `A` we get back a stream of `Has of
A`. We never checked that it indeed was the `A` we wanted. This change
introduces the check to see if they're equal.

### DIFF
--- a/librad/src/net/gossip.rs
+++ b/librad/src/net/gossip.rs
@@ -105,6 +105,13 @@ pub struct Has<A> {
     val: A,
 }
 
+impl<A> Has<A> {
+    /// Look at the value of the `Has`.
+    pub fn val(&self) -> &A {
+        &self.val
+    }
+}
+
 #[derive(Debug, Clone)]
 pub struct MembershipParams {
     /// Maximum number of active connections.
--- a/librad/src/net/gossip.rs
+++ b/librad/src/net/gossip.rs
@@ -28,6 +28,7 @@ use std::{
     iter,
     marker::PhantomData,
     num::NonZeroU32,
+    ops::Deref,
     sync::{
         atomic::{self, AtomicBool},
         Arc,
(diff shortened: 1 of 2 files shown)


librad/src/net/gossip.rs
```diff
@@ -105,6 +105,13 @@ pub struct Has<A> {
     val: A,
 }
 
+impl<A> Has<A> {
+    /// Look at the value of the `Has`.
+    pub fn val(&self) -> &A {
+        &self.val
+    }
+}
+
```
cloudhead, line 114: Could also be implemented as `Deref` :shrug:

librad/src/net/protocol.rs
```diff
@@ -189,11 +189,14 @@ where
...
     /// requested state.
-    pub async fn query(&self, want: A) -> impl futures::Stream<Item = gossip::Has<A>> {
-        self.gossip.query(want).await;
-        self.gossip.subscribe().await.filter_map(|evt| async move {
+    pub async fn query(&self, want: A) -> impl futures::Stream<Item = gossip::Has<A>>
+    where
+        A: PartialEq + Clone,
+    {
+        self.gossip.query(want.clone()).await;
+        self.gossip.subscribe().await.zip(futures::stream::repeat(want)).filter_map(|(evt, wanted)| async move {
```
cloudhead, line 197: Curious why you replicate `want` instead of checking the `has` with a captured `want`? Is it because it's not `Sync + Send`?

FintanH: See rant in IRC 😬

PR #263, opened 2020-08-05
Determine remote identity when TLS session is resumed

kim:
When a TLS session is resumed (which is enabled by default), rustls doesn't
remember the presented certificate, and doesn't provide a way to attach
additional session persistence information. However, in the case of incoming
connections, we can still fall back to the SNI (relying on the cert verifier to
require SNI). Outgoing connections know the SNI already, so they can use that
by the same reasoning.

### DIFF
--- a/librad/src/net/quic.rs
+++ b/librad/src/net/quic.rs
@@ -34,15 +34,26 @@ use crate::{
         connection::{self, CloseReason, LocalInfo, RemoteInfo},
         tls,
     },
-    peer::PeerId,
+    peer::{self, PeerId},
 };
 
 #[derive(Debug, Error)]
 pub enum Error {
+    #[error("Remote PeerId could not be determined")]
+    RemoteIdUnavailable,
+
+    #[error(transparent)]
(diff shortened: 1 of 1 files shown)


kim:
@xla This came up in @cloudhead + @FintanH 's attempts to make the on-demand git connections work (see [here](https://github.com/radicle-dev/radicle-link/compare/cloudfin/connect-git). The problem, however, is not to test that it works, but that it fails if you mess with the certificate verifier -- we can only trust the SNI if that is implemented correctly. I cannot think of a way to link both layers.

xla:
> @xla This came up in @cloudhead + @FintanH 's attempts to make the on-demand git connections work (see [here](https://github.com/radicle-dev/radicle-link/compare/cloudfin/connect-git)). The problem, however, is not to test that it works, but that it fails if you mess with the certificate verifier -- we can only trust the SNI if that is implemented correctly. I cannot think of a way to link both layers.

Schade, thanks for the additional context.

PR #265, opened 2020-08-05
Add BranchRef

FintanH:
We had no way of talking about references that live under `refs/heads` or
`refs/remotes`. This is useful for setting up remote fetch/push specs.

I chose a separate type rather than trying to shoe-horn it into the
existing model.

### DIFF
--- a/librad/src/git/types.rs
+++ b/librad/src/git/types.rs
@@ -21,440 +21,39 @@ use std::{
     marker::PhantomData,
 };
 
-use crate::{
-    git::{ext, refs::Refs},
-    hash::Hash,
-    peer::PeerId,
-    uri::RadUrn,
-};
+use crate::{git::refs::Refs, peer::PeerId, uri::RadUrn};
 
-pub type Namespace = Hash;
+pub mod existential;
+pub mod reference;
 
-#[derive(Clone, Copy, Debug, PartialEq)]
-pub enum RefsCategory {
-    Heads,
-    Rad,
-}
+pub use existential::{SomeNamespace, SomeReference};
+pub use reference::{Multiple, Namespace, Reference, RefsCategory, Single};
 
-impl Display for RefsCategory {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match self {
-            Self::Heads => f.write_str("heads"),
-            Self::Rad => f.write_str("rad"),
-        }
-    }
-}
-
-/// Type witness for a [`Reference`] that should point to a single reference.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub struct Single;
-
-/// Type witness for a [`Reference`] that should point to multiple references.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub struct Multiple;
-
-#[derive(Debug, Clone, PartialEq)]
-pub enum SomeReference {
-    Single(Reference<Single>),
-    Multiple(Reference<Multiple>),
-    BranchMultiple(BranchRef<Multiple>),
-    BranchSingle(BranchRef<Single>),
-}
-
-impl Display for SomeReference {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match self {
-            Self::Single(reference) => write!(f, "{}", reference),
-            Self::Multiple(reference) => write!(f, "{}", reference),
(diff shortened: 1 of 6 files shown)


FintanH:
It's kind of annoying that I'm repeating some of the logic, but I can't currently think of a way to talk about their similarities.

And what's really annoying is that we have no way of saying something like `spec(local: F<N>, remote: G<N>)` for creating the specs.

FintanH:
> Can we not just lift the presence or absence of a namespace to the type-level?

Ah, the use of `!` is good! I was also thinking of abstracting over the `Remote`, because something like `"origin"` isn't a `PeerId`.

librad/src/git/types.rs
```diff
@@ -57,13 +57,208 @@ pub struct Multiple;
...
+    pub remote: Option<String>,
+    pub name: String,
+    marker: PhantomData<N>,
+}
+
+impl<N> Display for BranchRef<N> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match &self.remote {
+            None => write!(f, "refs/heads/"),
+            Some(remote) => write!(f, "refs/remotes/{}/", remote),
```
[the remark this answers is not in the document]
FintanH: There's no `heads` under the `remotes` portion. It just continues on with branch names.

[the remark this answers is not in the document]
kim: Well, there isn't in a standard `git` repo, but there is in a `link` monorepo -- it's `refs/remotes/*/{heads,rad}`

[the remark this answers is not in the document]
FintanH: Ya, so I'm trying to have something available for both. So when we set up the `rad` remote (currently in proxy) we have something like:
```
[remote "rad"]
        url = file:///home/haptop/.local/share/radicle/git/
        fetch = +refs/namespaces/hwd1yreg3gj1znurj8ha3z75nd1ca5zpxoyfricw8s81n1th5rsqewmzw1r/refs/heads/*:refs/heads/*
        push = +refs/heads/*:refs/namespaces/hwd1yreg3gj1znurj8ha3z75nd1ca5zpxoyfricw8s81n1th5rsqewmzw1r/refs/heads/*
```

For the case of `refs/remotes/*/{heads,rad}` that would be covered in the `ElaborateRef`.

FintanH:
This is now Reasonably Polymorphic:tm:

librad/src/git/types.rs
```diff
@@ -21,73 +21,34 @@ use std::{
...
-    pub category: RefsCategory,
-    /// The suffix path of the reference, e.g. `self`, `id`, `banana/pineapple`.
-    pub name: String, // TODO: apply validation like `uri::Path`
-
-    /// Type witness for the cardinality this reference could point to, i.e. if
-    /// we are pointing to exactly one, zero or more, or both.
-    marker: PhantomData<N>,
-}
+/// A representation of git reference that is under `refs/namespace/<namespace>`
+pub type NamespaceRef<N> = Reference<Namespace, PeerId, N>;
```
kim, line 47: NamespacedRef?

librad/src/git/types/reference.rs
```diff
@@ -0,0 +1,280 @@
...
+
+    pub(super) _namespace: Namespaced,
+    pub(super) _cardinality: PhantomData<Cardinality>,
+}
+
+// Polymorphic definitions
+impl<Namespaced: Clone, R: Clone, N: Clone> Reference<Namespaced, R, N> {
+    /// Set the remote portion of thise reference.
+    ///
+    /// Note: This is consuming.
```
kim, line 69: Uh-oh, idiotmatic polizei instructed me: `set_x` is `&self -> ()`, `with_x` is `self -> Self`, and builder-style `x` is `&mut self -> &mut Self` :man_shrugging:

librad/src/git/types.rs
FintanH: Sure, why not!

librad/src/git/types/reference.rs
FintanH: Well, they should have apprehended you because I didn't change the naming just the location :stuck_out_tongue_closed_eyes: 
I'll change it while I'm shaving yaks :water_buffalo:

kim: Ouch :handcuffs:

PR #268, opened 2020-08-06
Use address hints to establish (git) connections if applicable

kim:
I ended up treating this as a proper connect, ie. add the connection to the pool, and start gossiping, because of the following:

Due to a bug in either `libgit2` or `git2-rs`, or just the way the Rust FFI works, we can't register our transport as "stateful". This means that `libgit2` will always try to open _two_ connections, one for the refs advertisement, and another one for getting the packfile. Always helpful, it will also call `close()` on our transport, so there isn't a way to know when it's done.

Although we don't incur a full TLS handshake thanks to session tickets, the connection establishment is still roughly equivalent to a TCP handshake.

Otoh, it is also semantically _correct_ to keep being connected, because if we got what we wanted from the peer, we are also interested in updates to it. This is a bit handwavy I know, but I think good for now.

### DIFF
--- a/librad/src/net/protocol.rs
+++ b/librad/src/net/protocol.rs
@@ -28,7 +28,7 @@ use std::{
 };
 
 use futures::{
-    future::TryFutureExt,
+    future::{BoxFuture, TryFutureExt},
     lock::Mutex,
     stream::{BoxStream, StreamExt, TryStreamExt},
 };
@@ -55,7 +55,7 @@ use crate::{
 #[derive(Debug, Clone)]
 pub enum ProtocolEvent<A> {
     Connected(PeerId),
-    Disconnected(PeerId),
+    Disconnecting(PeerId),
     Listening(SocketAddr),
     Gossip(gossip::Info<IpAddr, A>),
 }
@@ -105,10 +105,14 @@ pub enum Error {
     Io(#[from] io::Error),
 }
 
+pub type RunLoop = BoxFuture<'static, ()>;
+
 #[derive(Clone)]
 pub struct Protocol<S, A> {
     gossip: gossip::Protocol<S, A, IpAddr, quic::RecvStream, quic::SendStream>,
     git: GitServer,
+
+    endpoint: quic::Endpoint,
 
     connections: Arc<Mutex<HashMap<PeerId, quic::Connection>>>,
     subscribers: Fanout<ProtocolEvent<A>>,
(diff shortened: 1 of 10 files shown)


kim:
Ya I thought this interference with the membership protocol might disturb the force somehow, and I wasn't sure how to make it such that if the membership decides a connection should be dropped, it wouldn't forcefully drop any open git connections. We're just relying on refcounts now, which is how it's done in Rust, but obviously easy to break. There's some bigger reshuffling needed perhaps.

kim:
Please also observe the more legible tracing output :)

FintanH:
> Ya I thought this interference with the membership protocol might disturb the force somehow, and I wasn't sure how to make it such that if the membership decides a connection should be dropped, it wouldn't forcefully drop any open git connections. We're just relying on refcounts now, which is how it's done in Rust, but obviously easy to break. There's some bigger reshuffling needed perhaps.

Makes sense. So this one of the more esoteric parts of link in my mind. Is there any material I can go through to try understand what we're going for a little better? The epidemic broadcast tree paper linked in the docs, possibly?

kim:
> The epidemic broadcast tree paper linked in the docs, possibly?

Yes, both. [This](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/episub.md) is potentially also an interesting read for an attempt to optimise an overlay around "topics".

FintanH:
I gotz some reading to do so :)

librad/src/net/protocol.rs
```diff
@@ -194,90 +206,126 @@ where
...
+    /// convenience (one can pass [`None`], for example).
+    pub async fn open_git<Addrs>(
         &self,
         to: &PeerId,
-    ) -> Result<Upgraded<quic::Stream, upgrade::Git>, Error> {
-        self.open_stream(to, upgrade::Git).await
+        addr_hints: Addrs,
+    ) -> Result<Upgraded<quic::Stream, upgrade::Git>, Error>
+    where
+        Addrs: IntoIterator<Item = SocketAddr>,
```
cloudhead, line 220: FYI there's also https://doc.rust-lang.org/std/net/trait.ToSocketAddrs.html#tymethod.to_socket_addrs which is handy.

kim: This is fallible, though, I think it calls `gethostbyname` if it’s not an IP address. We can keep it pure, because we know we have `SocketAddr`s (or the caller can resolve).

cloudhead: True that, it can convert from all sorts of inputs.

PR #270, opened 2020-08-07
Polymorphic refspec constructor

### LINKED ISSUE #269: More Polymorphic Refspec
The `refspec` functions should take another reference that may have a different namespace and/or remote but still keep the same cardinality: https://github.com/radicle-dev/radicle-link/blob/master/librad/src/git/types/reference.rs#L236.

Some tests would be nice :eyes:

FintanH:
Fixes #269 

In our quest for MORE POLYMORPHISM, we move the refspec constructor of
two references into the polymorphic methods of Reference. We also allow
the remotes to change in a Refspec.
This means that we can pass in a Reference with a different Namespace
and Remote, and get back the Refspec for it.
The bonus is that the equality of the Cardinality of the two References is
enforced by the ReferenceInfo trait.

### DIFF
--- a/librad/src/git/types.rs
+++ b/librad/src/git/types.rs
@@ -115,14 +115,14 @@ impl<R> SymbolicRef<R> {
 }
 
 #[derive(Clone)]
-pub struct Refspec<R> {
-    pub(crate) remote: SomeReference<R>,
-    pub(crate) local: SomeReference<R>,
+pub struct Refspec<RemoteR, LocalR> {
+    pub(crate) remote: SomeReference<RemoteR>,
+    pub(crate) local: SomeReference<LocalR>,
     /// Indicate whether the spec should include the force flag `+`.
     pub force: Force,
 }
 
-impl<R: Display + Clone> Display for Refspec<R> {
+impl<R: Display + Clone, L: Display + Clone> Display for Refspec<R, L> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         if self.force.as_bool() {
             f.write_str("+")?;
@@ -131,7 +131,7 @@ impl<R: Display + Clone> Display for Refspec<R> {
     }
 }
 
-impl Refspec<PeerId> {
+impl Refspec<PeerId, PeerId> {
     /// [`Refspec`]s for fetching `rad/refs` in namespace [`Namespace`] from
     /// remote peer [`PeerId`], rejecting non-fast-forwards
     pub fn rad_signed_refs<'a>(
--- a/librad/src/git/types.rs
+++ b/librad/src/git/types.rs
@@ -115,10 +115,15 @@ impl<R> SymbolicRef<R> {
     }
 }
 
-/// A `Display`-like trait but only meant for `Refspec`s.
+/// If we want to be able to use a collection of [`Refspec`]s, where the remote
+/// portions don't line up, then we need to be able to talk about them using
+/// dynamically. In Rust this cab be done via `Box<dyn MyTrait>`.
 ///
-/// This allows us to treat them as opaque in collections, such a `Vec<Box<dyn
-/// AsRefspec>>`.
+/// In this case, we want to be able to turn [`Refspec`]s into `String`s. To
+/// avoid allowing anything that is `ToString` or `Display`, we opt for adding a
+/// `ToString`-like trait called `AsRefspec`.
+///
+/// **N.B.**: The only implementor of this trait should `Refspec`.
 pub trait AsRefspec {
     fn as_refspec(&self) -> String;
 }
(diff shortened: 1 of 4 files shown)


librad/src/git/types.rs
```diff
@@ -115,14 +115,14 @@ impl<R> SymbolicRef<R> {
 }
 
 #[derive(Clone)]
-pub struct Refspec<R> {
-    pub(crate) remote: SomeReference<R>,
-    pub(crate) local: SomeReference<R>,
+pub struct Refspec<RemoteR, LocalR> {
```
FintanH, line 118: Arrghhh I think I want these to be existential as well...

FintanH: But I think that might just make things more of a pain in the ass because I'd have 2 `Either`s so twice amount of case handling.

FintanH: Maybe having a `trait AsRefspec` would work and that way instead of using `Vec<Refspec<R, L>` we would have `Vec<Box<dyn AsRefspec>>`. To give an idea of what I'm thinking I'd like something like:
```rust
pub struct Remote {
  url: String, // possibly make Url
  name: String,
  fetch_specs: Vec<Box<dyn AsRefspec>>,
  push_specs: Vec<Box<dyn AsRefspec>>,
}
```

librad/src/git/types/remote.rs
```diff
@@ -0,0 +1,164 @@
...
+#[cfg(test)]
+mod tests {
+    use std::{io, marker::PhantomData};
+
+    use super::*;
+    use crate::{git::types::*, hash::Hash};
+    use librad_test::tempdir::WithTmpDir;
+
+    #[test]
+    fn can_create_remote() {
```
NunoAlexandre, line 125: I noticed that we prefix tests with `test_`, which is missing in this case.

FintanH: Emmmm I'm not sure that's necessarily true. I think there are a mix and match. Since they're under `test` modules anyway the `test_` prefix is redundant and I didn't give that enough thought at the time :sweat_smile:

NunoAlexandre: I might have happen to check a file where all tests were prefixed with `test_` :P but agree about the redundancy, just wanted to point out a possible inconsistency.

librad/src/git/types.rs
```diff
@@ -114,15 +115,55 @@ impl<R> SymbolicRef<R> {
     }
 }
 
+/// If we want to be able to use a collection of [`Refspec`]s, where the remote
+/// portions don't line up, then we need to be able to talk about them using
+/// dynamically. In Rust this cab be done via `Box<dyn MyTrait>`.
```
rudolfs, line 120:
```suggestion
/// dynamically. In Rust this can be done via `Box<dyn MyTrait>`.
```

librad/src/git/types.rs
```diff
@@ -114,15 +115,55 @@ impl<R> SymbolicRef<R> {
...
     pub force: Force,
 }
 
-impl<R: Display + Clone> Display for Refspec<R> {
+impl<R, L> Refspec<R, L> {
+    /// Allows to existentialise the `Refspec` into a dynamic `AsRefspec`. This
+    /// means we forget about the type parameters on `Refspec` and we can
+    /// then have a collection of `Refspec`s with differeing remotes.
+    ///
+    /// This function, while trivially, is useful for type-inference, since Rust
```
rudolfs, line 154:
```suggestion
    /// This function, while trivial, is useful for type-inference, since Rust
```

librad/src/git/types/remote.rs
```diff
@@ -0,0 +1,164 @@
...
+        WithTmpDir::new::<_, io::Error>(|path| {
+            let repo = git2::Repository::init(path).expect("failed to init repo");
+
+            let id = Hash::hash(b"geez");
+            let heads: FlatRef<String, _> = FlatRef::heads(PhantomData, None);
```
kim, line 130: It is a bit weird having to pass a `PhantomData` -- is there no way to `impl` around this? Or maybe create a type alias at least?

FintanH: Ya, I couldn't think of a way to get around this without repeating code. Like we could do:
```rust
impl FlatRef<Remote> {
  pub fn heads(/* yada yada */) -> Self {
    Reference::heads(PhantomData, remote)
  }
}
```

But it's verrrrrry boilerplatey... Macros? :scream_cat:

FintanH: > Or maybe create a type alias at least?

I thought about that, initially and had:
```rust
type NoNamespace = PhantomData<!>;
```

But didn't know where to go from there. We'd still need to pass in `PhantomData`.

FintanH: I'll create a follow-up ticket because I don't wanna spend more time shaving yaks :sweat_smile:

Issue #267, opened 2020-08-06
Git Trailers Utilities


kim:
We will be using [git trailers](https://git-scm.com/docs/git-interpret-trailers)
for "multisig" commits, and potentially other purposes, too. Unfortunately,
`libgit2` doesn't include any functionality for parsing or otherwise
manipulating trailers in a commit message.

We do not need to replicate exactly what `git-interpret-trailers` does, but we
SHOULD be able to deal with trailers which were not generated by us.

Specifically, what we want is:

* **Parse trailers**

  Given a `&str`, return an iterator of `(String, String)` pairs, such that:

    * Trailing whitespace at the end of the string slice is ignored
    * Only the last _paragraph_ of the string slice is considered
    * A paragraph is preceded by one or more empty lines (ie. whitespace
      including newlines), and does not itself contain empty lines (but it may
      contain newline characters)
    * The paragraph MUST consist of only trailers
    * A valid key-value pair consists of:

        * A `token`, with no leading whitespace, matching `[-[:alnum:]]+`
        * Optional horizontal whitespace
        * A `separator` (default `:`, _MAY_ be configurable)
        * Optional horizontal whitespace
        * A `value` consisting of UTF-8 characters
        * A newline or EOF

      Multi-line `value`s acc. to RFC 822 "folding" MAY be supported. This is
      expected to be somewhat rare in practice, so not a hard requirement.

* **Replace trailers**

  Given a `&str`, and a list of key-value pairs, return a `String`, such that
  any trailers from the input `&str` are removed, and replaced by rendering the
  supplied list. If the input does not contain any trailers, the supplied ones
  are simply appended as a paragraph. The result must be parseable acc. to
  **Parse trailers**.

kim:
There might be other convenience things we want to layer on top of this, such as `const`s for well-known trailers, a standard way to serialise to a `value`, combinators to manipulate the trailers list. But we'll add them as we go.

NunoAlexandre:
Thanks for this super clear feature spec, @kim!

Just posting some good reads on `git trailers`:

- https://blog.k-nut.eu/extending-git-commits-with-trailers
- https://git-scm.com/docs/git-interpret-trailers
- https://git.wiki.kernel.org/index.php/CommitMessageConventions

NunoAlexandre:
> Unfortunately, `libgit2` doesn't include any functionality for parsing or otherwise manipulating trailers in a commit message.

[...]

PR #271, opened 2020-08-07
Parse git commit message trailers

NunoAlexandre:
`DRAFT`

Addresses #267 

* **Parse trailers**

  Given a `&str`, return an iterator of `(String, String)` pairs, such that:

    - [x] Trailing whitespace at the end of the string slice is ignored
    - [x] Only the last _paragraph_ of the string slice is considered
    - [x] A paragraph is preceded by one or more empty lines (ie. whitespace
      including newlines), and does not itself contain empty lines (but it may
      contain newline characters)
    - [x] The paragraph MUST consist of only trailers
    - [x] A valid token-value pair consists of:
        - [x] A `token`, with no leading whitespace, composed of alphanumeric digits or '-'
        - [x] Optional horizontal whitespace
        - [x] A `separator` (default `:`, _MAY_ be configurable)
        - [x] Optional horizontal whitespace
        - [x] A `value` consisting of UTF-8 characters
        - [x] A newline or EOF

### DIFF
--- a/librad/Cargo.toml
+++ b/librad/Cargo.toml
@@ -37,7 +37,6 @@ tracing-futures = "0.2"
 urltemplate = "0.1"
 webpki = "0.21"
 
-
 [dependencies.futures]
 version = "0.3"
 features = ["thread-pool"]
--- a/librad/Cargo.toml
+++ b/librad/Cargo.toml
@@ -21,6 +21,7 @@ libc = "0.2"
 log = "0.4"
 multibase = "0.8"
 multihash = "0.11"
+nom = "5"
 nonempty = "0.5"
 olpc-cjson = "0.1"
 percent-encoding = "2.1"
@@ -35,6 +36,7 @@ tracing = "0.1"
 tracing-futures = "0.2"
 urltemplate = "0.1"
 webpki = "0.21"
+
 
 [dependencies.futures]
 version = "0.3"
--- a/librad/src/git/trailer.rs
+++ b/librad/src/git/trailer.rs
@@ -1,0 +1,256 @@
+// This file is part of radicle-link
+// <https://github.com/radicle-dev/radicle-link>
+//
+// Copyright (C) 2019-2020 The Radicle Team <dev@radicle.xyz>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 or
+// later as published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+#[derive(Debug, Clone, Eq, PartialEq, thiserror::Error)]
+pub enum Error<'a> {
+    #[error("The trailers paragraph is missing in the given message")]
+    MissingParagraph,
+
+    #[error(
+        "One or more trailers are not in the parseable format <token><separator><value>: '{0}'"
+    )]
+    Unparsable(&'a str),
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct Trailer<'a> {
+    pub token: Token<'a>,
+    pub values: Vec<&'a str>,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct Token<'a>(&'a str);
+
+/// Parse the trailers of the given message. It looks up the last paragraph
+/// of the message and attempts to parse each of its lines as a [Trailer].
+/// Fails if no trailers paragraph is found or if at least one trailer
+/// fails to be parsed.
+pub fn parse<'a>(message: &'a str, separators: &'a str) -> Result<Vec<Trailer<'a>>, Error<'a>> {
+    let trailers_paragraph =
+        match parser::paragraphs(message.trim_end()).map(|(_, ps)| ps.last().cloned()) {
+            Ok(None) | Err(_) => return Err(Error::MissingParagraph),
+            Ok(Some(p)) => p,
+        };
+
+    match parser::trailers(trailers_paragraph, separators) {
+        Ok((rest, trailers)) if rest.is_empty() => Ok(trailers),
+        Ok((unparseable, _)) => Err(Error::Unparsable(unparseable)),
+        Err(_) => Err(Error::Unparsable(trailers_paragraph)),
+    }
+}
+
+pub mod parser {
+    use super::{Token, Trailer};
+    use nom::{
+        branch::alt,
+        bytes::complete::{tag, take_until, take_while1},
+        character::complete::{line_ending, not_line_ending, one_of, space0, space1},
+        combinator::rest,
+        multi::{many0, separated_list},
+        sequence::{delimited, preceded, separated_pair, terminated},
+        IResult,
+    };
+
+    const EMPTY_LINE: &str = "\n\n";
+
+    pub fn paragraphs(s: &str) -> IResult<&str, Vec<&str>> {
+        separated_list(tag(EMPTY_LINE), paragraph)(s)
+    }
+
+    pub fn paragraph(s: &str) -> IResult<&str, &str> {
+        alt((take_until(EMPTY_LINE), rest))(s)
+    }
+
+    /// Parse all the possible trailers.
+    /// It stops when it can no longer parse valid trailers.
+    pub fn trailers<'a>(s: &'a str, separators: &'a str) -> IResult<&'a str, Vec<Trailer<'a>>> {
+        many0(|s| trailer(s, separators))(s)
+    }
+
+    /// Parse a trailer, which can have an inlined or multilined value.
+    pub fn trailer<'a>(s: &'a str, separators: &'a str) -> IResult<&'a str, Trailer<'a>> {
+        let parser = separated_pair(token, |s| separator(separators, s), values);
+        let (rest, (token, values)) = parser(s)?;
+        Ok((rest, Trailer { token, values }))
+    }
+
+    /// Parse a trailer token.
+    fn token(s: &str) -> IResult<&str, Token> {
+        take_while1(|c: char| c.is_alphanumeric() || c == '-')(s)
+            .map(|(i, token_str)| (i, Token(token_str)))
+    }
+
+    /// Parse the trailer separator, which can be delimited by spaces.
+    fn separator<'a>(separators: &'a str, s: &'a str) -> IResult<&'a str, char> {
+        delimited(space0, one_of(separators), space0)(s)
+    }
+
+    /// Parse the trailer values, which gathers the value after the separator
+    /// (if any) and possible following multilined values, indented by a
+    /// space.
+    fn values(s: &str) -> IResult<&str, Vec<&str>> {
+        let (r, opt_inline_value) = until_eol_or_eof(s)?;
+        let (r, mut values) = multiline_values(r)?;
+        if !opt_inline_value.is_empty() {
+            values.insert(0, opt_inline_value)
+        }
+        Ok((r, values))
+    }
+
+    fn multiline_values(s: &str) -> IResult<&str, Vec<&str>> {
+        many0(indented_line_contents)(s)
+    }
+
+    fn until_eol_or_eof(s: &str) -> IResult<&str, &str> {
+        alt((until_eol, rest))(s)
+    }
+
+    /// Parse an indented line, i.e, a line that starts with a space.
+    /// Extracts the line contents, ignoring the indentation and the
+    /// new line character.
+    fn indented_line_contents(s: &str) -> IResult<&str, &str> {
+        preceded(space1, until_eol_or_eof)(s)
+    }
+
+    /// Consume the input until the end of the line, ignoring the new line
+    /// character.
+    fn until_eol(s: &str) -> IResult<&str, &str> {
+        terminated(not_line_ending, line_ending)(s)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn parse_message_with_valid_trailers() {
+        let msg = r#"Subject
+
+A multiline
+description.
+
+Co-authored-by: John Doe <john.doe@test.com>
+Ticket: #42
+Tested-by:
+    John <john@test.com>
+    Jane <jane@test.com>
+Just-a-token:
+
+"#;
+        assert_eq!(
+            parse(msg, ":"),
+            Ok(vec![
+                new_trailer("Co-authored-by", vec!["John Doe <john.doe@test.com>"]),
+                new_trailer("Ticket", vec!["#42"]),
+                new_trailer(
+                    "Tested-by",
+                    vec!["John <john@test.com>", "Jane <jane@test.com>"]
+                ),
+                new_trailer("Just-a-token", vec![]),
+            ])
+        )
+    }
+
+    #[test]
+    fn parse_message_trailers_with_custom_separators() {
+        let separators = ":=$";
+        let msg = r#"Subject
+
+A multiline
+description.
+
+Co-authored-by: John Doe <john.doe@test.com>
+Ticket = #42
+Tested-by $User <user@test.com>
+    John <john@test.com>
+    Jane <jane@test.com>
+"#;
+        assert_eq!(
+            parse(msg, separators),
+            Ok(vec![
+                new_trailer("Co-authored-by", vec!["John Doe <john.doe@test.com>"]),
+                new_trailer("Ticket", vec!["#42"]),
+                new_trailer(
+                    "Tested-by",
+                    vec![
+                        "User <user@test.com>",
+                        "John <john@test.com>",
+                        "Jane <jane@test.com>"
+                    ]
+                ),
+            ])
+        )
+    }
+
+    #[test]
+    fn parse_message_trailers_with_missing_token() {
+        let msg = r#"Subject
+
+Good-trailer: true
+John Doe <john.doe@test.com> # Unparsable token due to missing token"#;
+        assert_eq!(
+            parse(msg, ":"),
+            Err(Error::Unparsable(
+                "John Doe <john.doe@test.com> # Unparsable token due to missing token"
+            ))
+        )
+    }
+
+    #[test]
+    fn parse_message_trailers_with_invalid_token() {
+        let msg = r#"Subject
+
+Good-trailer: true
+&!#: John Doe <john.doe@test.com> # Unparsable token due to invalid token"#;
+        assert_eq!(
+            parse(msg, ":"),
+            Err(Error::Unparsable(
+                "&!#: John Doe <john.doe@test.com> # Unparsable token due to invalid token"
+            ))
+        )
+    }
+
+    #[test]
+    fn parse_message_with_only_trailers() {
+        let msg = r#"Co-authored-by: John Doe <john.doe@test.com>
+Ticket: #42
+Tested-by: Tester <tester@test.com>
+"#;
+        assert_eq!(
+            parse(msg, ":"),
+            Ok(vec![
+                new_trailer("Co-authored-by", vec!["John Doe <john.doe@test.com>"]),
+                new_trailer("Ticket", vec!["#42"]),
+                new_trailer("Tested-by", vec!["Tester <tester@test.com>"]),
+            ])
+        )
+    }
+
+    #[test]
+    fn parse_empty_message() {
+        let msg = "";
+        assert_eq!(parse(msg, ":"), Err(Error::MissingParagraph))
+    }
+
+    fn new_trailer<'a>(token: &'a str, values: Vec<&'a str>) -> Trailer<'a> {
+        Trailer {
+            token: Token(token),
+            values,
+        }
+    }
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -717,6 +717,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e2abad23fbc42b3700f2f279844dc832adb2b2eb069b2df918f455c4e18cc646"
 
 [[package]]
+name = "lexical-core"
+version = "0.7.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "db65c6da02e61f55dae90a0ae427b2a5f6b3e8db09f58d10efab23af92592616"
+dependencies = [
+ "arrayvec",
+ "bitflags",
+ "cfg-if",
+ "ryu",
+ "static_assertions",
+]
+
+[[package]]
 name = "libc"
 version = "0.2.74"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -763,6 +776,7 @@ dependencies = [
  "minicbor",
  "multibase",
  "multihash",
+ "nom",
  "nonempty",
  "olpc-cjson",
  "percent-encoding 2.1.0",
@@ -997,6 +1011,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b93853da6d84c2e3c7d730d6473e8817692dd89be387eb01b94d7f108ecb5b8c"
 dependencies = [
  "hashbrown",
+]
+
+[[package]]
+name = "nom"
+version = "5.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ffb4262d26ed83a1c0a33a38fe2bb15797329c85770da05e6b828ddb782627af"
+dependencies = [
+ "lexical-core",
+ "memchr",
+ "version_check",
 ]
 
 [[package]]
@@ -1726,6 +1751,12 @@ name = "spin"
 version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6e63cff320ae2c57904679ba7cb63280a3dc4613885beafb148ee7bf9aa9042d"
+
+[[package]]
+name = "static_assertions"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
 
 [[package]]
 name = "syn"
--- a/librad/src/git.rs
+++ b/librad/src/git.rs
@@ -21,6 +21,7 @@ pub mod p2p;
 pub mod refs;
 pub mod repo;
 pub mod storage;
+pub mod trailer;
 pub mod types;
 
 pub(crate) mod header;


librad/src/git/trailer.rs
```diff
@@ -0,0 +1,139 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use std::convert::TryFrom;
+
+#[derive(Debug, Clone, Eq, PartialEq, thiserror::Error)]
+pub enum Error {
+    #[error("the given input is not a well-formed trailer")]
+    IllFormed,
```
kim, line 23: nit: We have been using sentence case since before radicle-dev/radicle-decisions#3 landed. I don't have a strong preference for either style, but I think this might be one of the very rare cases where consistency is preferable. Perhaps we can have a separate patch which converts everything to lower case in one go, if that's the preferred style?

librad/src/git/trailer.rs
```diff
@@ -0,0 +1,139 @@
...
+
+    #[error("could not parse the trailer due to an invalid token found")]
+    Token(#[from] InvalidToken),
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct Trailer {
+    token: Token,
+    value: String,
```
kim, line 32: Since we will either not touch the trailer at all, or parse the value into some other representation, it would be good to leave `value` at `&str`.

librad/src/git/trailer.rs
```diff
@@ -0,0 +1,139 @@
...
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct Token(String);
+
+#[derive(Debug, Clone, Eq, PartialEq, thiserror::Error)]
+#[error("invalid token")]
+pub struct InvalidToken;
+
+impl TryFrom<String> for Token {
```
kim, line 61: Accordingly, perhaps implement `TryFrom<String>` in terms of `FromStr`, not the other way round.

librad/src/git/trailer.rs
```diff
@@ -0,0 +1,139 @@
...
+        // which is clearly supported.
+        if value.chars().all(|c| c.is_alphanumeric() || c == '-') {
+            Ok(Token(value))
+        } else {
+            Err(InvalidToken)
+        }
+    }
+}
+
+impl TryFrom<&str> for Token {
```
kim, line 77: `FromStr`?

librad/src/git/trailer.rs
```diff
@@ -0,0 +1,139 @@
...
+
+#[derive(Debug, Clone, Eq, PartialEq, thiserror::Error)]
+#[error("invalid token")]
+pub struct InvalidToken;
+
+impl TryFrom<String> for Token {
+    type Error = InvalidToken;
+
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        // TODO(nuno): get this straight. The docs are not super clear.
```
kim, line 65: [This](https://github.com/git/git/blob/master/trailer.c#L613) is what git does, which I think is sensible.

NunoAlexandre: Cool. I am somewhat surprised at how loose the rules are and that things seem to work nonetheless.

kim: Welcome to git :joy:

librad/src/git/trailer.rs
NunoAlexandre: I am good with your argument for consistency. The standard to follow is to go lower case, the reason for that being that when building a stack of errors messages, having them lower case shows that continuity. We have adopted it fully in the registry and it's a good looking approach indeed.

I will make this now consistent with the other error messages in this repo and create a follow-up issue to make them all lower case.

librad/src/git/trailer.rs
NunoAlexandre: That's a good point.

NunoAlexandre: Wait, I actually had decided to use `String` since using `&str` requires the whole lifetime parameter bloat, which cascades. Is your concern purely on performance?

librad/src/git/trailer.rs
NunoAlexandre: See https://github.com/rust-lang/api-guidelines/issues/182#issuecomment-479657836, keeping in mind that 2 years have passed and `TryFrom` is stable and recommended.

I'd be fine either way, but if we only _need_ one then I'd keep `TryFrom` unless if there's a good reason to prefer `FromStr`?

kim: We are using `FromStr` everywhere, not least because of `str::parse`. Feel free to _also_ provide `TryFrom`.

librad/src/git/trailer.rs
kim: And what's the problem with lifetime parameters? Copying lots of small strings is almost always bad, and we'll be doing this a lot.

I was thinking that keeping `Token` owned would be necessary because of case-folding, but it looks like [unicase](https://docs.rs/unicase/2.6.0/unicase/) works fine with borrowed values, so we can keep it `&str`, too.

librad/src/git/trailer.rs
NunoAlexandre: Thanks. Moved to `FromStr` as default, using `TryFrom` as a convenience layer on top.

librad/src/git/trailer.rs
kim: Oh well, disregard my comments about `FromStr` then -- you'll indeed need `TryFrom` when the types have lifetimes.

NunoAlexandre:
@kim would you like to have a look at the latest revision? I have checked pretty much all the requirements, with a few points to tweak. See the items in the list above marked with :pencil2:

If there is an immediate need for this feature, I suggest that we move forward with the current approach and follow up with a move to use parser combinators and with supporting replacing trailers.

librad/src/git/trailer.rs
```diff
@@ -0,0 +1,273 @@
...
+            value,
+        })
+    }
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct Token<'a>(&'a str);
+
+#[derive(Debug, Clone, Eq, PartialEq, thiserror::Error)]
+#[error("Invalid token")]
```
FintanH, line 59: A better error message might be one that says only alphanumeric and `-`  characters are allowed

librad/src/git/trailer.rs
```diff
@@ -0,0 +1,273 @@
...
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct Trailer<'a> {
+    token: Token<'a>,
+    value: &'a str,
+}
+
+impl<'a> Trailer<'a> {
+    #[allow(dead_code)]
```
kim, line 36: How come this is necessary?

librad/src/git/trailer.rs
```diff
@@ -0,0 +1,273 @@
...
+    #[error("The given input is not a well-formed trailer")]
+    IllFormed,
+
+    #[error("Could not parse the trailer due to an invalid token found")]
+    Token(#[from] InvalidToken),
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct Trailer<'a> {
+    token: Token<'a>,
```
kim, line 31: Can we make the fields `pub`?

librad/src/git/trailer.rs
```diff
@@ -0,0 +1,273 @@
...
+            A subject that succinctly states what this commit does
+
+            A longer, more detailed paragraph that describes the what's
+            and why's of the changes being introduced by this commit. It
+            is usually capped at 72 chars-long lines.
+
+            # Ill-formed trailer, missing the token and the separator
+            John Doe <john.doe@test.com>
+            "#;
+        assert!(parse(msg, &[separator]).is_err())
```
FintanH, line 166: Is it possible to `assert_eq` on the error we expect? I always find this more useful for 1) documentation purposes 2) debugging purposes. But sometimes it's a PITA, so if that's the case you can ignore this :)

librad/src/git/trailer.rs
```diff
@@ -0,0 +1,273 @@
...
+/// of the message and attempts to parse each of its lines as a [Trailer].
+/// Fails if at least one trailer fails to be parsed.
+// Todo(nuno):
+//   * return iterator instead of Vec?
+//   * rsplit("\n\n").next() is always Some(_), even for "", so consider simpler
+//     return type.
+pub fn parse<'a>(
+    message: &'a str,
+    separators: &'a [char],
+) -> Result<Option<Vec<Trailer<'a>>>, Error> {
```
FintanH, line 84: Ya so isn't `None` the same as `vec![]`?

NunoAlexandre: Yeah, they are semantically the same. My goal really was to differentiate between the case where there is no trailers paragraph vs the case where there is a trailers paragraph but they ill-formed. The thing is that for the first case we can't really tell. Shall we just have `Result<Vec<Trailer, Error>>` then? It seems sensible to me.

librad/src/git/trailer.rs
NunoAlexandre: Yeah, we should do that, I agree with those criteria. Will do!

librad/src/git/trailer.rs
NunoAlexandre: Yes :ok_hand:

librad/src/git/trailer.rs
```diff
@@ -0,0 +1,273 @@
...
+
+impl<'a> Trailer<'a> {
+    #[allow(dead_code)]
+    /// Try and parse a [Trailer] form a given raw &str that is
+    /// expected to represent a trailer in raw form.
+    pub fn parse(s: &'a str, separators: &'a [char]) -> Result<Trailer<'a>, Error> {
+        let parts: Vec<&str> = s.split(&separators[..]).map(|s| s.trim()).collect();
+        match parts.as_slice() {
+            [token, value] => Self::new(*token, *value),
+            _ => Err(Error::IllFormed),
```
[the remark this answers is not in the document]
NunoAlexandre: This variant accounts for the whole line not being a well-formed trailer, ie, we couldn't parse a `<token><separator><value>` from it. So if parsing that format fails, we report that the input is ill-formed. Does that change your perspective?

librad/src/git/trailer.rs
NunoAlexandre: I was thinking, instead, to keep them private and provide getters and setters because encapsulation :nerd_face:

librad/src/git/trailer.rs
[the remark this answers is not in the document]
FintanH: Ya, but there can be multiple lines in a trailer, right? How will I know which line failed? Imagine a compiler that told you that you had a parse error but not which line :smile:

librad/src/git/trailer.rs
FintanH: Ya, I think `Result<Vec<Trailer>, Error>` makes more sense to me :)

librad/src/git/trailer.rs
NunoAlexandre: I added it right away before having added tests to silent the warning. Dropped it :+1:

kim:
> I noticed that currently, we support commit messages that only contain trailers.

Yeah, I was wondering about this, too. [git](https://github.com/git/git/blob/master/trailer.c#L824) wants the first line to be the subject, and doesn't consider it a trailer even if it would parse as one. _However_, we will also be using [git-notes](https://git-scm.com/docs/git-notes), where we have the same format by convention, but inside `blob`s not commit messages. Git doesn't interpret the note contents, though.

If we want to be pedantic, or need to be compatible, we would add wrappers to convert from `git2::Commit -> Vec<Trailer>` resp. `git2::Note -> Vec<Trailer>`. The parser can be the same.

librad/src/git/trailer.rs
```diff
@@ -0,0 +1,273 @@
...
+        } else {
+            Err(InvalidToken)
+        }
+    }
+}
+
+/// Parse the trailers of the given message. It looks up the last paragraph
+/// of the message and attempts to parse each of its lines as a [Trailer].
+/// Fails if at least one trailer fails to be parsed.
+// Todo(nuno):
```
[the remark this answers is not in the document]
NunoAlexandre: > > * return iterator instead of Vec?
> 
> We can leave it like this for now. An iterator would need to yield `Item = Result<Trailer, Error>`, which is more flexible, but also more annoying.

:ok_hand: 

> 
> > * rsplit("\n\n").next() is always Some(_), even for "", so consider simpler return type.
> 
> Yeah, `Option<Vec<A>>` is the same as either the empty or the 1-element vector, so the `Option` is redundant.

Yeah agree.

> 
> It would be good to note that the trailers are collected in reverse order.

Not sure if I get this? Is your suggestion to add that bit to the docs? If so, that isn't right as the trailers are collected in the normal order.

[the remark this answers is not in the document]
kim: >  If so, that isn't right as the trailers are collected in the normal order.

`rsplit` yields the elements in reverse order compared to how they appear in the input. Am I missing a `rev()` somewhere?

librad/src/git/trailer.rs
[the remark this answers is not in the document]
NunoAlexandre: I am not sure if you are referring to the RFC822 (see https://github.com/radicle-dev/radicle-link/pull/271#pullrequestreview-464834027) or suggesting preserving the info about which trailer as ill-formed.

Say we have:

```
Subject

Description

TokenA: ValueA
TokenB ValueB
TokenC: ValueC
```

Is your suggestion that the error should inform that the trailer line number #2 is ill-formed, or, say we have:

```
Subject

Description

TokenA: 
  Value1
  Value2
  Value3
```

... that we should inform which of the lines is ill-formed within a specific trailer is ill-formed? 

If the first, that can be useful indeed. Alternatively, we can include the str line that failed to parse in the error itself. If the second case,  I think that only the first line can be ill-formed, but I need to digest the contents of RFC822 better.

librad/src/git/trailer.rs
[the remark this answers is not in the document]
NunoAlexandre: You are not missing a `rev()`, the thing is that we iterate in reverse order over the paragraphs of the message, but _then_ we iterate in the normal order per paragraph, so the trailers are collected in their natural order.

librad/src/git/trailer.rs
kim: You already have all the necessary encapsulation by only allowing `Token` to be constructed through a "smart" constructor. More often than not, getters and setters are a PITA in Rust due to "partial borrow" issues and whatnot. This is just a datatype, being able to pattern-match on it is going to be the most convenient way to work with it.

NunoAlexandre: @kim, I was joking :hot_face:

NunoAlexandre:
> > I noticed that currently, we support commit messages that only contain trailers.
> 
> Yeah, I was wondering about this, too. [git](https://github.com/git/git/blob/master/trailer.c#L824) wants the first line to be the subject, and doesn't consider it a trailer even if it would parse as one. _However_, we will also be using [git-notes](https://git-scm.com/docs/git-notes), where we have the same format by convention, but inside `blob`s not commit messages. Git doesn't interpret the note contents, though.
> 
> If we want to be pedantic, or need to be compatible, we would add wrappers to convert from `git2::Commit -> Vec<Trailer>` resp. `git2::Note -> Vec<Trailer>`. The parser can be the same.

Cool, sounds good to me. I will leave it as is then, we can tweak and wrap things when we have real use cases to work against.

NunoAlexandre:
> > pencil2 @kim now I filter out comments and empty lines. Do we want to stricter instead?
> 
> Good call with the comments. Pedantically speaking, a blank line (ie. PCRE `\n\h*\n`) would count as a paragraph separator. I think it's good for now, as we will be dealing with programmatically-created messages for the time being.

Sounds good to me!

> 
> > A token, with no leading whitespace, composed of alphanumeric digits or '-'
> > pencil2 @kim I now trim leading and trailing whitespace. Will change to not trim leading whitespace
> 
> The reason for the "no leading whitespace" rule is that we may eventually need multi-line values. Acc. to RFC822 "folding", the following would be `Trailer { token: Token("Signed-off-by"), value: "X-Favourite-Food: banana Chucky <chuck@norris.gov>" }`:
> 
> ```
> Signed-off-by:
>   X-Favourite-Food: banana
>   Chucky <chuck@norris.gov>
> ```
> 
> The parsed `token` value should of course be trimmed of leading and trailing whitespace.

Right right, I see. In the original list of requirements, this point was not a hard requirement to I sort of put it aside, but I guess it's smarter to take it into account now. 

I assume that multi-line values should be preserved as-they-are when parsed? As in, that they should not be flattened or represented in a vector, as the first case would mean `serialize ∘ deserialize` to NOT be an idempotent operation and the latter to require the `Trailer.value`'s type to become `Vec<&'a str>`.

librad/src/git/trailer.rs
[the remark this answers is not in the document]
kim: Ah... right.

librad/src/git/trailer.rs
kim: You! _**shakes fist**_ Next time I'm gonna wrap you in a builder pattern!

librad/src/git/trailer.rs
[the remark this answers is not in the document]
FintanH: > suggesting preserving the info about which trailer is ill-formed.

I'm suggesting this. So if we're parsing the error would be `IllFormed` and it could be any of the 3 trailer lines and, exactly, you're saying the second line is ill-formed. Which I actually missed as I was scanning through your comment :joy: 

In the second case, we might want a nested error, like the "For '{Token}' the '{Value}' was ill-formed".

I actually haven't read the RFC so not sure if the second suggestion is useful :grimacing:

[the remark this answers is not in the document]
NunoAlexandre: Cool, sounds good to me :)

kim:
> I assume that multi-line values should be preserved as-they-are when parsed?

Oof, good point. 

RFC822 says 

> The process of moving  from  this  folded   multiple-line
        representation  of a header field to its single line represen-
        tation is called "unfolding".  Unfolding  is  accomplished  by
        regarding   CRLF   immediately  followed  by  a  LWSP-char  as
        equivalent to the LWSP-char.

Which is in-line with:  

```shell
$ echo -ne "Subject\n\nLong-trailer: asd\n qwe\n zxc"| git interpret-trailers --parse
Long-trailer: asd qwe zxc
```

_Yet_:

```shell
$ echo -ne "Subject\n\nLong-trailer: asd\n qwe\n zxc"| git interpret-trailers --trailer "Other: foo"
Subject

Long-trailer: asd
 qwe
 zxc
Other: foo
```

I'd say we preserve all whitespace, and add a combinator to contract `\n\s+` to `\n ` if and when we need it.

NunoAlexandre:
@kim @FintanH regarding [supporting multi-line values](https://github.com/radicle-dev/radicle-link/pull/271#issuecomment-671860357), I am losing track here.

I first thought it could be fairly easily achieved by splitting the trailers paragraph using a regex `r"\n(?:\S)"`, i.e, split the trailers by newlines followed by a non-whitespace character. However, Rust doesn't seem to support splitting with uncaptured groups, so with this approach, we ended up with the trailers well split but with the token missing its first character.

I got it working with this ad-hoc code, which is the ugliest thing I've written in a long time:

``` rust
fn split_trailers(paragraph: &str) -> Vec<String> {
    let mut trailers: Vec<String> = vec![];
    let mut value: String = "".to_string();
    for line in paragraph.lines().rev() {
        // A multi-line value
        if line.starts_with(" ") {
            if value.is_empty() {
                value = line.to_string();
            } else {
                value = line.to_owned() + &value;
            }
        }
        // A line defining a token
        else {
            if value.is_empty() {
                trailers.push(line.to_string());
            } else {
                let flatten = line.to_owned() + &value;
                trailers.push(flatten)
            }
            value = "".to_string();
        }
    }
    trailers.reverse();
    trailers
}

let trailers_paragraph = r#"Signed-off-by:
 Chucky <chuck@norris.gov>
 Arnold <arnold@olympia.com>
Thanks:
 Father
 Mother
 Wife"#;

let trailers = split_trailers(trailers_paragraph);
assert_eq!(
    trailers,
    [
        "Signed-off-by: Chucky <chuck@norris.gov> Arnold <arnold@olympia.com>",
        "Thanks: Father Mother Wife",
    ]
)
```

I have postponed moving to a parser combinator library to not be tackling both the problem itself and a new library at the same time, but perhaps that's just the trigger to do so? Unless if you have another idea?

FintanH:
@NunoAlexandre: Ya, I would only reach for string splitters and regex libraries if I'm parsing a single line. As soon as I would get into more structured data I'd honestly prefer a parser combinator library :) While there may be some learning involved, the docs seem to have good links to guides and I'm sure it'd be more productive to get something like this working in the long run.

librad/src/git/trailer.rs
```diff
@@ -54,6 +61,18 @@ impl<'a> Trailer<'a> {
     }
 }
 
+fn token_parser(s: &str) -> IResult<&str, &str> {
+    take_while(|c: char| c.is_alphanumeric() || c == '-')(s)
+}
+
+fn separator_parser(s: &str) -> IResult<&str, char> {
+    delimited(space0, nom::character::complete::char(':'), space0)(s)
```
FintanH, line 69: A slight nit: Parsing spaces here could be considered _too much_ work for a combinator. A separator parser would expect the separator and nothing else.
The same for `value_parser`, where it should be `zero_or_more(any_char)`. 

Then all the white space is dealt with a level above for a single trailer.

NunoAlexandre: You mean that a parser should be more granular, while here we are actually composing three parsers in one, right? Just to make sure I am getting your point straight :)

FintanH: Ya exactly. The `separator_parser` should only know how to parse the separator and not whitespace.

The thing that cares about white space is the trailer :)

NunoAlexandre: Sure thing :) First I want to figure the harder bits and then tweak all those nits :ok_hand:

FintanH: Sounds good :100:

NunoAlexandre:
Alright, so I have been diving through `nom`. The latest version, 5, is pretty cool. The issue I am encountering is that the docs are are scattered between different versions, where versions prior to number 5 look significantly different, making it harder to get to learn the library.

Some useful links I have been tabbing through:

- http://unhandledexpression.com/general/2019/06/17/nom-5-is-here.html
- https://github.com/benkay86/nom-tutorial
- https://github.com/Geal/nom/blob/master/doc/choosing_a_combinator.md

### What's left

- Support custom separators, now hardcoded to `':'`
- Proper and more granular error propagation
  Be able to report whether a trailer failed to parse at the token, separator, or value level
- Support multiline values
- Move all parsing code to nom parsers

NunoAlexandre:
`Update`: Just pushed a new version of the trailers parser using `nom` for the most part :tada: 

This version supports multiline values, custom separators, it reports from what point the input could no longer be parsed, it simplifies a lot of the machinery around Error types, and it's starting to look beautiful #nomnom :cookie: 
 
### What's left

* use nom to parse the last paragraph of the given message
* ignore commented lines
* add doc tests to each parser

Regarding the error granularity / using a custom error type matter, I am no longer convinced that we need it. Now we return an error when no trailers paragraph is found, `MissingParagraph`, or when a given trailer is unparseable, `Unparseable`. That gives a caller enough information as to why the trailers could not be extracted from a message.

librad/src/git/trailer.rs
```diff
@@ -0,0 +1,250 @@
...
+    /// Parse the trailer separator, which can be delimited by spaces.
+    fn separator<'a>(separators: &'a str, s: &'a str) -> IResult<&'a str, char> {
+        delimited(space0, one_of(separators), space0)(s)
+    }
+
+    /// Parse the trailer values, which gathers the value after the separator
+    /// (if any) and possible following multilined values, indented by a
+    /// space.
+    fn values(s: &str) -> IResult<&str, Vec<&str>> {
+        let (r, opt_inline_value) = until_eol(s)?;
```
NunoAlexandre, line 93: @FintanH note that I didn't use `alt`, as we must support both cases, not or the other. I.e, a trailer value can be specified inline AND in following indented lines:

``` bash
$ echo -ne "Subject\n\nLong-trailer:asd\n qwe\n zxc"| git interpret-trailers --parse
Long-trailer: asd qwe zxc
```

NunoAlexandre:
Comments are oddly supported in git trailers :thinking: 

* At the first trailer, it correctly ignores it and parses the rest:

``` bash
 ➜ echo -ne "Subject\n\n#Foo:Bar\nLong-trailer:asd\n qwe\n zxc\nX:123"| git interpret-trailers --parse             ~/radicle-link  git-trailers ●
Long-trailer: asd qwe zxc
X: 123
```

* At a trailer in between, it fails to parse all trailers:

``` bash
 ➜ echo -ne "Subject\n\nFoo:Bar\n#Long-trailer:asd\n qwe\n zxc\nX:123"| git interpret-trailers --parse             ~/radicle-link  git-trailers ●
```

* At the last trailer, it correctly parses all previous trailers:

``` bash
 ➜ echo -ne "Subject\n\nFoo:Bar\nLong-trailer:asd\n qwe\n zxc\n#X:123\n"| git interpret-trailers --parse           ~/radicle-link  git-trailers ●
Foo: Bar
Long-trailer: asd qwe zxc
```

kim:
> Comments are oddly supported in git trailers

It's because of this:

> Existing trailers are extracted from the input message by looking for a group of one or more lines that (i) is all trailers, or (ii) contains at least one Git-generated or user-configured trailer and consists of at least 25% trailers.

We don't need to support `(ii)` -- because, why?! -- and adding comment support to `(i)` is fine by me, although also not strictly required.

librad/src/git/trailer.rs
```diff
@@ -0,0 +1,245 @@
...
+
+    /// Parse the trailer separator, which can be delimited by spaces.
+    fn separator<'a>(separators: &'a str, s: &'a str) -> IResult<&'a str, char> {
+        delimited(space0, one_of(separators), space0)(s)
+    }
+
+    /// Parse the trailer values, which gathers the value after the separator
+    /// (if any) and possible following multilined values, indented by a
+    /// space.
+    fn values(s: &str) -> IResult<&str, Vec<&str>> {
```
FintanH, line 92: Can we return a `NonEmpty` instead? We shouldn't be able to have a `Trailer` without any values right?

librad/src/git/trailer.rs
```diff
@@ -0,0 +1,245 @@
...
+        let (r, opt_inline_value) = until_eol(s)?;
+        let (r, mut values) = multiline_values(r)?;
+        if !opt_inline_value.is_empty() {
+            values.insert(0, opt_inline_value)
+        }
+        Ok((r, values))
```
FintanH, line 98: I don't understand this. If `values` is empty then what happened to `opt_inline_value`?

librad/src/git/trailer.rs
```diff
@@ -0,0 +1,245 @@
...
+A longer, more detailed paragraph that describes the what's
+and why's of the changes being introduced by this commit. It
+is usually capped at 72 chars-long lines.
+
+Co-authored-by: John Doe <john.doe@test.com>
+Ticket: #42
+Tested-by:
+    John <john@test.com>
+    Jane <jane@test.com>
+Just-a-token:
```
FintanH, line 137: Oh, we can have empty values?!

librad/src/git/trailer.rs
NunoAlexandre: Wrong actually:

```
 ➜ echo -ne "Subject\n\n#Foo:Bar\nLong-trailer:asd\n qwe\n zxc\nX:\n\n\n"| git interpret-trailers --parse          ~/radicle-link  git-trailers ●
Long-trailer: asd qwe zxc
X:

```

librad/src/git/trailer.rs
NunoAlexandre: Yap: https://github.com/radicle-dev/radicle-link/pull/271#discussion_r470520559 :brain:

librad/src/git/trailer.rs
FintanH: I realised that below :sweat_smile: I'll mark this as resolved.

librad/src/git/trailer.rs
NunoAlexandre: Not sure I understand your question :confused: sorry. When the inline value is empty, we don't want to add it to the list of inlined values, or else the final `Trailer::values` is a vec with an empty string when the trailer only has no value at all or a vec starting with an empty string when the trailer has multiline values. By doing this, we cover all scenarios and we end up with a sensible final `Trailer::values`.

librad/Cargo.toml
```diff
@@ -38,6 +39,7 @@ tracing = "0.1"
 urltemplate = "0.1"
 webpki = "0.21"
 
+
```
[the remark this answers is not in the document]
NunoAlexandre: It looks like an extra newline but in reality, it's a door to a world filled with barbells. Fixed!

librad/src/git/trailer.rs
NunoAlexandre: > and adding comment support to `(i)` is fine by me, although also not strictly required.

given that I'm about to go on holiday and that it's not strictly required I propose we postpone implementing it until/when it's needed.

NunoAlexandre:
> and adding comment support to `(i)` is fine by me, although also not strictly required.

given that I'm about to go on holiday and that it's not strictly required I propose we postpone implementing it until/when it's needed.

librad/src/git/trailer.rs
```diff
@@ -0,0 +1,245 @@
...
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct Trailer<'a> {
+    token: Token<'a>,
+    values: Vec<&'a str>,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct Token<'a>(pub &'a str);
```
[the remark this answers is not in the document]
FintanH: I believe he just meant `Trailer` and not `Token`.

> You already have all the necessary encapsulation by only allowing Token to be constructed through a "smart" constructor.

This sentence specifically entails that `Token` should be constructed via the smart constructor and not a `pub` field

librad/src/git/trailer.rs
FintanH: I think it's just my confusion of the logic. Clearly your tests pass with the single value case. But when I'm reading this it seems like it's thrown away.

Purely my confusion and not something that needs resolving right away :)

NunoAlexandre: Ok :) My confusion where first was as to what happened when there were no multiline values. But since those are `many0`, if there are none we get an empty vec, then if the inline value is not empty we add it to this empty vec, otherwise, the trailer value is just empty.

librad/src/git/trailer.rs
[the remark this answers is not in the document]
NunoAlexandre: You're right, my misunderstanding :)

librad/src/git/trailer.rs
NunoAlexandre: So back to this one, I honestly don't see the benefit of breaking this further now. The notion of separator _in this context_ is one of a given set of characters being present amongst possible one or more spaces.  Moving the part of the space higher up just increases complexity and detail where it is least needed (in this case) IMHO.

FintanH: Sure, ya that makes sense :+1:

NunoAlexandre:
@kim @FintanH, thank you both so far for all the reviews and help :hand: 

This PR is now ready for review. Quoting all the met requirements, we have done:

> * **Parse trailers**
>   Given a `&str`, return an iterator of `(String, String)` pairs, such that:
>   
>   * [x]  Trailing whitespace at the end of the string slice is ignored
>   * [x]  Only the last _paragraph_ of the string slice is considered
>   * [x]  A paragraph is preceded by one or more empty lines (ie. whitespace
>     including newlines), and does not itself contain empty lines (but it may
>     contain newline characters)
>   * [x]  The paragraph MUST consist of only trailers
>   * [x]  A valid token-value pair consists of:
>     
>     * [x]  A `token`, with no leading whitespace, composed of alphanumeric digits or '-'
>     * [x]  Optional horizontal whitespace
>     * [x]  A `separator` (default `:`, _MAY_ be configurable)
>     * [x]  Optional horizontal whitespace
>     * [x]  A `value` consisting of UTF-8 characters
>     * [x]  A newline or EOF


I am about to go on holiday leave so please feel free to follow up the PR shall you want it in before my return.

Sorry for the time spent first implementing this using primitive Rust. I am a big fan of [Make It Work Make It Right Make It Fast](https://wiki.c2.com/?MakeItWorkMakeItRightMakeItFast) and I first both wanted to have a good understanding of the problem and check whether we _really_ needed a parser combinator library to make it work decently. The lesson is taken :)

Regarding checking 'nom' out, after getting my head around it I find it to be a pretty nice parser combinator library. The naming of some functions are confusing, to say the least, and the reported issue around the sparsed and mixed docs across different versions is painful at first, but once you get in the flow it's pretty nice :ok_hand: 

Special thanks to @FintanH for helping me refresh my brain around parser combinators (it had been a while) :yellow_heart:

librad/src/git/trailer.rs
```diff
@@ -0,0 +1,256 @@
...
+
+    /// Parse a trailer, which can have an inlined or multilined value.
+    pub fn trailer<'a>(s: &'a str, separators: &'a str) -> IResult<&'a str, Trailer<'a>> {
+        let parser = separated_pair(token, |s| separator(separators, s), values);
+        let (rest, (token, values)) = parser(s)?;
+        Ok((rest, Trailer { token, values }))
+    }
+
+    /// Parse a trailer token.
+    fn token(s: &str) -> IResult<&str, Token> {
```
FintanH, line 92: Curious as to why these are private and not `pub`?

librad/src/git/trailer.rs
```diff
@@ -0,0 +1,256 @@
...
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct Token<'a>(&'a str);
+
+/// Parse the trailers of the given message. It looks up the last paragraph
+/// of the message and attempts to parse each of its lines as a [Trailer].
+/// Fails if no trailers paragraph is found or if at least one trailer
+/// fails to be parsed.
+pub fn parse<'a>(message: &'a str, separators: &'a str) -> Result<Vec<Trailer<'a>>, Error<'a>> {
+    let trailers_paragraph =
+        match parser::paragraphs(message.trim_end()).map(|(_, ps)| ps.last().map(Clone::clone)) {
```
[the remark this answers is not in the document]
NunoAlexandre: That's not the same actually. That would clone the `Option`, not the value it wraps. (I am 99.99% sure I tried that :b)

librad/src/git/trailer.rs
NunoAlexandre: No particular reason didn't _need_ to make them `pub`.

librad/src/git/trailer.rs
```diff
@@ -0,0 +1,256 @@
...
+
+    #[error(
+        "One or more trailers are not in the parseable format <token><separator><value>: '{0}'"
+    )]
+    Unparsable(&'a str),
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct Trailer<'a> {
+    token: Token<'a>,
```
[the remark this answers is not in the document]
NunoAlexandre: Uh, yes. Fixing...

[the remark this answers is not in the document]
NunoAlexandre: Done!

librad/src/git/trailer.rs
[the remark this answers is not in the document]
NunoAlexandre: hm it does work. I was doing something different then. Pushed :ok_hand:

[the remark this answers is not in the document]
FintanH: Not it's not `clone`. It's [`cloned`](https://doc.rust-lang.org/std/option/enum.Option.html#method.cloned)

PR #272, opened 2020-08-07
Add Remote struct

FintanH:
Working on top of #270 

Add a way of setting up a rad remote to the monorepo, setting it up with
any push specs specified and the single fetch spec.
This utilises a trait to turn Refspecs into Strings, so we can have
multiple in a Vec without worrying about their types lining up.

### DIFF
--- a/librad/src/git/types.rs
+++ b/librad/src/git/types.rs
@@ -115,10 +115,15 @@ impl<R> SymbolicRef<R> {
     }
 }
 
-/// A `Display`-like trait but only meant for `Refspec`s.
+/// If we want to be able to use a collection of [`Refspec`]s, where the remote
+/// portions don't line up, then we need to be able to talk about them using
+/// dynamically. In Rust this cab be done via `Box<dyn MyTrait>`.
 ///
-/// This allows us to treat them as opaque in collections, such a `Vec<Box<dyn
-/// AsRefspec>>`.
+/// In this case, we want to be able to turn [`Refspec`]s into `String`s. To
+/// avoid allowing anything that is `ToString` or `Display`, we opt for adding a
+/// `ToString`-like trait called `AsRefspec`.
+///
+/// **N.B.**: The only implementor of this trait should `Refspec`.
 pub trait AsRefspec {
     fn as_refspec(&self) -> String;
 }
--- a/librad/src/git/types.rs
+++ b/librad/src/git/types.rs
@@ -25,6 +25,7 @@ use crate::{git::refs::Refs, peer::PeerId, uri::RadUrn};
 
 pub mod existential;
 pub mod reference;
+pub mod remote;
 
 pub use existential::{SomeNamespace, SomeReference};
 pub use reference::{Multiple, Namespace, Reference, RefsCategory, Single};
@@ -114,6 +115,14 @@ impl<R> SymbolicRef<R> {
     }
 }
 
+/// A `Display`-like trait but only meant for `Refspec`s.
+///
+/// This allows us to treat them as opaque in collections, such a `Vec<Box<dyn
+/// AsRefspec>>`.
+pub trait AsRefspec {
+    fn as_refspec(&self) -> String;
+}
+
 #[derive(Clone)]
 pub struct Refspec<RemoteR, LocalR> {
     pub(crate) remote: SomeReference<RemoteR>,
@@ -122,12 +131,28 @@ pub struct Refspec<RemoteR, LocalR> {
     pub force: Force,
 }
 
+impl<R, L> Refspec<R, L> {
+    pub fn into_dyn(self) -> Box<dyn AsRefspec>
+    where
+        R: Display + Clone + 'static,
+        L: Display + Clone + 'static,
+    {
+        Box::new(self)
+    }
+}
+
 impl<R: Display + Clone, L: Display + Clone> Display for Refspec<R, L> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         if self.force.as_bool() {
             f.write_str("+")?;
         }
         write!(f, "{}:{}", self.remote, self.local)
+    }
+}
+
+impl<R: Display + Clone, L: Display + Clone> AsRefspec for Refspec<R, L> {
+    fn as_refspec(&self) -> String {
+        self.to_string()
     }
 }
 
--- a/librad/src/git/types.rs
+++ b/librad/src/git/types.rs
@@ -123,9 +123,19 @@ impl<R> SymbolicRef<R> {
 /// avoid allowing anything that is `ToString` or `Display`, we opt for adding a
 /// `ToString`-like trait called `AsRefspec`.
 ///
-/// **N.B.**: The only implementor of this trait should `Refspec`.
-pub trait AsRefspec {
+/// **N.B.**: The only implementor of this trait should be `Refspec`. We enforce
+/// this by making `AsRefspec` a [sealed
+/// trait](https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed).
+pub trait AsRefspec: sealed::Sealed {
     fn as_refspec(&self) -> String;
+}
+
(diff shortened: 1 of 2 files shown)


librad/src/git/types/remote.rs
```diff
@@ -0,0 +1,120 @@
...
+    use super::*;
+    use crate::{git::types::*, hash::Hash};
+    use librad_test::tempdir::WithTmpDir;
+
+    #[test]
+    fn can_create_remote() {
+        WithTmpDir::new::<_, io::Error>(|path| {
+            let repo = git2::Repository::init(path).expect("failed to init repo");
+
+            let id = Hash::hash(b"geez");
```
NunoAlexandre, line 85: TIL [`geez`](https://www.merriam-webster.com/dictionary/geez)

librad/src/git/types/remote.rs
```diff
@@ -0,0 +1,120 @@
...
+                git_remote
+                    .fetch_refspecs().expect("failed to get fetchspecs")
+                    .iter()
+                    .collect::<Vec<Option<&str>>>(),
+                vec![Some("+refs/namespaces/hwd1yredksthny1hht3bkhtkxakuzfnjxd8dyk364prfkjxe4xpxsww3try/refs/heads/*:refs/heads/*")],
+            );
+
+            assert_eq!(
+                git_remote
+                    .push_refspecs().expect("failed to get fetchspecs")
```
NunoAlexandre, line 111:
```suggestion
                    .push_refspecs().expect("failed to push fetchspecs")
```

librad/src/git/types/remote.rs
```diff
@@ -0,0 +1,120 @@
...
+
+
+            let mut remote = Remote::rad_remote(path, fetch);
+            remote.add_pushes(vec![push].into_iter());
+            let git_remote = remote.create(&repo).expect("failed to create the remote");
+
+
+            assert_eq!(
+                git_remote
+                    .fetch_refspecs().expect("failed to get fetchspecs")
```
NunoAlexandre, line 103:
```suggestion
                    .fetch_refspecs().expect("failed to fetch refspecs")
```

?

librad/src/git/types/remote.rs
```diff
@@ -0,0 +1,120 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use std::path::{Path, PathBuf};
+
+use super::AsRefspec;
+
+pub struct Remote {
+    /// The file path to the git monorepo.
+    pub monorepo: PathBuf,
```
kim, line 24: This is wrong. The general form would be `url: Url`, and the link form `url: git::local::LocalUrl`

FintanH: Yup, I thought as much. Was trying to get something working first.

> The general form would be url: Url, and the link form url: git::local::LocalUrl

I suppose I could parameterise it by `Url` then

librad/src/git/types.rs
```diff
@@ -114,6 +115,14 @@ impl<R> SymbolicRef<R> {
     }
 }
 
+/// A `Display`-like trait but only meant for `Refspec`s.
```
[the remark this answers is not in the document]
FintanH: I want to be able to turn `Refspec`s into `String`s dynamically but not allow anything that implements `ToString` or `Display`.

I'll try be more verbose in this doc :)

librad/src/git/types.rs
```diff
@@ -122,6 +131,16 @@ pub struct Refspec<RemoteR, LocalR> {
...
+impl<R, L> Refspec<R, L> {
+    pub fn into_dyn(self) -> Box<dyn AsRefspec>
+    where
+        R: Display + Clone + 'static,
+        L: Display + Clone + 'static,
+    {
+        Box::new(self)
+    }
+}
+
```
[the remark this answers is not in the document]
FintanH: I was having a bad time with type-inference doing this manually, so I introduced this for ergonomics. I can document it.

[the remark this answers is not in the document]
FintanH: And as far as I know, there's no `IntoDyn` trait :shrug: At least not in `std`. Not sure if it's worth introducing the trait here. What do you think?

librad/src/git/types/remote.rs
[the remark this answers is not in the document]
FintanH: This is actually out-of-date, as Kim pointed out that I was wrong here. It should be `Url`, and more specifically for us it will be a `LocalUrl`, which is of the form `rad://<namespace>.git`

librad/src/git/types/remote.rs
FintanH: Oh geez :see_no_evil:

librad/src/git/types/remote.rs
FintanH: Emmmm not quite actually. It's not doing an action. These are called fetch refspecs, and below are called push refspecs. I was short-handing them to fetchspecs and pushspecs.

librad/src/git/types.rs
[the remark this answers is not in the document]
NunoAlexandre: Yeah, I'm not sure, but since we introduced `AsRefspec`, introducing `IntoDyn` wouldn't hurt. Up to you :) 

> I was having a bad time with type-inference doing this manually, so I introduced this for ergonomics. I can document it.

I see. yeah leaving a comment will be helpful :ok_hand:

librad/src/git/types.rs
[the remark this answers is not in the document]
kim: Something about this trips me up -- there's a lot of generic machinery to make these things typed, but then we're actually unifying with `String`. Is there not some intermediate representation of a refspec which includes all the permutations on the value level (like, `Option` fields), so we can have _one_ `Display` impl? The job of the polymorphic types would then be to make construction of this _safe_ depending on the context.

librad/src/git/types/remote.rs
NunoAlexandre: Ah, I see the confusion! Thanks for explaining and sorry for the noise :)

librad/src/git/types.rs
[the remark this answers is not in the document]
FintanH: Well, there's only one `Display` impl. Whereas this is a bit of a hack to get collections of `Refspec`s without the types clashing.

> Is there not some intermediate representation of a refspec which includes all the permutations on the value level (like, Option fields)

Can you expand on what you mean a bit? I'm not quite following.

But I agree, it's a bit awkward because we don't have something like an open union to extend out the cases.

librad/src/git/types.rs
```diff
@@ -114,6 +115,19 @@ impl<R> SymbolicRef<R> {
...
 
+/// If we want to be able to use a collection of [`Refspec`]s, where the remote
+/// portions don't line up, then we need to be able to talk about them using
+/// dynamically. In Rust this cab be done via `Box<dyn MyTrait>`.
+///
+/// In this case, we want to be able to turn [`Refspec`]s into `String`s. To
+/// avoid allowing anything that is `ToString` or `Display`, we opt for adding a
+/// `ToString`-like trait called `AsRefspec`.
+///
+/// **N.B.**: The only implementor of this trait should `Refspec`.
```
kim, line 126: Ok, well, I guess this is good for now. You can enforce that nothing else but `Refspec` can be `AsRefspec` by going through the gymnastics of defining a [sealed trait](https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed)

FintanH: Nice trick :ok: :cartwheeling:

FintanH:
> I'm lighting a scent stick today for HKTs in Rust (and another one for rank-n-types).

Pour one out for HKTs :tumbler_glass:

FintanH:
Ah damn, I forgot I was targetting https://github.com/radicle-dev/radicle-link/pull/270 with this branch. @kim @NunoAlexandre could you take a look at those changes? :see_no_evil:

PR #277, opened 2020-08-13
Basic seed node functionality

cloudhead:
This is a long-running process that tracks URNs of interest.


radicle-seed/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "radicle-seed"
+version = "0.1.0"
+authors = ["The Radicle Team <dev@radicle.xyz>"]
+edition = "2018"
+license = "GPL-3.0-or-later"
+
+[dependencies]
+librad = { path = "../librad" }
+async-trait = "0.1"
```
xla, line 10: Came across this recently: https://crates.io/crates/async-trait-ext

radicle-seed/src/main.rs
```diff
@@ -0,0 +1,46 @@
...
+    pub log: log::Level,
+}
+
+impl Options {
+    pub fn from_env() -> Self {
+        argh::from_env()
+    }
+}
+
+fn main() {
```
xla, line 32:
```suggestion
#[tokio::main]
async fn main() {
```

This will give you an executor and you can await inside your main. EMBRACE LE ASYNC

cloudhead: This was needed to run the executor :grimacing:

radicle-seed/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+
+        let (api, future) = peer.accept()?;
+        let mut events = api.protocol().subscribe().await;
+        let mode = self.config.mode;
+
+        // Spawn the background peer thread.
+        tokio::spawn(future);
+
+        // Start by tracking specified projects if we need to.
+        if let Mode::TrackUrns(urns) = &mode {
```
FintanH, line 143: I think it might be good to extract these into smaller functions, if possible :)

radicle-seed/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+use argh::FromArgs;
+use futures::executor;
+use log;
+
+#[derive(FromArgs)]
+/// Radicle Seed.
+pub struct Options {
+    /// track the specified peers only
+    #[argh(option)]
+    pub track_peers: Vec<PeerId>,
```
FintanH, line 15: Should this not be `Mode`?

radicle-seed/src/lib.rs
```diff
@@ -0,0 +1,191 @@
+#[macro_use]
+extern crate async_trait;
```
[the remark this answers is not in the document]
kim: I like this style for a few very commonly used macro-crates (namely `async-trait` and `lazy-static` are `extern crate`d in `librad`.

radicle-seed/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+        // Listen on gossip events. As soon as a peer announces something of interest,
+        // we check if we should track it.
+        while let Some(event) = events.next().await {
+            match event {
+                ProtocolEvent::Gossip(gossip::Info::Has(gossip::Has { provider, val })) => {
+                    let urn = &val.urn;
+                    let peer_id = &provider.peer_id;
+
+                    if mode.is_trackable(peer_id, urn) {
+                        Node::track_project(&api, urn, &provider)?;
```
kim, line 168: I suppose you don't want to crash if that fails, just spit out something in the logs.

radicle-seed/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+        todo!()
+    }
+}
+
+#[derive(Debug)]
+pub enum Mode {
+    /// Track everything we see, no matter where it comes from.
+    TrackEverything,
+    /// Track everything from these peers, and nothing else.
+    TrackPeers(Vec<PeerId>),
```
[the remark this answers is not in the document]
cloudhead: I *KNEW* you would say this

radicle-seed/src/main.rs
```diff
@@ -0,0 +1,52 @@
...
+        root: opts.root,
+        mode: if opts.track_peers.is_empty() {
+            Mode::TrackEverything
+        } else {
+            Mode::TrackPeers(opts.track_peers)
+        },
+    };
+    let node = Node::new(config).unwrap();
+
+    executor::block_on(node.run()).unwrap();
```
xla, line 51:
```suggestion
    node.run().await.unwrap();
```

radicle-seed/src/main.rs
cloudhead: Not that easy to have an enum-like interface on the cli!

radicle-seed/src/lib.rs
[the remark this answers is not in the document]
FintanH: :innocent:

radicle-seed/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+}
+
+#[derive(Clone)]
+pub struct Signer {
+    key: keys::SecretKey,
+}
+
+impl Signer {
+    fn new() -> Self {
+        let key = keys::SecretKey::new();
```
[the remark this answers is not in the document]
FintanH: > Also, Signer should actually be implemented for SecretKey in librad (ie. move it there if it isn't already).

Ya it's implemented there :)

radicle-seed/src/lib.rs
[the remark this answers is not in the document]
cloudhead: I thought about it but it feels a bit overkill and requires more work to construct.. Having duplicates in there doesn't cause any problems either :man_shrugging: -- but also not set on either approach.

radicle-seed/src/main.rs
FintanH: `argh` what's even the point :shrug: 

So how would start it in the `RadUrn` mode?

radicle-seed/src/lib.rs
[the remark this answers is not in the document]
FintanH: Can't it be constructed from any `Iterator` via `collect()`?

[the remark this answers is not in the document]
cloudhead: Indeed!

radicle-seed/src/main.rs
cloudhead: Just hadn't added the option yet. It's possible now :+1:

radicle-seed/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+        listen_addr: opts.listen.unwrap_or(default.listen_addr),
+        mode: if opts.track_peers.is_empty() {
+            Mode::TrackEverything
+        } else {
+            Mode::TrackPeers(opts.track_peers)
+        },
+    };
+    let node = Node::new(config).unwrap();
+
+    executor::block_on(node.run()).unwrap();
```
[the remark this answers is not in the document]
cloudhead: :sweat_smile: I guess `node.run().await.unwrap();` should do it then

radicle-seed/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+    /// Track everything we see, no matter where it comes from.
+    TrackEverything,
+    /// Track everything from these peers, and nothing else.
+    TrackPeers(HashSet<PeerId>),
+    /// Track the specified URNs.
+    TrackUrns(HashSet<RadUrn>),
+}
+
+impl Mode {
+    fn is_trackable(&self, peer: &PeerId, urn: &RadUrn) -> bool {
```
FintanH, line 82: Would it make more sense to have `Either<&PeerId, &RadUrn>`? Like, why would we want to have to pass a `urn` if just want to check if a `peer` is trackable?

radicle-seed/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+thiserror = "1.0"
+tokio = "0.2"
+futures = "0.3"
+argh = "0.1.3"
+log = { version = "0.4", features = ["std"] }
```
kim, line 15: Most likely you'll want to use [tracing](https://crates.io/crates/tracing) / [tracing-subscriber](https://crates.io/crates/tracing-subscriber) instead, to be able to debug the lower bowels.

radicle-seed/src/lib.rs
cloudhead: Because in the end, it's the URN that you are tracking, not the peer. So you always want to be passing in the URN

FintanH: Ah, I think I'm following. And you always have a `PeerId` because you need to ask for this from some peer? Or should _that_ be optional?

cloudhead: You always have it too in practice. You could *theoretically* have a bare URN, but can't think of a useful context where that would occur.

radicle-seed/Cargo.toml
cloudhead: Done.

radicle-seed/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
```
FintanH, line 15: Should you not remove the `log` dependency now?

radicle-seed/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+            .storage()
+            .clone_repo::<project::ProjectInfo, _>(url, addr_hints)
+            .and_then(|_| api.storage().track(urn, &peer_id));
+
+        match &result {
+            Ok(()) => {
+                tracing::info!("Successfully tracked project {} from peer {}", urn, peer_id,);
+            },
+            Err(err) => {
+                tracing::debug!(
```
FintanH, line 203: Why not `warn!`?

cloudhead: Because it's fairly normal for this to fail. A `warn` would imply that something could be wrong with the system and it needs to be checked out. Maybe @kim has a different opinion on this though.

radicle-seed/Cargo.toml
cloudhead: Done!

radicle-seed/src/lib.rs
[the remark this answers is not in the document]
cloudhead: Done in 9d04d7a

However, trying to test this with a secret key generated from `SecretKey` gives me:
```
thread 'main' panicked at 'A valid PKCS#8 document is valid. qed: CouldNotParseKeyPair', /home/cloudhead/src/radicle-dev/radicle-link/librad/src/net/tls.rs:65:28
```

It does parse properly though:
```
> echo -n 'cdb47a4eefeb8922d953e312785521e97f5bc0b9ce66ba07037c0f1086f9a37d' | cargo run -- --root .
```

radicle-seed/src/lib.rs
cloudhead: Done.

cloudhead:
This is now "feature complete" as far as I can tell, modulo the key input which I'm not sure I'm doing right.

I think the next step would be to run this on a real network and do another iteration on the feature after that.

radicle-seed/src/lib.rs
```diff
@@ -0,0 +1,271 @@
...
+}
+
+#[derive(Clone)]
+pub struct Signer {
+    key: keys::SecretKey,
+}
+
+impl Signer {
+    pub fn new<R: io::Read>(mut r: R) -> Result<Self, io::Error> {
+        use radicle_keystore::SecretKeyExt;
```
xla, line 66: Why is this `use` inlined?

radicle-seed/src/lib.rs
```diff
@@ -0,0 +1,271 @@
...
+
+impl keys::AsPKCS8 for Signer {
+    fn as_pkcs8(&self) -> Vec<u8> {
+        self.key.as_pkcs8()
+    }
+}
+
+/// Seed operational mode.
+#[derive(Debug)]
+pub enum Mode {
```
xla, line 100: Minor nit, but instead of prefixing `Track` to every variant you coul rename it to `TrackMode`.

cloudhead: True, I think we will eventually change this anyway to `TrackRule` and have them be a Monoid of sorts

radicle-seed/src/lib.rs
cloudhead: No particular reason except that it's only used here and is kind of esoteric

radicle-seed/src/lib.rs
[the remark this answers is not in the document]
kim: :thinking: What's that string you're `echo`ing? The `Display` impl for `SecretKey` outputs the public part (for whatever reason, and hex-encoded for wtf reason), which looks suspiciously like this. What `SecretKey::from_bytes_and_meta` expects are the actual (RFC 8032) bytes (I know `SecStr` is an unfortunate name, it's actually `type SecStr = SecVec<u8>`).

It's been bugging me for a while that there doesn't seem to be a way to validate that a byte slice is an actual ed25519 key (public or secret) -- libraries just check if it's 32 bytes. Perhaps the proper way is to pass the public key as a "checksum" as a CLI option.

radicle-seed/src/lib.rs
kim: We will find out when things don't work as expected, and `RUST_LOG=debug` is too noisy :)

radicle-seed/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+
+                for urn in urns {
+                    let mut peers = api.providers(urn.clone()).await;
+                    // Attempt to track until we succeed.
+                    while let Some(peer) = peers.next().await {
+                        if Node::track_project(&api, urn, &peer).is_ok() {
+                            break;
+                        }
+                    }
+                }
```
kim, line 258:
```suggestion
                futures::stream::iter(urns)
                    .for_each_concurrently(Some(urns.len()), |urn| {
                        let mut peers = api.providers(urn.clone()).await;
                        tokio::time::timeout(Duration::from_secs(120), {
                            // Attempt to track until we succeed.
                            while let Some(peer) = peers.next().await {
                                if Node::track_project(&api, urn, &peer).is_ok() {
                                    break;
                                }
                            }
                        }).unwrap_or_else(|| tracing::warn!("Timeout trying to track {}", urn))
                    }) 
```

(not sure if typechecks)

cloudhead:
Sounds good!

It doesn't seem like there's consensus in the community about which way to go re:directories, so it's up to us. We should tackle this in one go for radicle-tracker and radicle-seed.

xla:
> One nit though: I'd _really_ prefer to **not** redundantly prefix every crate in this repo with `radicle-`. Cargo doesn't require that the directory name has to match the crate name, and the prefix is just rather annoying for navigation. I know this is handled somewhat inconsistently in the ecosystem (and, accordingly, here), but let's just not over-radicle-ize.

+ 1111. It is sufficient if the respective `Cargo.toml`s have the prefix in the crate name and the directories in the workspace are unprefixed.

radicle-seed/src/lib.rs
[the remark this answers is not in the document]
cloudhead: Aha! Yeah it's basically `println!("{}", SecretKey::new())`, so that explains it. I guess it's a valid curve point but not a valid private key, or something like that. The checksum idea is interesting. Otherwise it could work a bit like `ssh -i`, where it would look for the `.pub`..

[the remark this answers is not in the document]
kim: > I guess it's a valid curve point

It's literally only checked if the input is 64 bytes long. `ed25519-zebra` uses the seed value for byte representations, which is safer, but not usable for us as we need to be able to convert to other wrapper types.


> Otherwise it could work a bit like `ssh -i`

There is actually another subtlety: once the `Storage` has been initialised, it is [tied to the key used for that](https://github.com/radicle-dev/radicle-link/blob/895e73438e6ec910d2054ad80e21571305b2c716/librad/src/git/storage.rs#L501). Perhaps we should [bootstrap](https://github.com/radicle-dev/radicle-link/blob/895e73438e6ec910d2054ad80e21571305b2c716/librad/src/net/peer.rs#L304-L313) the storage first, as the error might be more meaningful.

So yeah, I'm not sure. I guess the best we can do when constructing a `SecretKey` from a slice is to sign some random data, and see if that verifies.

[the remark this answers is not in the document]
kim: > the seed value

Well, disregard this, it's the same thing: 32 random bytes

PR #278, opened 2020-08-13
String type for Canonical JSON

kim:
Here is a picture of a yak:


kim:
Canonical form:

:water_buffalo: :haircut: 

---
Become a sponsor! https://home.unicode.org/adopt-a-character/

PR #279, opened 2020-08-14
Two tests with register break one or the other

FintanH:


FintanH:
Closed in favour of #280

PR #280, opened 2020-08-17
Use a HashMap for shared Settings ref

FintanH:
The local transport was using a global mutex on `Option<Settings>`. In
concurrent runs tests would end up overwriting each other's version of
this mutex and would end up in inconsistent states.

To solve this we need to separate test case's runs. Initially, we
attempted to use another lock `Mutex<()>` and have a bracketed runner,
which would lock, fill the `Option<Settings>`, run the action, and clean
up & release the lock. This still ended up in race conditions.

This solution uses a HashMap from PeerId to Settings. This works since
we expect individual runs to have separate PeerIds. The consequence of
this is that we need a way to look up the Settings when running the git
action. We encode this in the LocalUrl, so we add the PeerId as the
username of the Url that's passed.

### DIFF
--- a/librad/src/git/local/url.rs
+++ b/librad/src/git/local/url.rs
@@ -24,23 +24,42 @@ use thiserror::Error;
 
 use crate::{
     hash::{self, Hash},
+    peer::{self, PeerId},
     uri::{self, RadUrn},
 };
 
 #[derive(Clone)]
 pub struct LocalUrl {
     repo: Hash,
+    peer_id: PeerId,
 }
 
 impl LocalUrl {
     pub fn repo(&self) -> &Hash {
         &self.repo
     }
+
+    pub fn peer_id(&self) -> &PeerId {
+        &self.peer_id
+    }
+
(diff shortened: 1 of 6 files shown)


librad/src/git/local/url.rs
```diff
@@ -24,23 +24,42 @@ use thiserror::Error;
...
+        &self.peer_id
+    }
+
+    pub fn from_urn(urn: RadUrn, peer_id: PeerId) -> Self {
+        Self {
+            repo: urn.id,
+            peer_id,
+        }
+    }
 }
```
kim, line 52: Nit: for consistency with `git::p2p::url::GitUrl`, let's just make the fields `pub` and name `peer_id` `local_peer_id`

FintanH:
> Oof the pain. I'm working on convincing libgit2 to make auth.h part of the public API. This means a programmatic way for custom transports to ask for credentials, and being able to use the credentials API.

That would be so much better :bow: The pain is real...

PR #290, opened 2020-08-28
Trailers: turn values into cows

kim:
Converting some type into a `Trailer` likely involves some
serialisation, which yields an owned string.


FintanH (approved):
Mooooo-ve over references

PR #292, opened 2020-08-31
Trailers: rendering (via Display), with customisable separator

kim:
There are some lifetime gymnastics involved here, not all of which I was able to
sort out. Specifically, I'd prefer to implement `DisplayMany` for any opaque
`Iterator`, but oh my...

Also note how disappointingly type inference fails if you pass a closure to
compute the separator.


FintanH (approved):
Cool :+1: 

Clippy is annoyed fyi :)

FintanH (approved):

FintanH (approved):

Issue #282, opened 2020-08-17
Rename top-level crates


cloudhead:
We currently use the `radicle-` prefix for some top-level crates, eg. `radicle-tracker` and `radicle-seed`. These should be renamed to simply `tracker` and `seed`, as explained in https://github.com/radicle-dev/radicle-link/pull/277#pullrequestreview-468492849

PR #294, opened 2020-08-31
Drop the 'radicle' prefix from top-level crates

NunoAlexandre:
Closes https://github.com/radicle-dev/radicle-link/issues/282


kim:
I think the crate name itself (like, `name` in `Cargo.toml`) should still have the prefix, as we want to eventually some day publish them to crates.io. It's just that the directory doesn't need the prefix, which makes it more pleasant to use tab-completion and other navigation tools.

FintanH (approved):
Thanks, Nuno :)

xla:
> I think the crate name itself (like, `name` in `Cargo.toml`) should still have the prefix, as we want to eventually some day publish them to crates.io. It's just that the directory doesn't need the prefix, which makes it more pleasant to use tab-completion and other navigation tools.

:+1:

NunoAlexandre:
> I think the crate name itself (like, `name` in `Cargo.toml`) should still have the prefix, as we want to eventually some day publish them to crates.io. It's just that the directory doesn't need the prefix, which makes it more pleasant to use tab-completion and other navigation tools.

That sounds sensible, sorry for overlooking.

Cargo.toml
```diff
@@ -3,6 +3,6 @@ members = [
   "librad",
   "librad-test",
   "git-helpers",
-  "radicle-tracker",
-  "radicle-seed",
+  "tracker",
+  "seed",
```
xla, line 7:
```suggestion
  "radicle-tracker",
  "radicle-seed",
```

They are still referenced by the crate name in cargo, instead of the directory name.

NunoAlexandre: Not really. If you apply that change, cargo will fail with: 

```
error: failed to read `/home/nuno/radicle-link/radicle-tracker/Cargo.toml`

Caused by:
  No such file or directory (os error 2)
```

xla: :O My bad, it's the case for declaration deps. Disregard and sorry for the noise.

NunoAlexandre: No worries mate :)

FintanH (approved):
~radicle-~ approved

xla (commented):
💲 ⏫ 🐁 🚩

xla (approved):
💲 ⏫ 🐁 🚩

PR #297, opened 2020-09-01
Relevant peer announcements

### LINKED ISSUE #295: [applying changes] Relevant peer announcements
Currently, when we receive an announcement from the network we will consider it relevant even if the peer isn't one of our tracked remotes.

We would prefer that it wasn't greedy, and rather would only apply changes when it's from a peer that we track for this `RadUrn`, but still gossip the message across the rest of our network.

NunoAlexandre:
Closes #295

### DIFF
--- a/librad/src/git/storage.rs
+++ b/librad/src/git/storage.rs
@@ -994,10 +994,9 @@ where
     }
 
     pub fn is_tracked(&self, urn: &RadUrn, peer: &PeerId) -> bool {
-        match self.backend.find_remote(&tracking_remote_name(urn, peer)) {
-            Ok(_) => true,
-            Err(_) => false,
-        }
+        self.backend
+            .find_remote(&tracking_remote_name(urn, peer))
+            .is_ok()
     }
 }
 
--- a/librad/src/git/storage.rs
+++ b/librad/src/git/storage.rs
@@ -993,14 +993,12 @@ where
         Ok(())
     }
 
-    // TODO(nuno): quoting Kim: [this] is a bit too optimistic I think
-    // -- git2 unifies everything into a single Error enum, but we only
-    // want to return false if the ErrorCode is NotFound, and otherwise
-    // bubble up the error.
-    pub fn is_tracked(&self, urn: &RadUrn, peer: &PeerId) -> bool {
-        self.backend
-            .find_remote(&tracking_remote_name(urn, peer))
-            .is_ok()
+    pub fn is_tracked(&self, urn: &RadUrn, peer: &PeerId) -> Result<bool, Error> {
+        match self.backend.find_remote(&tracking_remote_name(urn, peer)) {
+            Ok(_) => Ok(true),
+            Err(e) if is_not_found_err(&e) => Ok(false),
+            Err(e) => Err(Error::from(e)),
+        }
     }
 }
 
--- a/librad/src/git/storage.rs
+++ b/librad/src/git/storage.rs
@@ -993,6 +993,10 @@ where
         Ok(())
     }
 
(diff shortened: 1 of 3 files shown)


librad/src/git/storage.rs
```diff
@@ -992,6 +992,12 @@ where
 
         Ok(())
     }
+
+    pub fn is_tracked(&self, urn: RadUrn, peer: PeerId) -> bool {
```
kim, line 996: Can be done in `O(1)` by `self.backend.find_remote(&tracking_remote_name(urn, peer))`. Also, passing references is preferable I think.

NunoAlexandre: Thanks. I had checked `find_remote` but had not seen `tracking_remote_name` :+1:

librad/src/git/storage.rs
```diff
@@ -992,6 +992,16 @@ where
 
         Ok(())
     }
+
+    // TODO(nuno): quoting Kim: [this] is a bit too optimistic I think
+    // -- git2 unifies everything into a single Error enum, but we only
+    // want to return false if the ErrorCode is NotFound, and otherwise
```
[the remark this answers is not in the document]
NunoAlexandre: Done. I felt a bit resistant in doing this since it adds a lot of bloat while the bubbled up error is not fully propagated, but it is sensible to return `Result<bool, Error>` anyways.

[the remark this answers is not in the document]
FintanH: Strictly better than swallowing all errors :)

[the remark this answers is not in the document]
NunoAlexandre: Yes indeed :)

NunoAlexandre:
@kim @FintanH could you have a look?

librad/src/net/peer.rs
```diff
@@ -436,43 +436,47 @@ where
...
+                    Err(e) => {
+                        tracing::error!(err = %e, "Git::Storage::is_tracked error");
+                        return PutResult::Error;
+                    },
+                };
                 let res = match has.rev {
                     // TODO: may need to fetch eagerly if we tracked while offline (#141)
-                    None => PutResult::Uninteresting,
-                    Some(Rev::Git(head)) => {
+                    Some(Rev::Git(head)) if is_tracked => {
```
kim, line 449: I think we can actually fix the TODO: if we only check `is_tracked`, we'll also fetch if we haven't yet fetched any data from the tracked peer.

librad/src/git/storage.rs
```diff
@@ -992,6 +992,14 @@ where
 
         Ok(())
     }
+
+    pub fn is_tracked(&self, urn: &RadUrn, peer: &PeerId) -> Result<bool, Error> {
```
FintanH, line 996: It would be great to document new functions :pray:

FintanH (approved):
Nice work :raised_hands:

librad/src/git/storage.rs
xla: Why would be catched with a `#[warn(missing_docs)]` and `--D warnings` in the CI run.

xla (approved):
🎑 ♑️ ⛲️ 🍚

librad/src/git/storage.rs
FintanH: That's a bigger effort though, currently.

xla: #302

xla (approved):
😗 🌜 🖼 🍿

librad/src/net/peer.rs
NunoAlexandre: Addressing this in a follow up PR.

PR #300, opened 2020-09-02
Include Utilities

### LINKED ISSUE #287: [working copy] Remote Includes
The [**Working Copies**](https://github.com/radicle-dev/radicle-link/blob/master/docs/rfc/identity_resolution.md#working-copies) section of the identity resolution hints at how we may be able to manage remotes from `librad` as opposed to any upstream clients. The proposal is that it should take advantage of [`git-config`](https://git-scm.com/docs/git-config).

My mini-proposal here is the following. For every `Project` namespace created there MUST be one include file. The include file MUST contain one remote configuration, which is the `rad` remote, i.e. the local peer's remote branch to the monorepo.
The following are operations on the include file:
  * If a **peer** is **added to the remote** of the namespace then we MUST **add it to the include file**.
  * If a **peer** is **removed from the remote** of the namespace then we MUST **remove it from the include file**.

The upstream users of `librad` MAY then include the namespace include file as:
```
[include]
    path = /path/to/<namespace>.inc
```

Suggestions:
* We may want to keep https://github.com/radicle-dev/radicle-link/issues/235 in mind when adding the remotes.
* It would be ideal to have a data structure where we can serialise and deserialise the include data.

FintanH:
Fixes #287 

Adds utilities to create an include file that has a list of remotes. This will allow upstream users to generate include files for their working directories to keep track of remotes.

### DIFF
--- a/librad/src/git/include.rs
+++ b/librad/src/git/include.rs
@@ -1,0 +1,228 @@
+// This file is part of radicle-link
+// <https://github.com/radicle-dev/radicle-link>
+//
+// Copyright (C) 2019-2020 The Radicle Team <dev@radicle.xyz>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 or
+// later as published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use std::{
+    convert::TryFrom,
+    marker::PhantomData,
+    path::{self, PathBuf},
+};
+
+use git2::{Config, ConfigLevel};
+
+use crate::{
+    git::{
+        local::url::LocalUrl,
+        types::{remote::Remote, AsRefspec, FlatRef, Force},
+    },
+    meta::user::User,
+    peer::PeerId,
+};
+
+/// An `Include` is a representation of an include file which we want to
+/// generate for working copies.
+///
+/// When generated it will include a list of remotes of the form:
+/// ```text
+/// [remote "<handle>@<peer_id>"]
+///     url = rad://<local_peer_id>@<repo>.git
+///     fetch = refs/remotes/<peer_id>/heads/*:refs/remotes/<handle>@<peer_id>/*
+/// ```
+///
+/// This file can then be added to the working copy's `config` file as:
+/// ```text
+/// [include]
+///     path = <path>/<repo>.inc
+/// ```
+pub struct Include<Path> {
+    /// The list of remotes that will be generated for this include file.
+    pub remotes: Vec<Remote<LocalUrl>>,
+    /// The directory path where the include file will be stored.
+    pub path: Path,
+    /// The namespace and `PeerId` this include file is interested in. In other
+    /// words, the generated include file should be for some project that
+    /// lives under this namespace, of this URL, in the monorepo.
+    ///
+    /// Note that the final file name will be named after
+    /// the namespace.
+    pub local_url: LocalUrl,
+}
+
+impl<Path> Include<Path> {
+    /// Create a new `Include` with an empty set of remotes.
+    pub fn new(path: Path, local_url: LocalUrl) -> Self {
+        Include {
+            remotes: vec![],
+            path,
+            local_url,
+        }
+    }
+
+    /// Create the include file by creating a [`Config`].
+    pub fn create(self) -> Result<Config, git2::Error>
+    where
+        Path: AsRef<path::Path>,
+    {
+        git2::Config::try_from(self)
+    }
+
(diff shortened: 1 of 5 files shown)


FintanH:
@kim Ideally, we'd probably have a CLI for this as well. Any opinion on having the exe in `librad` vs another workspace?

librad/src/git/include.rs
```diff
@@ -0,0 +1,239 @@
...
+        types::{remote::Remote, AsRefspec, FlatRef, Force, Reference},
+    },
+    hash::Hash,
+    meta::user::User,
+    peer::PeerId,
+    uri::RadUrn,
+};
+
+/// An `Include` is a representation of an include file for which we want to
+/// generate for working copies.
```
xla, line 38:
```suggestion
/// An `Include` is a representation of an include file which we want to
/// generate for working copies.
```

xla (dismissed):
utACK

💅 🙏 ✌️ 👰

xla (dismissed):
🌒 👼 🍟 🚷

librad/src/git/include.rs
```diff
@@ -0,0 +1,239 @@
...
+    /// Create the include file by creating a [`Config`].
+    pub fn create(self) -> Result<Config, git2::Error>
+    where
+        Path: AsRef<path::Path>,
+    {
+        git2::Config::try_from(self)
+    }
+
+    /// Convenience for adding an iterator of `Remote`s.
+    pub fn add_remotes(&mut self, remotes: impl Iterator<Item = Remote<LocalUrl>>) {
```
kim, line 85: If you put an `IntoIterator` bound here, the caller can pass a collection type or an iterator.

librad/src/git/include.rs
```diff
@@ -0,0 +1,239 @@
...
+};
+
+/// An `Include` is a representation of an include file which we want to
+/// generate for working copies.
+///
+/// When generated it will include a list of remotes of the form:
+/// ```text
+/// [remote "<handle>@<peer_id>"]
+///     url = rad://<repo>@<peer_id>.git
+///     fetch = refs/heads/*:refs/remotes/<repo>/*
```
kim, line 44: This doesn't look right. Wasn't the idea to generate a remote per tracked peer of a project? So I'd expect it to look more like:

```
[remote "<handle>@<peer_id>"]
    url = rad://<local_peer_id_because_libgit>@<urn.id>.git
    fetch = refs/remotes/<peer_id>/refs/heads/*:refs/remotes/<handle>@<peer_id>/*
```

It's possible that the types ensure that, but it's hard to tell from this patch alone.

kim (commented):
> Ideally, we'd probably have a CLI for this as well.

Can we just revive the CLI as a top-level `radicle-cli` project already? Even if this is just a collection of binaries for now, the build can produce binary artifacts. I didn't quite understand why that wasn't considered when #254 landed.

librad/src/git/include.rs
FintanH: oooh thanks!

FintanH:
> Can we just revive the CLI as a top-level radicle-cli project already? Even if this is just a collection of binaries for now, the build can produce binary artifacts. I didn't quite understand why that wasn't considered when #254 landed.

Sounds fine with me :+1: So just to be clear, these utilities would live in `librad` and then we can make an (set of) exe(s) in `radicle-cli`?

librad/src/git/include.rs
FintanH: Mmmm you're right. Lemme double-check this!

kim:
> So just to be clear, these utilities would live in librad

Yaaa up to you. I think that, since this uses a lot of our internal wrapper types, you'd end up factoring out all of those into a separate crate if both the cli and librad would depend on them. So I guess less hassle to go with what you say for now.

FintanH:
> Yaaa up to you. I think that, since this uses a lot of our internal wrapper types, you'd end up factoring out all of those into a separate crate if both the cli and librad would depend on them. So I guess less hassle to go with what you say for now.

True.. I have had the thought that we could separate out all our nice git machinery :grimacing: But I don't want to take that on right now.

That aside, I'll create the executable in a follow-up PR :)

kim:
> I'll create the executable in a follow-up PR 

I actually meant a separate repo -- we still have this unsolved problem of having a Cargo.lock checked in, which we shouldn't because this is a libary. I don't really know how to ensure that `master` doesn't suddenly break, but perhaps this is simply to gatekeep all merges such that they occur sequentially.

FintanH:
> I actually meant a separate repo

Ah! Ya, even better. The factor out comment makes more sense now too :joy: I'll create that repository after this goes in so. Can't wait to have to deal with the whole CI setup shenanigans again :japanese_goblin:

librad/src/git/include.rs
FintanH: Added a test and found a bug I'd left :sweat_smile: https://github.com/radicle-dev/radicle-link/pull/300/commits/373c80b032dd9bb3a3e19536e6ab17b064b3e7a3

xla (commented):
utACK

🌞 🏘 ⚓️ 📩

xla (dismissed):

xla (approved):
🔊 🐧 🚱 🤘

kim (approved):

PR #301, opened 2020-09-03
Identities: generic types and verification

kim:
Aims to define all git-independent aspects of the identities system
generically, such that we can reason about it sans IO. The respective
modules are not part of the surface API of `librad`, and may later be
hidden (modulo `rustdoc` rendering).

### DIFF
--- a/librad/src/identities/delegation/indirect.rs
+++ b/librad/src/identities/delegation/indirect.rs
@@ -1,0 +1,228 @@
+// This file is part of radicle-link
+// <https://github.com/radicle-dev/radicle-link>
+//
+// Copyright (C) 2019-2020 The Radicle Team <dev@radicle.xyz>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 or
+// later as published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use std::{
+    collections::{
+        btree_map::{self, Entry},
+        BTreeMap,
+        BTreeSet,
+    },
+    fmt::{Debug, Display},
+};
+
+use either::Either;
+
+use crate::keys::PublicKey;
+
+use super::{generic, sealed, Delegations, Direct};
+
+pub mod error {
+    use std::fmt::{Debug, Display};
+
+    use thiserror::Error;
+
+    use super::PublicKey;
+
+    #[derive(Debug, Error)]
+    pub enum FromIter<R: Display + Debug> {
+        #[error("duplicate key `{0}`")]
+        DuplicateKey(PublicKey),
+
+        #[error("duplicate identity with root `{0}`")]
+        DuplicateIdentity(R),
+    }
+
+    #[derive(Debug, Error, Eq, PartialEq)]
+    #[error("double vote")]
+    pub struct DoubleVote;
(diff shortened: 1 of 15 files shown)


librad/src/identities/sign.rs
```diff
@@ -0,0 +1,201 @@
...
+    sig: keys::Signature,
+}
+
+impl From<(PublicKey, keys::Signature)> for Signature {
+    fn from((key, sig): (PublicKey, keys::Signature)) -> Self {
+        Self { key, sig }
+    }
+}
+
+impl TryFrom<Trailer<'_>> for Signature {
```
kim, line 51: Uh well, this actually belongs in the next patch, but whatever.

librad/src/identities/generic/tests.rs
```diff
@@ -0,0 +1,201 @@
...
+        )
+    }
+
+    #[test]
+    fn quorum_below_threshold(id in gen_identity::<Boring>()) {
+        // Hitting the threshold exactly is one too few
+        let threshold = id.quorum_threshold();
+        let signatures: Signatures = BTreeMap::from(id.signatures)
+            .into_iter()
+            .take(threshold)
```
FintanH, line 64: Is there any way we can test `1..threshold` for this? Technically we're only ever testing `signatures == threshold` but not `< threshold`, bar the `0`-case above.

FintanH (approved):
Sick bifunctor :clap: 

Don't think I can say much more than nice work :raised_hands:

librad/src/identities/delegation/indirect.rs
```diff
@@ -0,0 +1,228 @@
...
+                Either::Left(key) => insert(key, None)?,
+                Either::Right(id) => {
+                    if !roots.insert(id.root.clone()) {
+                        return Err(DuplicateIdentity(id.root));
+                    }
+
+                    ids.push(id);
+                    let pos = ids.len() - 1;
+
+                    for key in &ids[pos].doc.delegations {
```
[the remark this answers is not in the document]
kim: Yeah, `ids.push(id)` moves the `id`. But I also need to get at `pos`, which I only know after I `push`ed.

librad/src/identities/generic.rs
```diff
@@ -0,0 +1,543 @@
...
+#[cfg(test)]
+pub(crate) mod tests;
+
+/// The identity document, carrying metadata `T` and trust delegations `D`.
+///
+/// In `git`, this is represented as a `blob`, where the previous revision
+/// `replaces` is a `tree` oid.
+#[derive(Clone, Debug, PartialEq, serde::Deserialize)]
+pub struct Doc<T, D, Revision> {
+    /// Protocol version. Always serialised as `0` (zero).
```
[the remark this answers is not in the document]
kim: Das Spezifikation says that the protocol version is zero. I wanted to avoid `new` and builders and [s|g]etters just because of that. We could also omit the field entirely, but then we'll have to remember to add it if and when the version is zero+1. So, whatever you put there, it'll come out as zero :)

librad/src/identities/generic/tests.rs
kim: Hm good point, let me `prop_flat_map` something

xla (dismissed):
Extra dope! utACK

👵 🎁 🕙 🚛

librad/src/identities/generic/tests.rs
kim: PTAL: d72c71c5a6e6723615b4862858d8e5e8b28ed7b5

kim:
I would absolutely welcome contributions to the property test suite -- there is more variance to be explored, but I'm suffering a bit of `proptest` fatigue (needed to increase my scrollback buffer to get at the actual error messages, for instance).

FintanH (approved):

xla (dismissed):
🌾 〽️ 🙌 🐴

kim:
Merged due to backporting from #310 would be too much work.

PR #305, opened 2020-09-03
Add connected check to protocol

### LINKED ISSUE #298: Add connected check to protoocl
For operators of peers to have an easy way to check for connectivity in the network we should expose something like:  `
pub async fn is_connected(&self) -> bool { self.connections.lock().await.len() }` to the `Protocol`.

xla:
Closes #298

### DIFF
--- a/librad/src/net/protocol.rs
+++ b/librad/src/net/protocol.rs
@@ -193,6 +193,11 @@ where
         self.gossip.announce(have).await
     }
 
+    /// Returns `true` if there is at lest one active connection.
+    pub async fn is_connected(&self) -> bool {
+        !self.connections.lock().await.is_empty()
+    }
+
     /// Query the network for an update
     ///
     /// Answers from the network will be available as `ProtocolEvent::Gossip`


librad/src/net/protocol.rs
```diff
@@ -193,6 +193,11 @@ where
         self.gossip.announce(have).await
     }
 
+    /// Returns `true` if there is at lest one active connection.
```
NunoAlexandre, line 196:
```suggestion
    /// Returns `true` if there is at least one active connection.
```

NunoAlexandre (commented):
Just a typo :ice_skate:

librad/src/net/protocol.rs
```diff
@@ -193,6 +193,11 @@ where
         self.gossip.announce(have).await
     }
 
+    /// Returns `true` if there is at least one active connection.
+    pub async fn is_connected(&self) -> bool {
```
FintanH, line 197: What do we think about giving it a parameter for a threshold? Like we only consider ourselves connected if `x` connections are found?

xla: I'd like to hear a use-ase for that requirement. Because if we go that specific we could alternatiely just implement a method which returns the number of currently active connections.

kim: Ya, `num_connected -> usize` may also be useful.

kim (dismissed):

librad/src/net/protocol.rs
xla: Happy to change it, if we think it's the better API. @kim @NunoAlexandre Any strong feelings on this?

FintanH: > may also be useful.

Could do both, indeed :)

xla: How excited are we about 749b657?

FintanH (approved):
This <======================================================================================================================================================================================================================================================> Excited

librad/src/net/protocol.rs
kim: In the spirit of Rust, both

kim (approved):

librad/src/net/protocol.rs
FintanH: "Clippy: is_empty should be implemented with len"

PR #309, opened 2020-09-08
Test PeerApi::providers

NunoAlexandre:
Test the `PeerApi::providers` function

### DIFF
--- a/librad/src/net/gossip.rs
+++ b/librad/src/net/gossip.rs
@@ -822,7 +822,6 @@ where
         let excluding = excluding.into();
 
         let mut connected_peers = self.connected_peers.lock().await;
-
         futures::stream::iter(
             connected_peers
                 .iter_mut()
--- a/librad/src/net/gossip.rs
+++ b/librad/src/net/gossip.rs
@@ -818,6 +818,8 @@ where
         M: Into<Rpc<Addr, Broadcast>>,
         X: Into<Option<&'a PeerId>>,
     {
+        tracing::trace!("Gossip::broadcast");
+
         let rpc = rpc.into();
         let excluding = excluding.into();
 
--- a/librad/src/net/gossip.rs
+++ b/librad/src/net/gossip.rs
@@ -818,12 +818,11 @@ where
         M: Into<Rpc<Addr, Broadcast>>,
         X: Into<Option<&'a PeerId>>,
     {
-        tracing::trace!("Gossip::broadcast");
-
         let rpc = rpc.into();
         let excluding = excluding.into();
 
         let mut connected_peers = self.connected_peers.lock().await;
+
         futures::stream::iter(
             connected_peers
                 .iter_mut()
--- a/librad/src/net/peer.rs
+++ b/librad/src/net/peer.rs
@@ -195,9 +195,9 @@ where
         &self,
         urn: RadUrn,
     ) -> impl Future<Output = impl futures::Stream<Item = PeerInfo<IpAddr>>> {
+        let span = tracing::trace_span!("PeerApi::providers", urn = %urn);
         let protocol = self.protocol.clone();
         let target_urn = urn.clone();
-        let span = tracing::trace_span!("Peer::providers");
 
         async move {
             // Listen for `Has` gossip events where a provider
--- a/librad/src/net/peer.rs
+++ b/librad/src/net/peer.rs
@@ -195,38 +195,34 @@ where
         &self,
         urn: RadUrn,
     ) -> impl Future<Output = impl futures::Stream<Item = PeerInfo<IpAddr>>> {
+        let protocol = self.protocol.clone();
+        let target_urn = urn.clone();
         let span = tracing::trace_span!("Peer::providers");
 
-        let protocol = self.protocol.clone();
-        let urn2 = urn.clone();
         async move {
-            let events = protocol.subscribe().await.filter_map(move |evt| {
+            // Listen for `Has` gossip events where a provider
+            // claims to have the target urn.
+            let providers_stream = protocol.subscribe().await.filter_map(move |evt| {
                 future::ready(match evt {
-                    ProtocolEvent::Gossip(gossip::Info::Has(gossip::Has { provider, val })) => {
-                        if val.urn == urn {
-                            println!("Found it for urn {}", urn);
-                            Some(provider)
-                        } else {
-                            println!("val.urn ({}) differs from urn ({})", val.urn, urn);
-                            None
-                        }
+                    ProtocolEvent::Gossip(gossip::Info::Has(gossip::Has { provider, val }))
+                        if val.urn == urn =>
+                    {
+                        Some(provider)
                     },
-                    x => {
-                        println!("Other case for {}: {:?}", urn, x);
-                        None
-                    },
+                    _ => None,
                 })
             });
+            // Query the network for providers for the given target urn.
             protocol
                 .query(Gossip {
-                    urn: urn2,
+                    urn: target_urn,
                     rev: None,
                     origin: None,
                 })
                 .instrument(span)
                 .await;
 
-            events
+            providers_stream
         }
     }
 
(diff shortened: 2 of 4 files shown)


librad/src/net/peer.rs
```diff
@@ -199,12 +199,19 @@ where
...
+                            println!("Found it for urn {}", urn);
+                            Some(provider)
+                        } else {
+                            println!("val.urn ({}) differs from urn ({})", val.urn, urn);
+                            None
+                        }
+                    },
+                    x => {
+                        println!("Other case for {}: {:?}", urn, x);
+                        None
```
NunoAlexandre, line 213: These changes are purely for temporary debugging purposes

FintanH: fyi: tracing might be handier since it has the span information from the caller of this function.

librad/src/net/gossip.rs
```diff
@@ -822,6 +822,7 @@ where
         let excluding = excluding.into();
 
         let mut connected_peers = self.connected_peers.lock().await;
+
```
xla, line 825:
```suggestion
```

librad/src/net/peer.rs
```diff
@@ -195,9 +196,13 @@ where
         urn: RadUrn,
     ) -> impl Future<Output = impl futures::Stream<Item = PeerInfo<IpAddr>>> {
         let protocol = self.protocol.clone();
-        let urn2 = urn.clone();
+        let target_urn = urn.clone();
+        let span = tracing::trace_span!("Peer::providers");
```
xla, line 200:
```suggestion
        let span = tracing::trace_span!("Peer::providers", urn = %urn);
```

Would also put it at the start of the method.

librad/tests/propagation_basic.rs
```diff
@@ -331,3 +331,64 @@ async fn all_metadata_returns_only_local_projects() {
...
+        radicle
+            .sign(
+                &peer1_key,
+                &Signatory::User(alice.urn()),
+                &resolves_to_alice,
+            )
+            .unwrap();
+
+        let peer1_id = peer1.peer_id().clone();
+        let repo_urn = tokio::task::spawn_blocking(move || {
```
FintanH, line 364: Good find :+1: This blocking IO stuff is quite the pain

FintanH (approved):
Nice work both of you :bow:

librad/tests/propagation_basic.rs
NunoAlexandre: kudus to the one and only, @xla

xla (approved):
🈲 🚝 🏆 💦

librad/tests/propagation_basic.rs
kim: Why is this necessary? `create_repo` is completely sync.

librad/tests/propagation_basic.rs
```diff
@@ -331,3 +331,64 @@ async fn all_metadata_returns_only_local_projects() {
...
+        let res = timeout(
+            Duration::from_secs(5),
+            peer2.providers(repo_urn).await.next(),
+        )
+        .await;
+
+        match res {
+            Ok(Some(peer_info)) => assert_eq!(
+                peer_info.peer_id, peer1_id,
+                "Expected it to be {} but got {} instead",
```
kim, line 382:
```suggestion
                "Expected peer id {} but got {} instead",
```

librad/src/net/peer.rs
```diff
@@ -194,10 +195,14 @@ where
...
+        let span = tracing::trace_span!("PeerApi::providers", urn = %urn);
         let protocol = self.protocol.clone();
-        let urn2 = urn.clone();
+        let target_urn = urn.clone();
+
         async move {
-            let events = protocol.subscribe().await.filter_map(move |evt| {
+            // Listen for `Has` gossip events where a provider
+            // claims to have the target urn.
+            let providers_stream = protocol.subscribe().await.filter_map(move |evt| {
```
kim, line 205: I'd prefer not to introduce hungarian notation and program flow comments. If it's unclear what the type of an expression is, or what it does, refactor it so you can add a type annotation and rustdoc commentary.

kim (changes requested):

librad/src/net/peer.rs
NunoAlexandre: I'm fine with dropping the comment, triggered by having `providers` doing two things.

Regarding the "Hungarian notation", I believe you are referring to the change from `let events` to `let providers_stream`? If so, I don't get your point, as "events" provides no clarity or meaning whereas "providers_stream" names it after exactly what it is.

librad/tests/propagation_basic.rs
NunoAlexandre: I think the issue is that `create_repo` _is_ sync, i.e, blocking. Looking at the docs of `tokio::task::spawn_blocking`: 

> In general, issuing a blocking call or performing a lot of compute in a future without yielding is not okay, as it may prevent the executor from driving other futures forward. This function runs the provided closure on a thread dedicated to blocking operations.

Or did I misunderstood the issue, @xla?

FintanH: I was a bit sceptical too and tried running it without the spawn blocking. The call to `providers` fails.

Is there some asynchrony between the storage and receiving gossip requests maybe?

NunoAlexandre: The issue wasn't quite around receiving gossip requests, it was that we were querying the network for the given URN before the seed had connected.

kim: Sure, but you don't rely on other futures being polled while this block is executed. This is different from `fetch_repo`, f.ex., which is not _truly_ sync, as it has to drive the network stack's executor by means of going through the (sync) FFI. Note that it does not help to mark that method `async` -- if it is not spawned via a blocking-aware spawner, nothing will happen.

Async Rust is just lovely.

librad/src/net/peer.rs
kim: "Stream" is the type of the expression. `providers` would suffice as a name.

librad/tests/propagation_basic.rs
FintanH: > The issue wasn't quite around receiving gossip requests, it was that we were querying the network for the given URN before the seed had connected.

But the `testnet` setup code waits for the seeds to be connected to each other. So that seems weird to me.

What I meant by my gossip comment was that this block indicates to me that we needed to ensure that the project was created in `peer1` before `peer2` made the query asking for the URN. Indicating further that `peer1` had to ensure it had the project before it _received_ a gossip query to check it did indeed have the URN requested.

So my confusion as to why it's needed persists tbh...

NunoAlexandre: @FintanH thanks for the clarification, I understand now (also understand something @xla has said before).

> So my confusion as to why it's needed persists tbh...

indeed...

kim: Oooh, so what happens here is this:

The test harness waits until `n` peer connections have been established, but it doesn't wait for the gossip membership to settle. By spawning the create repo stuff, we accidentally give it enough time to do that. If we don't, asking the other end for something wins the race (but the other end is not "there" yet).

So tl;dr we need a way to make the test harness wait on `Join` completion, otherwise our tests only work accidentally.

FintanH: Are we able to inspect those types of `ProtocolEvent`s currently?

FintanH: Can confirm it's a matter of delaying after trying out the test by using `tokio::time::delay_for(Duration::from_secs(1)).await;` instead.

kim: #313

kim (approved):

FintanH (approved):

xla (approved):
🌰 🔈 🔆 🌎

PR #310, opened 2020-09-08
Identities: git implementation

kim:
`TODO(kim): moar testing`

### DIFF
--- a/librad/src/identities/git.rs
+++ b/librad/src/identities/git.rs
@@ -1,0 +1,760 @@
+// This file is part of radicle-link
+// <https://github.com/radicle-dev/radicle-link>
+//
+// Copyright (C) 2019-2020 The Radicle Team <dev@radicle.xyz>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 or
+// later as published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use std::{convert::TryFrom, marker::PhantomData};
+
+use either::*;
+use futures::executor::block_on;
+use multihash::Multihash;
+
+use crate::{
+    git::ext,
+    identities::{
+        delegation::{self, Delegations},
+        generic::{self, Signed, Verified},
+        payload::{self, ProjectPayload, UserPayload},
+        sign::{Signature, Signatures},
+        urn::Urn,
+    },
+    internal::canonical::Cjson,
+    keys::PublicKey,
+    signer::Signer,
+};
+
+pub mod error;
+pub mod iter;
+
+pub use generic::Verifying;
+
+mod load;
+mod sign;
+
+#[cfg(test)]
+pub(crate) mod tests;
+
+use iter::Iter;
+use load::ByOid;
+
+pub type Revision = ext::Oid;
+pub type ContentId = ext::Oid;
+
+pub type Doc<T, D> = generic::Doc<T, D, Revision>;
+pub type Identity<T> = generic::Identity<T, Revision, ContentId>;
+
+pub type SignedIdentity<T> = generic::Verifying<Identity<T>, Signed>;
+pub type VerifiedIdentity<T> = generic::Verifying<Identity<T>, Verified>;
+
+pub type UserDoc = Doc<UserPayload, delegation::Direct>;
+pub type ProjectDoc = Doc<ProjectPayload, IndirectDelegation>;
+
+pub type User = Identity<UserDoc>;
+pub type Project = Identity<ProjectDoc>;
+
+#[non_exhaustive]
+pub enum SomeIdentity {
+    User(User),
+    Project(Project),
+}
+
+pub type SignedUser = SignedIdentity<UserDoc>;
+pub type SignedProject = SignedIdentity<ProjectDoc>;
+
+pub type VerifiedUser = VerifiedIdentity<UserDoc>;
+pub type VerifiedProject = VerifiedIdentity<ProjectDoc>;
+
+pub type VerificationError = generic::error::Verify<Revision, ContentId>;
+
+pub type IndirectDelegation = delegation::Indirect<UserPayload, Revision, ContentId>;
+
+#[derive(Clone)]
+pub struct Git<'a, T> {
+    repo: &'a git2::Repository,
+    _marker: PhantomData<T>,
+}
+
+impl<'a, T: 'a> Git<'a, T> {
+    pub fn new(repo: &'a git2::Repository) -> Self {
+        Self {
+            repo,
+            _marker: PhantomData,
+        }
+    }
+
+    /// Convenience to specialise `T` to [`User`].
+    pub fn as_user(&self) -> Git<'_, User> {
+        Git {
+            repo: self.repo,
+            _marker: PhantomData,
+        }
+    }
+
+    /// Convenience to specialise `T` to [`Project`].
+    pub fn as_project(&self) -> Git<'_, Project> {
+        Git {
+            repo: self.repo,
+            _marker: PhantomData,
+        }
+    }
+
+    /// Read an identity whose type is not statically known from commit `oid`.
+    ///
+    /// The only guarantee about the returned value is that it is well-formed --
+    /// it may or may not pass verification.
+    pub fn some_identity(&self, oid: git2::Oid) -> Result<SomeIdentity, error::Load> {
+        SomeIdentity::try_from(self.by_oid(oid))
+    }
+
+    /// Traverse the history with head commit `head`, yielding identities of
+    /// type `T`.
+    ///
+    /// The iterator yields elements in **reverse order**, ie. oldest-first. No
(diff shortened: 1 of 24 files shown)


kim:
I believe this is fairly _safe_ now, although I wouldn't recommend conducting an actual multisig consensus in-band (but nobody promised that, amirite?).

librad/src/identities/git.rs
```diff
@@ -0,0 +1,763 @@
...
+        root.verify(progeny)
+    }
+
+    //// Helpers ////
+
+    fn by_oid(&self, oid: git2::Oid) -> ByOid<'a> {
+        (self.repo, oid)
+    }
+
+    fn is_in_ancestry_path(&self, commit: git2::Oid, tree: git2::Oid) -> Result<bool, git2::Error> {
```
FintanH, line 191: Is this different from [graph_descendant_of](https://docs.rs/git2/0.13.11/git2/struct.Repository.html#method.graph_descendant_of)?

librad/src/identities/git.rs
```diff
@@ -0,0 +1,763 @@
...
+            } else if ours.revision == theirs.revision {
+                Ok(Action::SlowFwd)
+            } else if Some(&ours.revision) == theirs.doc.replaces() {
+                Ok(Action::SuccRev)
+            } else {
+                Err(error::Merge::RevisionMismatch)
+            }
+        }?;
+
+        //println!("action: {:?}", action);
```
FintanH, line 316:
```suggestion
```

:eyes:

librad/src/identities/git.rs
```diff
@@ -0,0 +1,763 @@
...
+        let author = self.repo.signature()?;
+
+        self.repo
+            .commit(
+                None,
+                &author,
+                &author,
+                &sign::CommitMessage::new(message, signatures).to_string(),
+                &tree,
+                parents.iter().collect::<Vec<_>>().as_slice(),
```
FintanH, line 387: Since you collected above couldn't you just call `as_slice`? Or is it something to do with references? :roll_eyes: 
```suggestion
                parents.as_slice(),
```

librad/src/identities/git.rs
```diff
@@ -0,0 +1,763 @@
...
+    /// If both `payload` and `delegations` evaluate to `None`, or their values
+    /// result in the same revision as `base`, no new commit is made, and
+    /// the result is the unwrapped [`User`] of the `base` argument.
+    ///
+    /// Otherwise, the result is a new [`User`] whose parent is `base`.
+    pub fn update<S>(
+        &self,
+        base: SignedUser,
+        payload: impl Into<Option<UserPayload>>,
+        delegations: impl Into<Option<delegation::Direct>>,
```
FintanH, line 462: I find the option of not passing either of these a bit unergonomic. Unless there is some use case for passing neither and just getting back the `base`? :shrug: 

Would it make more sense to use something like `These` so that we have to update at least one if not both?

librad/src/identities/git.rs
```diff
@@ -0,0 +1,763 @@
...
+        }
+
+        let doc = Doc {
+            version: 0,
+            replaces: Some(base.revision),
+            payload: payload.unwrap_or_else(|| base.doc.payload.clone()),
+            delegations: delegations
+                .or_else(|| Some(base.doc.delegations.clone()))
+                .map(payload::UserDelegations::from)
+                .unwrap(),
```
FintanH, line 484: I'm a little confused by this chain. Why does this end up being safe to `unwrap`?

FintanH (commented):
Sick stuff :ok_hand: 

Lefts some inquiries, but otherwise I'm happy (as usual)

librad/src/identities/git.rs
kim: Yeah, it looks for the `tree` (= `Revision`), not the `commit` (= `ContentId`)

librad/src/identities/git.rs
kim: Oh ja, very annoying: it wants a `&[&Commit<'_>]`. Dunno how to make one without allocating two `Vec`s.

librad/src/identities/git.rs
kim: Hmya, but can we make an ergonomic `These` in rrrrrst? Builder pattern maybe? :stuck_out_tongue:

librad/src/identities/git.rs
kim: Because `or_else` makes it always `Some`. Can do a pattern match if you prefer.

librad/src/identities/git.rs
FintanH: Shameful plug: https://crates.io/crates/these, I gave it a go when I was learning Rust for my interview here :sweat_smile:

librad/src/identities/git.rs
FintanH: Fair enough :(

librad/src/identities/git.rs
FintanH: It's probably too rough around the edges anyway and this is certainly not a blocker

kim:
> And `UserDelegations::from` needs `Option<Delegations>`?

Haha, no! You have no idea how many times I need to re-lookup the docs for
`Option` and `Result`. Can't seem to remember the types of those `ok`,
`or_else`, `kk_thx`.

It's like this:

delegations // Option<delegation::Direct>
    // If `None`, conjure a `Some`
    .or_else(|| Some(base.doc.delegations.clone()))
    // So we have `Some` but we want it to be `Option<payload::UserDelegations>`
    .map(payload::UserDelegations::from)
    // Now this is `Some`, amirite?
    .unwrap(),

FintanH:
Ah! So maybe it should be:
```rust
payload::UserDelegations::from(delegations.unwrap_or_else(|| base.doc.delegations.clone()))
```

So we unwrap the Some, but if it's None then provide the default value, and finalising it with the `from` call.

FintanH:
Got the above compiling :point_up: Want me to commit?

librad/src/identities/git.rs
kim: It's also that we can't statically prevent that this doesn't end up with the same hash -- you could be passing non-`None` values, but they are the same as in `base`. I figured that this shouldn't be an error, but the method be idempotent -- you just get the same thing back, and proceed with your updates.

FintanH: Yup, that check made sense to me :+1: I was more just thinking of the ergonomics of at least attempting to update. No biggy anyway :)

FintanH (approved):
Veryyy niiiiice :+1:

xla (dismissed):
Very dope! The test cases turned out very crisp and easy to follow. :clap: 

🍍 😆 🖍 3️⃣

librad/src/identities/git.rs
```diff
@@ -0,0 +1,760 @@
...
+    where
+        Doc: Delegations + generic::Replaces<Revision = Revision>,
+        <Doc as Delegations>::Error: std::error::Error + Send + Sync + 'static,
+
+        Identity<Doc>: TryFrom<ByOid<'a>, Error = error::Load>,
+    {
+        let mut progeny = Iter::<'_, Identity<Doc>>::new(self.repo, head)
+            .map_err(generic::error::Verify::history)?;
+
+        // TODO(kim): should we skip non-quorum commits at the beginning?
```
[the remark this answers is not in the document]
kim: It is actually the case that this doesn't _have_ to start at the first commit. It would be perfectly possible to create the initial revision with any number > 1 of trust delegations, but sign that only with the key at hand (which means no quorum). There isn't a way, however, to collect the signatures until quorum is reached _in band_, because replication demands to find a verified revision (that's the whole point).

So, the only reasons to skip until we find a quorum would be to support some kind of oob initialisation -- for example, via your favourite hosting service -- or to propose a new identity within the context of an already-valid project (and later just transplanting that branch). Idk if it is preferable to adhere to Postel's Law here, or to encode the youre-supposed-to-hold-it-this-way path and fix things once we learn about what crazy things people are trying to do.

librad/src/identities/git.rs
```diff
@@ -0,0 +1,760 @@
...
+
+pub type VerifiedUser = VerifiedIdentity<UserDoc>;
+pub type VerifiedProject = VerifiedIdentity<ProjectDoc>;
+
+pub type VerificationError = generic::error::Verify<Revision, ContentId>;
+
+pub type IndirectDelegation = delegation::Indirect<UserPayload, Revision, ContentId>;
+
+#[derive(Clone)]
+pub struct Git<'a, T> {
```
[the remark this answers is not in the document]
kim: Not a fan. I like to think of these things as `forall 'a` -- if I later decide that some other reference needs to be bound to that lifetime, I'd have to change that name everywhere.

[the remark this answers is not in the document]
xla: Reasonable.

librad/src/identities/git.rs
```diff
@@ -0,0 +1,760 @@
...
+impl<'a> Git<'a, User> {
+    /// Attempt to read a [`User`] from commit `oid`, without verification.
+    pub fn get(&self, oid: git2::Oid) -> Result<User, error::Load> {
+        self.get_generic(oid)
+    }
+
+    /// Verify the user history with head commit `head`.
+    ///
+    /// The returned [`VerifiedUser`] is the **most recent** identity for which
+    /// the verification succeeded -- which may or may not be `head`.
```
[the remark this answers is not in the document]
kim: This is indeed a bit odd, but only of significance in tests. The only way this can happen is if there are unconfirmed proposals at the tip.

FintanH (approved):

xla (approved):
💹 🌳 🔃 👸

PR #312, opened 2020-09-09
Derive all Arbitrary impls by hand

kim:
Derive macros somehow interfere with typechecker error reporting, for
some reason even when the type being impl'd is not related at all.
`proptest_derive::Arbitrary` is the worst offender, although `serde`
derive macros leak as well.

Henceforth, macros shallt generally be frowned upon in this library, an
`proptest_derive` permanently banned.


FintanH (approved):
:+1:

PR #314, opened 2020-09-09
PeerApi: Require timeout for providers query

kim:
It does not make sense to poll the responses stream for more than a
couple of seconds.

### DIFF
--- a/librad/src/net/peer.rs
+++ b/librad/src/net/peer.rs
@@ -20,13 +20,15 @@ use std::{
     io,
     net::{IpAddr, SocketAddr},
     sync::{Arc, Mutex},
+    time::Duration,
 };
 
 use either::Either;
 use futures::{
-    future::{self, BoxFuture},
+    future::{self, BoxFuture, FutureExt},
     stream::StreamExt,
 };
+use futures_timer::Delay;
 use thiserror::Error;
 use tracing_futures::Instrument as _;
 
@@ -191,25 +193,47 @@ where
     /// [`RadUrn`] actually has it, nor that it is reachable using any of
     /// the addresses contained in [`PeerInfo`]. The implementation may
     /// change in the future to answer the query from a local cache first.
+    ///
+    /// The returned [`futures::Stream`] will be complete after the supplied
+    /// `timeout` has elapsed, whether or not any responses have been yielded
+    /// thus far. This is to prevent callers from polling the stream
+    /// indefinitely, even though no more responses can be expected. A realistic
+    /// timeout value is in the order of 10s of seconds.
     pub fn providers(
         &self,
         urn: RadUrn,
+        timeout: Duration,
     ) -> impl Future<Output = impl futures::Stream<Item = PeerInfo<IpAddr>>> {
         let span = tracing::trace_span!("PeerApi::providers", urn = %urn);
         let protocol = self.protocol.clone();
         let target_urn = urn.clone();
 
         async move {
-            let providers = protocol.subscribe().await.filter_map(move |evt| {
-                future::ready(match evt {
-                    ProtocolEvent::Gossip(gossip::Info::Has(gossip::Has { provider, val }))
-                        if val.urn == urn =>
-                    {
-                        Some(provider)
-                    },
-                    _ => None,
-                })
-            });
+            let providers = futures::stream::select(
+                futures::stream::once(
+                    async move {
+                        Delay::new(timeout).await;
+                        Err("timed out")
+                    }
+                    .boxed(),
+                ),
+                protocol
+                    .subscribe()
+                    .await
+                    .filter_map(move |evt| {
+                        future::ready(match evt {
+                            ProtocolEvent::Gossip(gossip::Info::Has(gossip::Has {
+                                provider,
+                                val,
+                            })) if val.urn == urn => Some(provider),
+                            _ => None,
+                        })
+                    })
+                    .map(Ok),
+            )
+            .take_while(|x| future::ready(x.is_ok()))
+            .map(Result::unwrap);
+
             protocol
                 .query(Gossip {
                     urn: target_urn,
--- a/librad/tests/propagation_basic.rs
+++ b/librad/tests/propagation_basic.rs
@@ -337,9 +337,6 @@ async fn all_metadata_returns_only_local_projects() {
 /// peer 1.
 #[tokio::test]
 async fn providers_works() {
-    use std::time::Duration;
-    use tokio::time::timeout;
-
     logging::init();
     const NUM_PEERS: usize = 2;
     let peers = testnet::setup(NUM_PEERS).await.unwrap();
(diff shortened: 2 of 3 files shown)


xla (dismissed):
Generally in favour of this change and putting sensible upper bounds to this kind of interface. What was the original use-case for the providers API? Was it driven by the seed node needs? As it currently shapes up don't think the proxy is gonna have much use for it.

🍀 🚨 🏄 🎦

librad/src/net/peer.rs
```diff
@@ -191,25 +193,47 @@ where
...
+                                provider,
+                                val,
+                            })) if val.urn == urn => Some(provider),
+                            _ => None,
+                        })
+                    })
+                    .map(Ok),
+            )
+            .take_while(|x| future::ready(x.is_ok()))
+            .map(Result::unwrap);
```
FintanH, line 235: Should callers not be able to handle the timeout error instead of receiving a panic?

kim: It does not panic.

kim:
> What was the original use-case for the providers API?

Whenever you want to get at something for which you only know the URN.

librad/src/net/peer.rs
FintanH: Ah... the `take_while` above :facepalm:

FintanH (dismissed):

NunoAlexandre:
I would have appreciated being asked to review this since it's touching what I am working on.

As I said in https://github.com/radicle-dev/radicle-upstream/pull/875#issuecomment-689409350, I'd see a benefit in having this being treated as a caller concern and see this fn as a low-level facility. Now, there is a single timeout, whereas maybe we will meet a use case where throttling the timeout for each attempt would be useful.

Moving on.

kim:
> maybe we will meet a use case where throttling the timeout for each attempt would be useful

Yeah, and maybe you will realise this only after you have shot yourself in the foot by refusing to put a timeout there at all. You can still add a per-element timeout if you want.

NunoAlexandre:
> > maybe we will meet a use case where throttling the timeout for each attempt would be useful
> 
> Yeah, and maybe you will realise this only after you have shot yourself in the foot by refusing to put a timeout there at all. You can still add a per-element timeout if you want.

How exactly would we shot ourselves in the foot now that we all know we need to timeout and now that this was actually tested using timeouts?

I'm fine with having this in, it's just not happy to have that linked discussion dismissed and have this done on the side without calling me in.

kim:
You're replacing reasoning with ritual, Nuno. The returned stream is
not infinite -- if it yields responses after a few seconds, then that
is coincidental. If you want that, `subscribe` to all events and keep
polling indefinitely. It was a mistake in the original API to not
impose that timeout, and I've corrected that. Depending on how I see
this used, I might consider to make the timeout a protocol parameter,
or a constant altogether.

NunoAlexandre:
I'd genuinely appreciate if you help me understand your points to ease collaboration in the future:

> You're replacing reasoning with ritual, Nuno. 

How? Not a defensive question, I am just not getting your point and I'd happy to see it.

> if it yields responses after a few seconds, then that
is coincidental.

Is that so under the understanding that seeds can only take so long to reply to the broadcasted `Want` plus added latency?
If that's it, that makes sense!

>  It was a mistake in the original API to not impose that timeout, and I've corrected that.

Yeah, with the previous point in mind this makes total sense. I was under the assumption that a seed could take an indefinite amount of time to reply: for example, say it was down for a while and would then pick up received events (as I have set up in an actor model in the past where we'd process those events from a mailbox). I see that's not the case here though.

> I might consider to make the timeout a protocol parameter,
or a constant altogether.

Given that the timeout would be scoped around peer operations and have the points above confirmed, that'd make sense. 

Again - as I said already - I didn't (and now don't at all) oppose these changes, but if I had been involved in this PR as I believe I should this would have been a well-rounded constructive discussion instead of a post-mortem one. Do we agree on that? That matters to build a collaborating environment.

kim:
> How?

I explained the pitfalls of this method on the other PRs. What I get
back tells me that the semantics are not understood, and the
discussion derails into questions of formatting documentation. The
only possible conclusion I can draw from this is that I misdesigned
the original API, and I don't see why it requires a democratic process
to rectify that.

> Is that so under the understanding that seeds can only take so long to reply to the broadcasted Want plus added latency?

It's not really about seeds -- you might not even be connected to one
directly. What happens is that we just broadcast a message, which
travels around the network until it hits a node which is able and
willing to respond to it, and then that response has to find its way
back. Or it never reaches such a node. Or it reaches multiple such
nodes. If we get a response back, it could be in response to our
message, or it could be that someone else asked for the same message,
and we're coincidentally on the path the response took through the
network. In any case, after our message has propagated through the
network, we can't expect any more responses (except those which are
actually answering someone else). Since we don't know when that will
be the case, we have to resort to a timer.

The `providers` method **pretends** that this whole procedure is like
a normal request / response flow, just because it's more convenient to
set up this dance of "stick message in one slot, and wait on other
slot until something comes out which looks about right" once. It does
not have these familiar semantics, though. In fact, as the
documentation says, the implementation is supposed to change at some
point to not even hit the network in most cases.

> say it was down for a while and would then pick up received events

That is unfortunately the crux of many applications on top of
peer-to-peer networks: there is no place to store-and-forward things.

NunoAlexandre:
>> How?
> I explained the pitfalls of this method on the other PRs. What I get back tells me that the semantics are not understood, and the discussion derails into questions of formatting documentation. The only possible conclusion I can draw from this is that I misdesigned the original API, and I don't see why it requires a democratic process to rectify that.

We are arguing about different things, Kim. 

I am not arguing this change required a democratic vote. My point is that since this changeset touches a piece I had just touched here and that I was actively touching in the `upstream`, _and_ since I had tried to discuss this exact point with little avail, it would only be sensible to involve me.

Thus, in the future please involve me in PRs and discussions involving pieces I have or am actively working on.

> > Is that so under the understanding that seeds can only take so long to reply to the broadcasted Want plus added latency?
> It's not really about seeds -- you might not even be connected to one directly. What happens is that we just broadcast a message, which travels around the network until it hits a node which is able and willing to respond to it, and then that response has to find its way back. Or it never reaches such a node. Or it reaches multiple such nodes. If we get a response back, it could be in response to our message, or it could be that someone else asked for the same message, and we're coincidentally on the path the response took through the network. In any case, after our message has propagated through the network, we can't expect any more responses (except those which are actually answering someone else). Since we don't know when that will be the case, we have to resort to a timer. The `providers` method **pretends** that this whole procedure is like a normal request / response flow, just because it's more convenient to set up this dance of "stick message in one slot, and wait on other slot until something comes out which looks about right" once. It does not have these familiar semantics, though. In fact, as the documentation says, the implementation is supposed to change at some point to not even hit the network in most cases.

This is super insightful and in my view should be added to the docs of `providers` or be documented in a more general place if better suited.

kim:
> I am not arguing this change required a democratic vote. [...] it would only be sensible to involve me.

In what way then? I made you aware of this patch in the context of your work.

> This is super insightful and in my view should be added to the docs of providers or be documented in a more general place if better suited.

I think it is fair to expect familiarity with peer-to-peer networking
from users of this library. Higher-level documentation which explains
more fundamental concepts would certainly be appreciated, but perhaps
it makes more sense to tackle this once the protocol has stabilised.

NunoAlexandre:
> > I am not arguing this change required a democratic vote. [...] it would only be sensible to involve me.
> > In what way then? I made you aware of this patch in the context of your work.

As a reviewer. I only found out about this PR when I was double-checking all the closed discussions on the other PR to make sure I had not missed anything (that is, you posted it there but then the discussion was closed. It's super easy to miss such an announcement in the midst of a PR with so many discussions.

> > This is super insightful and in my view should be added to the docs of providers or be documented in a more general place if better suited.
> I think it is fair to expect familiarity with peer-to-peer networking from users of this library. Higher-level documentation which explains more fundamental concepts would certainly be appreciated, but perhaps it makes more sense to tackle this once the protocol has stabilised.

That's fair :+1:

PR #319, opened 2020-09-14
Switch crypto to ed25519-zebra

### LINKED ISSUE #315: Use ed25519-zebra instead of libsodium
Reasons:

1. `libsodium` makes for slower builds (see also radicle-dev/radicle-keystore#7)
2. [`ed25519-zebra`](https://crates.io/crates/ed25519-zebra) has a batch verification API, which could potentially give a nice speedup for some things
3. The motivation for creating `ed25519-zebra` was that the ZCash project realised that a bugfix(!) in the way `libsodium` validates signatures could break their consensus. While not much of a concern for us immediately, their solution is attractive for two reasons:

    i. There is an unambiguous spec of how signature validation is supposed to work in `radicle-link` (ZIP 215)
    ii. It is, in essence, _more compatible_ with all kinds of ed25519 signatures, so we should never need to worry about upgrades

cloudhead:
Closes #315 

~~Everything compiles with patched `ed25519-zebra`, however tests fail. I'm pretty sure the serialization changes broke it.~~

### DIFF
--- a/librad/Cargo.toml
+++ b/librad/Cargo.toml
@@ -28,13 +28,13 @@ rcgen = "0.8"
 regex = "1.3"
 serde_bytes = "0.11"
 serde_json = "1.0"
-sodiumoxide = "0.2"
 thiserror = "1.0"
 tracing = "0.1"
 tracing-futures = "0.2"
 unicode-normalization = "0.1"
 urltemplate = "0.1"
 webpki = "0.21"
+ed25519-zebra = { version = "2.1.2", path = "../../ed25519-zebra" }
 
 [dependencies.either]
 version = ">= 1.3, 1"
@@ -104,6 +104,7 @@ futures-await-test = "0"
 futures_ringbuf = "0"
 pretty_assertions = "0"
 proptest = "0"
+rand_chacha = "0.2"
 tempfile = "3.1"
 
 [dev-dependencies.librad-test]
--- a/librad/Cargo.toml
+++ b/librad/Cargo.toml
@@ -104,7 +104,6 @@ futures-await-test = "0"
 futures_ringbuf = "0"
 pretty_assertions = "0"
 proptest = "0"
-rand_chacha = "0.2"
 tempfile = "3.1"
 
 [dev-dependencies.librad-test]
--- a/librad/Cargo.toml
+++ b/librad/Cargo.toml
@@ -34,7 +34,11 @@ tracing-futures = "0.2"
 unicode-normalization = "0.1"
 urltemplate = "0.1"
 webpki = "0.21"
-ed25519-zebra = { version = "2.1.2", path = "../../ed25519-zebra" }
+
+[dependencies.ed25519-zebra]
+version = "2.1.2"
+branch = "main"
+git = "https://github.com/radicle-dev/ed25519-zebra"
 
 [dependencies.either]
 version = ">= 1.3, 1"
--- a/librad/src/keys.rs
+++ b/librad/src/keys.rs
@@ -170,7 +170,7 @@ impl sign::Signer for SecretKey {
     type Error = Infallible;
 
     fn public_key(&self) -> sign::PublicKey {
-        sign::PublicKey(self.0.into())
+        sign::PublicKey(ed25519::VerificationKey::from(&self.0).into())
     }
 
     async fn sign(&self, data: &[u8]) -> Result<sign::Signature, Self::Error> {
--- a/librad/src/keys.rs
+++ b/librad/src/keys.rs
@@ -300,7 +300,7 @@ impl<'de> Deserialize<'de> for PublicKey {
 
                         ed25519::VerificationKeyBytes::try_from(data)
                             .map(PublicKey)
-                            .map_err(|err| serde::de::Error::custom(err))
+                            .map_err(serde::de::Error::custom)
                     },
                 }
             }
@@ -429,7 +429,7 @@ impl<'de> Deserialize<'de> for Signature {
 
                         ed25519::Signature::try_from(data)
                             .map(Signature)
-                            .map_err(|err| serde::de::Error::custom(err))
+                            .map_err(serde::de::Error::custom)
                     },
                 }
             }
--- a/librad/src/keys.rs
+++ b/librad/src/keys.rs
@@ -84,11 +84,7 @@ impl SecretKey {
 
     #[cfg(test)]
     pub fn from_seed(seed: [u8; 32]) -> Self {
-        use rand::SeedableRng;
-
-        let rng = rand_chacha::ChaCha8Rng::from_seed(seed);
-        let sk = ed25519::SigningKey::new(rng);
-        Self(sk)
+        Self(ed25519::SigningKey::from(seed))
     }
 
     pub(crate) fn from_secret(sk: ed25519::SigningKey) -> Self {
--- a/librad/src/keys.rs
+++ b/librad/src/keys.rs
@@ -476,7 +476,7 @@ pub mod tests {
     const DATA_TO_SIGN: &[u8] = b"alors monsieur";
 
     pub fn gen_secret_key() -> impl Strategy<Value = SecretKey> {
-        any::<[u8; 32]>().prop_map(|seed| SecretKey::from_seed(seed))
+        any::<[u8; 32]>().prop_map(SecretKey::from_seed)
     }
 
     pub fn gen_public_key() -> impl Strategy<Value = PublicKey> {
--- a/librad/src/keys.rs
+++ b/librad/src/keys.rs
@@ -15,17 +15,23 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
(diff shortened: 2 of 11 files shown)


librad/src/keys.rs
```diff
@@ -59,27 +49,27 @@ pub trait SignError: error::Error + Send + Sync + 'static {}
...
 impl From<sign::PublicKey> for PublicKey {
     fn from(other: sign::PublicKey) -> PublicKey {
-        PublicKey(ed25519::PublicKey(other.0))
+        PublicKey(ed25519::VerificationKeyBytes::from(other.0))
     }
 }
 
 /// A signature produced by `Key::sign`
-#[derive(Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
+#[derive(Clone, Debug, Eq, PartialEq)]
```
cloudhead, line 67: We don't use these instances in this crate, and they aren't available on ed25519-zebra. If we use them elsewhere, we'd have to patch upstream.

librad/src/keys.rs
```diff
@@ -339,35 +333,35 @@ impl<'b> minicbor::Decode<'b> for PublicKey {
...
+            .ok()
     }
 
     pub fn to_bs58(&self) -> String {
-        bs58::encode(self.0)
+        let bytes: [u8; 64] = self.0.into();
+        bs58::encode(&bytes[..])
             .with_alphabet(bs58::alphabet::BITCOIN)
             .into_string()
     }
```
cloudhead, line 367: This is ugly because zebra's `Signature` doesn't have an `as_ref() -> &[u8]` due to the inner data being split in two. There's an issue for it upstream.

librad/src/keys.rs
```diff
@@ -379,9 +373,9 @@ impl fmt::Display for Signature {
...
 }
 
-impl AsRef<[u8]> for Signature {
-    fn as_ref(&self) -> &[u8] {
-        self.0.as_ref()
+impl From<Signature> for [u8; 64] {
+    fn from(sig: Signature) -> [u8; 64] {
+        sig.0.into()
     }
 }
```
cloudhead, line 380: Can't have `AsRef<[u8]>` on `Signature` for now, as explained above.

librad/src/keys.rs
xla: We could leave an inline TODO with a link to the upstream issue.

cloudhead:
Not quite sure anymore if the errors are due to serialization bugs. Here's an example failing test: 

```
---- git::storage::config::tests::test_init stdout ----
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `PeerId(PublicKey(VerificationKeyBytes("94b3dc72749d4905649973b2e6783a037eb6c9c6e5c0b773f8529d6b32f0711d")))`,
 right: `PeerId(PublicKey(VerificationKeyBytes("9ca65dafd8e81e98b2cae25a95163df24b70f9d311b59903ee6dbd5ff9313eaf")))`', librad/src/git/storage/config.rs:238:9
```

librad/src/keys.rs
```diff
@@ -131,12 +123,13 @@ impl SecretKey {
...
                 });
-                let seed = yasna::construct_der(|writer| writer.write_bytes(&self.0[..32]));
+                let seed =
+                    yasna::construct_der(|writer| writer.write_bytes(&self.0.as_ref()[..32]));
                 writer.next().write_bytes(&seed);
                 writer
                     .next()
                     .write_tagged(yasna::Tag::context(1), |writer| {
-                        let public = &self.0[32..];
+                        let public = &self.0.as_ref()[32..];
```
kim, line 132: Does this not crash? The slice returned by `AsRef<[u8]>` is only 32 bytes. 

Iirc, the first thing written needs to be the secret part, and the second the public, which is commonly stored in one 64 bytes array. But not so with the zebra!

librad/src/keys.rs
kim: This method can be considered deprecated, so nothing to worry about.

librad/Cargo.toml
```diff
@@ -28,13 +28,13 @@ rcgen = "0.8"
...
 serde_bytes = "0.11"
 serde_json = "1.0"
-sodiumoxide = "0.2"
 thiserror = "1.0"
 tracing = "0.1"
 tracing-futures = "0.2"
 unicode-normalization = "0.1"
 urltemplate = "0.1"
 webpki = "0.21"
+ed25519-zebra = { version = "2.1.2", path = "../../ed25519-zebra" }
```
kim, line 37: `ls -la ../../../../etc/shadow` aah interesting

kim (commented):
> Not quite sure anymore if the errors are due to obvious serialization bugs.

I suspect the conversion from raw `Signer` bytes might not be kosher.

librad/Cargo.toml
xla: Local first.

kim: Now I regret not having used `cat`.

librad/src/keys.rs
```diff
@@ -88,29 +78,31 @@ impl From<sign::Signature> for Signature {
...
+        let sk = ed25519::SigningKey::new(rand::thread_rng());
         Self(sk)
     }
 
     #[cfg(test)]
-    pub fn from_seed(seed: &ed25519::Seed) -> Self {
-        ensure_initialised();
-        let (_, sk) = ed25519::keypair_from_seed(seed);
+    pub fn from_seed(seed: [u8; 32]) -> Self {
+        use rand::SeedableRng;
```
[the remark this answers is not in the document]
cloudhead: Ahhh, the method name and usage confused me, yeah it should be really simple actually.

librad/Cargo.toml
cloudhead: :joy:

librad/src/keys.rs
cloudhead: Oh wow, it doesn't crash in the expected way, but the test was failing. Should be fixed in 650ba5f

librad/src/keys.rs
```diff
@@ -59,27 +49,27 @@ pub trait SignError: error::Error + Send + Sync + 'static {}
...
 
 /// The public part of a `Key``
 #[derive(Clone, Debug, Eq, PartialEq, PartialOrd, Ord, Hash)]
-pub struct PublicKey(ed25519::PublicKey);
+pub struct PublicKey(ed25519::VerificationKeyBytes);
 
 impl From<sign::PublicKey> for PublicKey {
     fn from(other: sign::PublicKey) -> PublicKey {
-        PublicKey(ed25519::PublicKey(other.0))
+        PublicKey(ed25519::VerificationKeyBytes::from(other.0))
```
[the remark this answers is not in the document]
cloudhead: Should we add that test to radicle-keystore, since there are other similar tests there?

cloudhead:
Ok, this is ready for final review.

* For now I am pointing to our fork of ed25519-zebra, but hopefully our patch gets merged and we can use upstream.
* To benefit from the lower dependency burden of ed25519-zebra, we should also migrate radicle-keystore to it, since librad still depends on sodiumoxide through that.

librad/src/keys.rs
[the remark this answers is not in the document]
kim: Would be great, yes. I'm wondering if this is enough, though.

librad/src/keys.rs
```diff
@@ -59,27 +49,27 @@ pub trait SignError: error::Error + Send + Sync + 'static {}
 impl<T: error::Error + Send + Sync + 'static> SignError for T {}
 
 /// A device-specific signing key
-#[derive(Clone, Eq, PartialEq)]
+#[derive(Copy, Clone)]
```
kim, line 52: Wondering (for a long time, actually) if we should make `PublicKey` `Copy`, too. Would probably be a clippy-fest.

kim (approved):
Great, thanks!

> To benefit from the lower dependency burden of ed25519-zebra, we should also migrate radicle-keystore to it, since librad still depends on sodiumoxide through that (I'm happy to take that on).

That'd be great. Note radicle-dev/radicle-keystore#7, tho

xla (approved):
utACK

🎧 📕 🗞 😷

librad/src/keys.rs
cloudhead: Hahaha, yes, since it's only 32 bytes it's fairly cheap to copy.

cloudhead:
> Great, thanks!
> 
> > To benefit from the lower dependency burden of ed25519-zebra, we should also migrate radicle-keystore to it, since librad still depends on sodiumoxide through that (I'm happy to take that on).
> 
> That'd be great. Note [radicle-dev/radicle-keystore#7](https://github.com/radicle-dev/radicle-keystore/issues/7), tho

Aha yes, we need the cryptobox thing :thinking: I'll have a look.

librad/src/keys.rs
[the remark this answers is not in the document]
cloudhead: Gonna start with that and we can figure out how to proceed.. Maybe librad can sort of "call into" keystore's tests or something..

Issue #307, opened 2020-09-04
Revisit Cargo.lock problem


kim:
As per the [official recommendations](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries), binaries should have a `Cargo.lock` in version control, but libraries shouldn't. Since we don't have any binaries anymore, we should therefore not have a `Cargo.lock` in version control.

So far so good, but the issue is that, if the `master` build accesses the network, and pulls in some updated crate which causes the build to fail, that not only sucks, but we can also not reproduce which transitive closure was good, and which wasn't. It is also easy to bypass the "PR must be up-to-date with master" check of GH, so concurrent merges or ones in quick succession would make this worse.

A few possibilities:

* Publish the generated `Cargo.lock` as a build artifact, so we can at least cope with it post-mortem
* Keep `Cargo.lock` under version control, but run two parallel builds on `master`, of which one just removes it and "soft fails" the build worst case
* Set up a merge bot like `bors`/`homu` which sequentialises all `master` merges/builds
* Install a BDFL on a server, who does the same thing
* Run a sophisticated consensus algorithm over IRC

PR #325, opened 2020-09-17
Lockless cargo

kim:
Tries out approach 1 of #307 to see what happens.

Findings so far:

* [SemVer is a lie](https://blog.illicitonion.com/rust-minimum-versions-semver-is-a-lie/) (yeah, I tried `-Z minimal-versions`)
* It is not clear when cargo decides to pick a different version for a workspace crate (even with relaxed constraints)


kim:
This should win me an award:

image

FintanH:
I wanna see those 19 lines gone too

kim:
Alright, I'd like to see us roll lockless. We have the `radicle-bins` repo, which is supposedly going to have a `Cargo.lock`, and this branch already caught a diverging version constraint in the sub-crates.

I also think that we should employ a more lenient versioning policy (ie. no upper bounds where possible), but I'll leave that for another time.

xla (dismissed):
👒 ☠ 🔚 🎆

PR #326, opened 2020-09-17
Storage Tests

FintanH:
Added tests to confirm the setting of default rad/self and checking if a URN exists when calling `create_repo` twice. Turns out there was a bug in the latter so I'm including the fix.


xla (approved):
📺 🍿 📁 🗣

kim (approved):

PR #327, opened 2020-09-18
Pooled Storage, Upgrade Keystore, Network Lifecycle

kim:


kim:
Can't say this is an improvement yet. Now you have to make sure to drop the
storage borrowed from the pool, and also move values all the time due to
spawn_blocking.

librad/tests/propagation_basic.rs
```diff
@@ -335,7 +359,7 @@ async fn all_metadata_returns_only_local_projects() {
 /// Given that a) a peer 1 holds a given URN and b) that same peer is a seed of
 /// a peer 2, verify that requesting peer 2 for providers for said URN returns
 /// peer 1.
-#[tokio::test]
+#[tokio::test(core_threads = 4)]
```
kim, line 362: This one's now flaky :man_shrugging: :fried_shrimp:

xla:
Would it help if we move the `spawn_blocking` further down and employ something that looks similar on the Api? Like this:

``` rust
api.storage(move |storage| storage.clone_repo()).await?
```

It would not help with the moving, but we have to move it move it anyway. But it could beneficial as in that the caller doesn't need to remember to call `spawn_blocking`. Not sure but guess the drop would also be ensured with that.

kim:
> ``` rust
> api.storage(move |storage| storage.clone_repo()).await?
> ```

Yeah that's a good idea. It would preclude using `block_in_place` I think, but
since that contains a footgun, perhaps that'd be for the better.

librad/src/git/storage.rs
```diff
@@ -456,12 +452,7 @@ impl<S: Clone> Storage<S> {
 }
 
 impl Storage<NoSigner> {
-    /// Open the `Storage` found at the given [`Path`]'s `git_dir.
-    ///
-    /// The `Storage` must have been initialised with [`Storage::init`] prior to
-    /// calling this method.
-    pub fn open(paths: &Paths) -> Result<Self, Error> {
-        let backend = git2::Repository::open_bare(paths.git_dir())?;
+    pub fn open(backend: git2::Repository) -> Result<Self, Error> {
```
FintanH, line 455: One worry I'm having going through this is that the `Repository` is assumed to be opened bare right? I wonder if we should have a newtype around `git2::Repository` and a method `from_paths` which is the smart constructor for constructing that newtype.

librad/src/git/storage/pool.rs
```diff
@@ -0,0 +1,58 @@
...
+
+use deadpool::managed::{Manager, RecycleResult};
+
+use super::{Error, Storage};
+
+pub type Pool<S> = deadpool::managed::Pool<Storage<S>, Error>;
+pub type Pooled<S> = deadpool::managed::Object<Storage<S>, Error>;
+
+#[derive(Clone)]
+pub struct VPoS<S> {
```
FintanH, line 30: What is this acronym? :sweat_smile:

FintanH (commented):
This is looking reasonable to me :+1: Concurrency Without Fear:tm:

librad/src/git/storage/pool.rs
xla: Very Proof of Steak.

kim: Hint: it’s a MANAGER

xla: VP of Storage!!!

FintanH: :joy: ah, so it did go over my head...

kim:
@FintanH @xla Ready for review now. Please see messages of rebased commits.

kim:
Note that handing out that run loop future is now somewhat pointless, we could also just `spawn` it interiorly. I left it in for now, as we might come up with error conditions to be propagated up, or a way to restart just the network stack.

librad/src/net/peer.rs
```diff
@@ -380,40 +410,51 @@ where
...
+        block_in_place(|| match head.into() {
             None => git.has_urn(&urn).unwrap_or(false),
             Some(head) => git.has_commit(&urn, head).unwrap_or(false),
-        }
+        })
+    }
+
+    async fn is_tracked(&self, urn: &RadUrn, peer: &PeerId) -> Result<bool, PeerStorageError> {
+        let git = self.inner.get().await?;
+        Ok(block_in_place(|| git.is_tracked(urn, peer))?)
```
xla, line 457: When we call `block_in_place` here, does that mean we force the threaded runtime on consumers as alluded to in a FIXME further up?

librad/src/net/protocol.rs
```diff
@@ -116,6 +121,53 @@ pub struct Protocol<S, A> {
...
+            git: self.git.clone(),
+            endpoint: self.endpoint.clone(),
+            connections: self.connections.clone(),
+            subscribers: self.subscribers.clone(),
+            ref_count: self.ref_count.clone(),
+        }
+    }
+}
+
+impl<S, A> Drop for Protocol<S, A> {
```
xla, line 154: What happens to the run_loop when the protocol is dropped?

librad/src/net/protocol.rs
```diff
@@ -152,6 +205,27 @@ where
                 local.addr = %local_addr
             );
 
+            // Future which resolves once our refcount reaches zero.
+            //
+            // This ensures spawned tasks handling ingress/egress streams
+            // terminate when the last reference to `this` is dropped.
+            struct Bomb {
```
xla, line 212: :bomb: :boom:

librad/src/net/quic.rs
```diff
@@ -103,6 +103,19 @@ impl Endpoint {
...
     }
+
+    /// Close all of this endpoint's connections immediately and cease accepting
+    /// new connections.
+    ///
+    /// Should only ever by called when the protocol stack is shut down.
+    ///
+    /// Morally, this method should consume `self`, but we shut down from a
+    /// `Drop` impl, so all we got is `&mut self`.
+    pub(crate) fn shutdown(&mut self) {
```
xla, line 114: Do you want to exclusively enforce drop for protoocl shutdown semantics? Or is there a possibility to have this be part of the public interface. Found myself in the situation where in tests I want to spin up a peer have do some operations and then shut it down.

librad/tests/working_copy.rs
```diff
@@ -59,7 +58,7 @@ use librad_test::{
 /// 7. peer2 creates an include file, based of the tracked users of the project
 /// i.e. peer1 8. peer2 includes this file in their working copy's config
 /// 9. peer2 fetches in the working copy and sees the commit
-#[tokio::test]
+#[tokio::test(core_threads = 2)]
```
xla, line 61: Do we know why we have to increase the thread count here?

xla (dismissed):
Extra dope! Best thing is the introduction of a pool for the protocol as well to address potentially halting progress from long-running fetchs. :clap: 

📂 🚵 🚦 🔪

librad/src/net/protocol.rs
kim: It keeps running -- the full shutdown only takes place when both are dropped (which incidentally works in `upstream`). As noted, the run loop does not make much sense anymore, but I want to think about whether we want it to return an error in some cases, or if we can redesign it somehow so we could just restart the network with some altered parameters.

librad/src/net/quic.rs
kim: There is no other way -- the `quinn::Endpoint` only releases all resources when dropped. So what we could have is to _suspend_ the network, but I'm not sure that makes a lot of sense.

librad/src/net/protocol.rs
xla: Agreed, we should marinate this a bit more before we commit to anything. The only thing worth considering at this point is if the `AbortHandle` that is used in a couple of places now (`testnet` and the upstream proxy) should be the returned by librad.

librad/src/net/quic.rs
xla: Sound. Guess as long as the documentation of the `Protocol` explicitely states the drop semantics it should suffice.

librad/tests/working_copy.rs
kim: Actually, I don't know. It didn't seem to work on the single-threaded runtime at some point, irrespective of the `block_in_place`, but perhaps that should receive a bit more analysis.

librad/src/net/protocol.rs
kim: It's not really possible nor meaningful to hand out an `AbortHandle`, or stick some channel receiver into the protocol, _unless_ the protocol would _internally_ drop all the resources (and runaway spawned tasks). But then, what do you do with such an object? You'd need to query it at runtime to see if it is actually running or not.

I'd say async Rust is just inherently anti-modular at its current stage, and RAII the predominant idiom. Up to the library author to make sure `drop` does what it says it does.

xla: Understood!

librad/tests/propagation_basic.rs
```diff
@@ -38,7 +39,7 @@ use librad_test::{
     },
 };
 
-#[tokio::test]
+#[tokio::test(core_threads = 2)]
```
FintanH, line 42: This worked on my machine without this

librad/tests/propagation_basic.rs
```diff
@@ -65,27 +66,33 @@ async fn can_clone() {
...
+                storage.open_repo(radicle_urn).unwrap();
+            })
+            .await
+            .unwrap();
     })
     .await;
 }
 
-#[tokio::test]
+#[tokio::test(core_threads = 2)]
```
FintanH, line 95: As well as this.

librad/tests/propagation_basic.rs
```diff
@@ -112,27 +119,33 @@ async fn can_clone_disconnected() {
...
```
FintanH, line 148: This wouldn't work without the core_threads :+1:

librad/tests/propagation_basic.rs
```diff
@@ -265,24 +286,29 @@ async fn fetches_on_gossip_notify() {
...
+            .await
+            .unwrap()
+            .unwrap();
+        assert!(peer2_has_commit);
     })
     .await;
 }
 
-#[tokio::test]
+#[tokio::test(core_threads = 3)]
```
FintanH, line 311: This worked fine without core_threads

librad/tests/propagation_basic.rs
FintanH: This worked on my machine with 2 threads :shrug:

librad/tests/working_copy.rs
FintanH: Without it, I get this weird error `thread 'can_fetch' panicked at 'can call blocking only when running on the multi-threaded runtime'`

FintanH (dismissed):
These refcounts are TERRIFYING.

I did a quick look into the treading story but I wasn't enlightened. Maybe with a fresher mind :sweat_smile:

librad/tests/working_copy.rs
kim: Yeah that's expected, `block_in_place` is only available on the multithreaded runtime. The question is if it worksed if we change that to `spawn_blocking`, and remove that `core_threads` thing.

kim: Alright, so, it does work if we make `PeerStorage` always use `spawn_blocking`. I'm leaving `block_in_place` in place for tests, because it's just much more convenient to use (you don't have to move it move it). Which means that the tests where it is used need to have `core_threads > 1`.

FintanH: Ah you beat me to it :joy: 

There's only one test that needs it I think: `fetches_on_gossip_notify`

FintanH: Because I came to the same conclusion as you :+1:

FintanH: @kim the working copy tested also worked for me without the core_threads but I see you didn't change it?

xla (approved):
claps

🦄 🎋 🏕 🍙

FintanH (approved):

xla (approved):

PR #328, opened 2020-09-21
Ask and clone a project

FintanH:
Adding a test to confirm the expected steps of asking for a RadUrn
and having to clone from the PeerId that responded.

### DIFF
--- a/librad/tests/propagation_basic.rs
+++ b/librad/tests/propagation_basic.rs
@@ -384,3 +384,62 @@ async fn providers_works() {
     })
     .await;
 }
+
+#[tokio::test]
+async fn ask_and_clone() {
+    logging::init();
+    const NUM_PEERS: usize = 2;
+    let peers = testnet::setup(NUM_PEERS).await.unwrap();
+
+    testnet::run_on_testnet(peers, NUM_PEERS, async move |mut apis| {
+        let (peer1, peer1_key) = apis.pop().unwrap();
+        let mut alice = Alice::new(peer1_key.public());
+        let mut radicle = Radicle::new(&alice);
+        let resolves_to_alice = alice.clone();
+        alice
+            .sign(&peer1_key, &Signatory::OwnedKey, &resolves_to_alice)
+            .unwrap();
+        radicle
+            .sign(
+                &peer1_key,
+                &Signatory::User(alice.urn()),
+                &resolves_to_alice,
+            )
+            .unwrap();
+
+        let repo_urn = tokio::task::spawn_blocking(move || {
+            let git = peer1.storage();
+            git.create_repo(&alice).unwrap();
+            git.create_repo(&radicle).unwrap().urn
+        })
+        .await
+        .unwrap();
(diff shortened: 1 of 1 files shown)


kim (commented):
This is basically the same as `providers_worksed`, isn't it?

Perhaps we should start gardening those integration tests a bit -- they're all very IO-bound and prone to become a steaming bowl of copy-pasta carbs.

xla:
> Perhaps we should start gardening those integration tests a bit -- they're all very IO-bound and prone to become a steaming bowl of copy-pasta carbs.

Plus they are time sensitive. This PR was sparked because I had a hole in my understanding of when link automatically fetches, which I know understand is only when we track the `(project_urn, peer_id)` tuple. If we can collapse into `providers_works` - that be ideal.

FintanH:
Yup! Happy to do that.

Another question I had in my head was: would this combined functionality be useful in the API?

kim:
> Another question I had in my head was: would this combined functionality
> be useful in the API?

Ya I think this was once considered -- there is a whole bunch of timeouts and
errors to handle, though, so that would indeed be rather high-level. Perhaps
explore all of those, and how to surface them, first?

xla:
> Another question I had in my head was: would this combined functionality be useful in the API?

As it is based on `providers` it mimics a request/response pattern, which in many scenarios is not valuable. For example the entire reason for radicle-dev/radicle-upstream#903 to be this ellaborate is to account for the fact that we get to know about an interesting `Have` well outside of the timeout of `providers`, maybe after restarts and so on. IMHO we should consider building out support for these scenarios in the protocol rather than give false hope of instant availability.

xla (approved):
🕍 🖱 🏫 😺

librad/tests/propagation_basic.rs
```diff
@@ -384,3 +384,62 @@ async fn providers_works() {
...
+        let peer_id = match res {
+            Some(peer_info) => peer_info.peer_id,
+            None => panic!("Expected to have obtained peer1 but got None instead"),
+        };
+
+        assert!(!peer2.storage().has_urn(&repo_urn).unwrap());
+
+        {
+            let url = repo_urn.clone().into_rad_url(peer_id);
+            let git = peer2.storage().reopen().unwrap();
```
kim, line 387: Curious: what makes you think it should be reopened here?

FintanH: I need an owned `Storage` for the `move` in `spawn_blocking`. I actually looked at other tests to gauge what was best practice and found this https://github.com/radicle-dev/radicle-link/blob/master/librad/tests/propagation_basic.rs#L167-L179.

Do you see another way of getting around this?

NunoAlexandre (approved):
:small_airplane:

librad/tests/propagation_basic.rs
kim: I see, yeah. Will be obsolete anyways with #327

PR #329, opened 2020-09-21
Add includes paths

xla:
To support work in radicle-dev/radicle-upstream#894 while leaving the
possibility to use the same generated includes in a possible rad cli
this change adds a new path that can be used to store include files
scope to projects in the monorepo.

### DIFF
--- a/librad/src/paths.rs
+++ b/librad/src/paths.rs
@@ -29,6 +29,7 @@ use directories::ProjectDirs;
 pub struct Paths {
     keys_dir: PathBuf,
     git_dir: PathBuf,
+    git_includes_dir: PathBuf,
 }
 
 impl Paths {
@@ -46,6 +47,7 @@ impl Paths {
         Self {
             keys_dir: config_dir.join("keys"),
             git_dir: data_dir.join("git"),
+            git_includes_dir: config_dir.join("git-includes"),
         }
         .init()
     }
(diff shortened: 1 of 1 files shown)


FintanH (dismissed):
Makes sense to me :+1:

librad/src/paths.rs
```diff
@@ -46,6 +47,7 @@ impl Paths {
         Self {
             keys_dir: config_dir.join("keys"),
             git_dir: data_dir.join("git"),
+            includes_dir: config_dir.join("includes"),
```
xla, line 50: More aptly might be `git-includes` as these are specific to git repositories.

FintanH: Mmm that's a good point. Which actually makes me wonder, should the includes path just be a child of the git dir?

xla: Was wondering the same, two points:

* I see the includes as config not as data
* having more writers to the monorepo path might not be desirable

FintanH: Fair points :+1: Happy with a separate path, and happy with `git-includes`

kim (approved):
lgtm

FintanH (approved):
:kissing::ok_hand:

NunoAlexandre (approved):
:1234:

PR #330, opened 2020-09-21
Add method to paths for git includes

xla:
As a follow-up to #329  and a more convenient way to get the underlying
`&Path` than using `all_dirs`.

### DIFF
--- a/librad/src/paths.rs
+++ b/librad/src/paths.rs
@@ -80,6 +80,10 @@ impl Paths {
         &self.git_dir
     }
 
+    pub fn git_include_dir(&self) -> &Path {
+        &self.git_includes_dir
+    }
+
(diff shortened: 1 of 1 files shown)


NunoAlexandre (approved):
LGTM

librad/src/paths.rs
```diff
@@ -80,6 +80,10 @@ impl Paths {
         &self.git_dir
     }
 
+    pub fn git_include_dir(&self) -> &Path {
```
kim, line 83:
```suggestion
    pub fn git_includes_dir(&self) -> &Path {
```

?

FintanH (approved):
:clap:

NunoAlexandre (approved):

PR #331, opened 2020-09-22
Savely overwrite include files

xla:
Until now git2s `Config` was used to handle the actual file on disk for
includes. While an intuitive approach it makes it hard to reason about
the state of the actual file, as reliably replacing the entire content,
if for example the tracking graph has changed drastically, would require
building a diff and then invoke `remove` on per remote basis.

As a drop-in replacement `Include` now offers a `save` method which --
while consuming the struct -- writes the content to temporary file and
then atomically replaces the old version.


librad/src/git/include.rs
```diff
@@ -140,88 +161,114 @@ fn fetch_entry(remote: &Remote<LocalUrl>) -> (String, String) {
...
-            let include = constructor(path);
-            let config = Config::try_from(include)?;
-            Ok(config)
-        })
-    }
+    use super::*;
 
     #[test]
-    fn can_create_trivial_include() -> Result<(), Error> {
+    fn can_create_and_update() -> Result<(), Error> {
```
kim, line 171: A few suggestions here:

* I'd like to see us moving more towards `lazy_static`'d keys where it does not actually matter
* Can the checks for presence of entries not be solved more concisely by iterating using [`entries`](https://docs.rs/git2/0.13.11/git2/struct.Config.html#method.entries) with a glob?
* What happens if you move the new file while still holding on to a live `Config`? Which version does it see?

xla: > * I'd like to see us moving more towards `lazy_static`'d keys where it does not actually matter

Which keys do you think would be better placed in `lazy_static`?

> * Can the checks for presence of entries not be solved more concisely by iterating using [`entries`](https://docs.rs/git2/0.13.11/git2/struct.Config.html#method.entries) with a glob?

That should be possible.

> * What happens if you move the new file while still holding on to a live `Config`? Which version does it see?

What is the scenario? Are there foreseeable cases where something holds onto the include file as `git2::Config`?

kim:
> Which keys do you think would be better placed in `lazy_static`?

All the `SecretKey`s. You can create one `from_seed`.

> What is the scenario? Are there foreseeable cases where something holds
> onto the include file as `git2::Config`?

Yes of course -- there nothing preventing you from generating the same config
concurrently. Same for programmatic access to the working copy -- whoever
triggers re-generating the include file might do so in the background, while the
UI permits unsynchronised operations on that repo.

Unfortunately, `git2::Config` API does not make use of Rust's type system to
prevent concurrent writers statically. You will have noticed, however, that
multiple writers cause a runtime error -- you're supposed to take a snapshot for
read-only use. That aside, `libgit2` employs file locking and tries hard to
detect modifications by other processes -- the question here is if that is
subverted by just renaming the file under its buttocks. Since `rename(2)` does
not affect any open file descriptors, the `save` operation may actually not be
safe[sic!] without external synchronisation.

xla:
@kim I dunno how, but somehow git2 config gets to know about the latest values, at least judging by c38848f where the test holds on to a single open config throughout all updates.

kim:
Thanks, that's what I wanted to make sure.

kim (approved):

xla:
@kim Does d3ea3ed address your comment about using seeds and `lazy_static` for the keys?

librad/src/git/include.rs
```diff
@@ -71,12 +78,27 @@ impl<Path> Include<Path> {
         }
     }
 
-    /// Create the include file by creating a [`Config`].
-    pub fn create(self) -> Result<Config, git2::Error>
+    /// Writes the contents of the [`git2::Config`] of the include file to disk.
```
FintanH, line 81:
```suggestion
    /// Writes the contents of the [`git2::Config`] of the include file to disk.
    /// The path of the file is determined by [`Include::file_path`].
```

kim (approved):
:+1:

kim:
> d3ea3ed

Nice hash!

FintanH (approved):
My documentation suggestion is just a suggestion. Take it or leave it :)

PR #332, opened 2020-09-23
Ensure temporary include file is created

xla:
If no mutations are done to the `git2::Config` it does not create the
file on disk. To ensure that the most up-to-date version is written, the
file is touched.

### DIFF
--- a/librad/src/git/include.rs
+++ b/librad/src/git/include.rs
@@ -86,6 +86,13 @@ impl<Path> Include<Path> {
         let tmp_dir = tempfile::tempdir()?;
         let tmp_path = tmp_dir.path().join(self.local_url.repo.to_string());
 
+        // We emulate the glorious `touch` to ensure the file exists, as git2::Config
+        // doesn't seem to create on open.
+        std::fs::OpenOptions::new()
+            .create(true)
+            .write(true)
+            .open(tmp_path.clone())?;
+
         let mut config = git2::Config::open(&tmp_path)?;
 
         for remote in &self.remotes {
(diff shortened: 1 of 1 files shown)


FintanH:
Hmmm, I'd argue that we should say `remotes: NonEmpty<Remote>` instead. There's no point in creating an empty `Config`

kim (approved):
Either way seems fine to me. Although if we refuse to create empty files, we'll need to verify that `git` doesn't complain if it gets told to include a non-existing path.

xla:
> Hmmm, I'd argue that we should say `remotes: NonEmpty<Remote>` instead. There's no point in creating an empty `Config`

If we go down that route, we are back at actually dififng states. If we go from remotes to no remotes we need to check if there is a config file present, if so we have to delete it. Indiscriminately writing the file alleviates a whole host of incosistent states we need to account for.

FintanH:
> If we go down that route, we are back at actually dififng states.

I'm not sure where I see that necessity arising.

The point of the `Include` file is that it should be there iff there are `Remote`s, otherwise it's just an empty config file. So that's why I'm advocating for `NonEmpty`.

> If we go from remotes to no remotes we need to check if there is a config file present, if so we have to delete it.

If we only allow these include files to be created with `NonEmpty` then we shouldn't* be going from remotes to no remotes.

If we want some way to delete it, then I would say we would split `Include` into variants where the `Delete` variant just needs to be able to build the file path.

[0] Subject to someone editing the file manually

xla:
> If we want some way to delete it, then I would say we would split `Include` into variants where the `Delete` variant just needs to be able to build the file path.

That's a fair direction to take, do you think it should block this PR? Would you accept a follow-up issue?

FintanH:
Ya, I think I'd prefer that approach personally and see it as contending against this approach. So unless this is needed for stop-gap or something I'd rather see the other implementation.

Are you in quick need of this?

xla:
> Are you in quick need of this?

Yes it breaks upstreams tests and the track_fake_peer business in there

FintanH (approved):

xla:
Follow-up captured in #333 (satisfying issue number).

PR #334, opened 2020-09-23
Update radicle-keystore

cloudhead:
Reaping the benefits of less salt in our crypto

* [x] Fix tests


git-helpers/src/remote_helper.rs
```diff
@@ -88,7 +88,12 @@ pub fn run() -> anyhow::Result<()> {
 fn get_signer(git_dir: &Path, keys_dir: &Path, url: &LocalUrl) -> anyhow::Result<BoxedSigner> {
     let pass = credential::Git::new(git_dir).get(url)?;
     let file = keys_dir.join(SECRET_KEY_FILE);
-    let keystore = FileStorage::<_, PublicKey, _, _>::new(&file, Pwhash::new(pass));
+    let kdf_params = if cfg!(test) {
+        *radicle_keystore::crypto::KDF_PARAMS_TEST
+    } else {
+        *radicle_keystore::crypto::KDF_PARAMS_PROD
+    };
```
cloudhead, line 95: :man_shrugging:

cloudhead: Can we just use prod here?

cloudhead: :man_shrugging: :man_shrugging: :man_shrugging: :man_shrugging: :man_shrugging: :man_shrugging: :man_shrugging: :man_shrugging:

cloudhead:
Ok so, the tests are passing but I don't think this is quite right, because now the integration tests are slow, since they run in "prod" and prod uses the harder settings.

git-helpers/tests/remote.rs
```diff
@@ -112,7 +112,10 @@ fn setup_entity(paths: &Paths, key: SecretKey) -> anyhow::Result<RadUrn> {
 fn setup_keystore(dir: &Path, key: SecretKey) -> anyhow::Result<()> {
     let mut keystore = keystore::FileStorage::<_, PublicKey, _, _>::new(
         &dir.join("librad.key"),
-        keystore::crypto::Pwhash::new(SecUtf8::from(PASSPHRASE)),
+        keystore::crypto::Pwhash::new(
+            SecUtf8::from(PASSPHRASE),
+            *keystore::crypto::KDF_PARAMS_PROD,
```
xla, line 117: As this is in the context of tests shouldn't it use test params?

git-helpers/src/remote_helper.rs
xla: Will it even be conditionally compiled? AFAIU the remote helper is invoked as a binary and `cfg(test)` depends on `cargo test`.

git-helpers/tests/remote.rs
cloudhead: Yah, see my top-level comment. It should be TEST I think, but also shouldn't fail the test :sweat_smile:

cloudhead:
> Will it even be conditionally compiled? AFAIU the remote helper is invoked as a binary and cfg(test) depends on cargo test.

This is what I was wondering, so yeah we need a different solution that makes the tests not slow..

kim:
I already had some reason to do this in #327 [here](https://github.com/radicle-dev/radicle-link/pull/327/commits/e2a54418f5d69e3051aa8238ea2c8b758110bcb2). 

I couldn't find a way to make the remote helper tests use the faster params -- and I don't want this to be configurable by the user. One thing to try out would be to check at runtime whether `CARGO_BIN_EXE_git-remote-rad` is set.

cloudhead:
> I already had some reason to do this in #327 [here](https://github.com/radicle-dev/radicle-link/pull/327/commits/e2a54418f5d69e3051aa8238ea2c8b758110bcb2).
> 

Ah great.

> I couldn't find a way to make the remote helper tests use the faster params -- and I don't want this to be configurable by the user. One thing to try out would be to check at runtime whether `CARGO_BIN_EXE_git-remote-rad` is set.

Hmm. I mean it's not *that* slow I guess. We could also `#[ignore]` and have CI additionally run with `--ignored`.

PR #335, opened 2020-09-23
Add helper to set include directive

xla:
Commonly in tandem to the generation of the include file itself, working
copies need to be equipped with an include directive pointing to said
files. To avoid miss-spellings in the config key this change exports
a helper for that.

### DIFF
--- a/librad/src/git/include.rs
+++ b/librad/src/git/include.rs
@@ -30,6 +30,9 @@ use crate::{
     peer::PeerId,
 };
 
+/// Config key to reference generated include files in working copies.
+const GIT_CONFIG_PATH_KEY: &str = "include.path";
+
 #[derive(Debug, thiserror::Error)]
 #[non_exhaustive]
 pub enum Error {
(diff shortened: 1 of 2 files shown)


FintanH (approved):
:+1:

librad/src/git/include.rs
```diff
@@ -30,6 +30,9 @@ use crate::{
     peer::PeerId,
 };
 
+/// Config key to reference generated include files in working copies.
+const GIT_CONFIG_PATH_KEY: &str = "include.path";
```
FintanH, line 34:
```suggestion
pub const GIT_CONFIG_PATH_KEY: &str = "include.path";
```

FintanH (changes requested):
Scratch that

librad/src/git/include.rs
FintanH: I think it would still make sense to have this public even though you're exposing the function publically. Say for example, someone wanted to test for this being set.

NunoAlexandre (approved):

PR #336, opened 2020-09-24
Menage a troi

FintanH:
We add a test that does the following:
* Peer 1 creates a project
* Peer 1 makes a commit to their default branch
* Peer 2 clones from Peer 1
* Peer 3 clones from Peer 2

The expectation is that Peer 2 has Peer 1's `heads/*`, which it does.
It is also expected that Peer 3 has Peer 1's `heads/*`, which it doesn't.

### DIFF
--- a/librad/src/git/types.rs
+++ b/librad/src/git/types.rs
@@ -213,14 +213,19 @@ impl Refspec<PeerId, PeerId> {
             // Heads
             //
             // `+refs/namespaces/<namespace>/refs[/remotes/<peer>]/heads/* \
-            // :refs/namespaces/<namespace>/refs/remotes/<peer>/refs/heads/*`
+            // :refs/namespaces/<namespace>/refs/remotes/<peer>/heads/*`
             {
                 let their_singed_rad_refs = rad_signed_refs_of(tracked_peer.clone())?;
                 for (name, target) in their_singed_rad_refs.heads {
                     let name_namespaced = format!("refs/namespaces/{}/{}", namespace, name);
                     if let Some(name) = name.strip_prefix("refs/heads/") {
+                        let name_namespaced_remote = format!(
+                            "refs/namespaces/{}/refs/remotes/{}/heads/{}",
+                            namespace, tracked_peer, name
+                        );
                         let targets_match = remote_heads
                             .get(name_namespaced.as_str())
+                            .or_else(|| remote_heads.get(name_namespaced_remote.as_str()))
                             .map(|remote_target| remote_target == &*target)
                             .unwrap_or(false);
 
--- a/librad/tests/propagation_basic.rs
+++ b/librad/tests/propagation_basic.rs
@@ -447,7 +447,7 @@ async fn ask_and_clone() {
     .await;
 }
 
-#[tokio::test(core_threads = 3)]
+#[tokio::test(core_threads = 2)]
 async fn menage_a_troi() {
     logging::init();
 
--- a/librad/tests/propagation_basic.rs
+++ b/librad/tests/propagation_basic.rs
@@ -17,14 +17,18 @@
 
 #![feature(async_closure)]
 
-use std::time::Duration;
+use std::{marker::PhantomData, time::Duration};
 
 use futures::{future, stream::StreamExt};
 use tempfile::tempdir;
 use tokio::task::block_in_place;
 
 use librad::{
-    git::{local::url::LocalUrl, storage},
+    git::{
+        local::url::LocalUrl,
+        storage,
+        types::{remote::Remote, FlatRef, Force, NamespacedRef},
+    },
     meta::{entity::Signatory, project::ProjectInfo},
     net::peer::{FetchInfo, Gossip, PeerEvent, Rev},
     signer::SomeSigner,
@@ -32,6 +36,7 @@ use librad::{
 };
 
 use librad_test::{
+    git::initial_commit,
     logging,
     rad::{
         entity::{Alice, Radicle},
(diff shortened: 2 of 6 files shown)


librad/tests/propagation_basic.rs
```diff
@@ -396,3 +397,86 @@ async fn ask_and_clone() {
     })
     .await;
 }
+
+#[tokio::test]
+async fn menage_a_troi() {
```
xla, line 402: #HRRRRRRRR

kim:
I think the issue is that `Refspec::fetch_heads` doesn't consider `rad/signed_ref`, which means that `peer3` doesn't receive `peer1`'s through `peer2`, which in turn means it is going to ignore `peer1`'s heads (on `peer2`), because it has no authorship proof.

FintanH:
Hmmm, I'm not sure I'm following. Via my inspection of logs, when peer3 clones from peer2 it's looking at peer1's signed_ref, but it fails the check mentioned yesterday: https://github.com/radicle-dev/radicle-link/blob/master/librad/src/git/types.rs#L222-L225.

But is that what you're saying, it doesn't consider it because of this check?

kim:
> clones from peer2 it's looking at peer1's signed_ref

Yeah, but does it actually _have_ them?

FintanH:
> Yeah, but does it actually _have_ them?

Well the fact we iterate over it https://github.com/radicle-dev/radicle-link/blob/master/librad/src/git/types.rs#L219 would suggest to me that we do. Or am I misunderstanding?

FintanH:
Right, so looking at logs some more I think I see what you're getting at. The ref we're looking for isn't in the `remote_heads` so that's why the `bool` ends up as `false`.

Can you elaborate a bit more why this is the case? I'm not really understanding the above :innocent:

kim:
[`fetch_heads`](https://github.com/radicle-dev/radicle-link/blob/1d9c4c107e6ac1816342a35c0558262fba4a96ff/librad/src/git/types.rs#L201) receives the list of advertised refs from the other side (`remote_heads`), and computes the subset which we actually want to fetch. We can't use any wildcards here, so must enumerate all refs we want. 

What it clearly doesn't include in this list is `refs/namespaces/<namespace>/refs/remotes/<peer>/rad/signed_refs` for each `tracked_peer`. So that means we're not going to fetch it.

I think what would greatly enhance the debugging experience would be to make `Fetcher::prefetch` and `Fetcher::fetch` collect which refs we actually got (via the `update_tips` callback). I remember we talked about this, but nobody picked it up. Right now, much is _assumed_ to be present in the repo after a fetch, but we don't really verify that that's the case.

FintanH:
So this seems of interest. Inspecting some log output we have:
* The reference we're looking for `refs/namespaces/hwd1yre8c8tqmkpsp3u6r9zj7aixn7s86u1nfzpsaeng9hq599shub59tco/refs/heads/master`
* For Peer 2 cloning the repo it appears in the remote heads: 
  ```
  "refs/namespaces/hwd1yre8c8tqmkpsp3u6r9zj7aixn7s86u1nfzpsaeng9hq599shub59tco/refs/heads/master": 5bca421266cc1a8ecdc585aed98048b9ed086850
  ```
* For Peer 3 it appears but the ref looks a bit strange:
  ```
  "refs/namespaces/hwd1yre8c8tqmkpsp3u6r9zj7aixn7s86u1nfzpsaeng9hq599shub59tco/refs/remotes/hyyx4hr3hqi43jo8tua3ct1indk9rwxkriedepw4bmbtjd9tz5zwbe/heads/master": 5bca421266cc1a8ecdc585aed98048b9ed086850
  ```

For some reason, it seems to be tucked under as a remote of Peer 1, where that remote I'm presuming is Peer 2.

kim (commented):
Ok, so that looks like Peer2 is not correctly mapping to
`refs/namespaces/<project>/remotes/<peer1>/heads/master`

FintanH:
Oh sorry, I'm wrong it is mapped to Peer 1. Got myself confused. However, in fetch heads it doesn't seem like we account for looking under a remote. We just look at the top level `refs/namespaces/hwd1yre8c8tqmkpsp3u6r9zj7aixn7s86u1nfzpsaeng9hq599shub59tco/refs/heads/master`, instead of that _and_ `refs/namespaces/hwd1yre8c8tqmkpsp3u6r9zj7aixn7s86u1nfzpsaeng9hq599shub59tco/remotes/*/heads/master`. Does that sound right to you?

librad/src/git/types.rs
```diff
@@ -222,9 +222,12 @@ impl Refspec<PeerId, PeerId> {
                 for (name, target) in their_singed_rad_refs.heads {
                     let name_namespaced = format!("refs/namespaces/{}/{}", namespace, name);
                     if let Some(name) = name.strip_prefix("refs/heads/") {
+                        let name_namespaced_remote = format!("refs/namespaces/{}/refs/remotes/{}/heads/{}", namespace, tracked_peer, name);
```
FintanH, line 225: @kim this fixed it. Does it make sense to you? It seems to make sense to me.

kim: :ok_hand:

FintanH: :raised_hands:

librad/tests/propagation_basic.rs
```diff
@@ -458,3 +446,111 @@ async fn ask_and_clone() {
     })
     .await;
 }
+
+#[tokio::test(core_threads = 3)]
```
kim, line 450: Does it _require_ 3?

FintanH: Not strictly :innocent:

xla: It requries troi.

FintanH:
Any reviews going? :eyes:

kim (dismissed):

FintanH:
Had to rebase again because of the docs additions :facepalm:

kim (approved):

PR #338, opened 2020-09-24
Let testnet converge

### LINKED ISSUE #313: librad-test: Wait for join completion instead of connections
The `testnet` harness blocks until `n` `Connected` events have been received.
This doesn't mean, however, that the respective other end will respond to gossip
messages sent from this end -- the protocol has to complete the membership join
processes first.

We'll need a way to bubble up more gossip state changes, and change the test
harness to match on the proper one.

xla:
Closes #313

***

This change-set is merely an exercise to probe if I can develop some
limited intuition how data flows through the network stack. A lot of
open questions I hope to get challenged on:

* Is exposing `Joined` alongside `Gossip` as high-level protocol event
  variant sensible?
* Should membership events be surfaced differently?
* Is it correct that only connecting peers will send `Join` messages?
* Is `Neighbour` as significant as it is send by peers which already are
  connected?
  * If that is the case should we extend the testnet harness to wait for
    the right combination of `Join` and `Neighbour`?


FintanH:
One thing you could do is check that it works for the `ask_and_clone` test, right? That's where we first realised this was a problem, iirc.

xla:
@FintanH The changes apply to all integration tests afaik.

FintanH:
Sorry, I wasn't clear. What I meant was that this spawn blocking shouldn't be needed if this works: https://github.com/radicle-dev/radicle-link/blob/master/librad/tests/propagation_basic.rs#L363-L369

xla:
> Sorry, I wasn't clear. What I meant was that this spawn blocking shouldn't be needed if this change works: https://github.com/radicle-dev/radicle-link/blob/master/librad/tests/propagation_basic.rs#L363-L369

That block performs blocking I/O and as we learned to truly have a-kitchen-sync these need to run on blocking threads, hence the work in #327. I do wonder so what a good scenario would be to make sure the change has the desired effect.

FintanH:
> That block performs blocking I/O

I don't think that's true. I tested the same code by just adding a sleep after the project creation with the same effect. I was wonder under the impression it was due to this synchronisation issue.

xla:
It definitely does, looking at `has_urn` and `has_ref`, which both are used:

https://github.com/radicle-dev/radicle-link/blob/99db345f9c091e6d6eb1b8529b3815f1ce02c6e1/librad/src/git/storage.rs#L310-L325

WIth the upcoming changes to how we handle storage resources we won't be able to control how this I/O is done anyway, which makes relying on it more incidental. It would be preferrable to have a setup where we don't have to think about the runtime internals of tokio and/or if we parameterised the test scheduler correctly. Could you show a diff with the sleep you mentioned?

FintanH:
So I'm only going off of what @kim had mentioned in the PR where the test was created, see:
* https://github.com/radicle-dev/radicle-link/pull/309#discussion_r485534688
* https://github.com/radicle-dev/radicle-link/pull/309#discussion_r485557371

> Could you show a diff with the sleep you mentioned?

Sorry, the diff was just on my local. It was essentially a call to `thread::sleep` instead of the blocking call.

librad/src/net/gossip.rs
```diff
@@ -111,6 +112,18 @@ where
...
 }
 
+#[derive(Clone, Debug)]
+pub enum MembershipInfo<Addr>
+where
+    Addr: Clone + Eq + Hash,
+{
+    Join {
+        advertised_info: PeerAdvertisement<Addr>,
+        remote_id: PeerId,
```
kim, line 122: Do you think it is significant to surface this? It's just the last hop of the random walk a `Join` message takes. There might be a case for tracing the entire path, but probably only in a debug build.

librad/tests/convergence.rs
```diff
@@ -0,0 +1,40 @@
...
+
+#[tokio::test]
+async fn converge() -> Result<(), Box<dyn std::error::Error>> {
+    logging::init();
+
+    for num_peers in 1..10 {
+        let peers = testnet::setup(num_peers).await?;
+        timeout(
+            Duration::from_secs(1),
+            testnet::run_on_testnet(peers, num_peers, |mut _apis| async move {
```
kim, line 32: I'm not sure this actually tests anything :thinking: :joystick:

Interestingly, I'm not even sure if it would make sense to rely on snapshotting the set of currently-connected peers for each peer -- it depends on the network size and the protocol parameters when the network reaches a "steady state". We'll need to make the test harness more modular I think, so we can have e.g. a test for broadcast atomicity (send message, observe all peers eventually receive it) without carrying the git baggage.

xla: > I'm not sure this actually tests anything :thinking: :joystick:

Just validates that the counting and boolean logic in `wait_converged` is doing what it should. It was a useful helper while working on the message accumulation.

> Interestingly, I'm not even sure if it would make sense to rely on snapshotting the set of currently-connected peers for each peer -- it depends on the network size and the protocol parameters when the network reaches a "steady state". We'll need to make the test harness more modular I think, so we can have e.g. a test for broadcast atomicity (send message, observe all peers eventually receive it) without carrying the git baggage.

Absolutely. Ideally the outcome is that we have a test suite with different param configs and actually assert the partial view per peer. Would leave that for a future exercise, unless it is critical to accept this PR.

librad/src/net/gossip.rs
xla: Don't think so, unless an explicit use-case comes up. Was more mindless mirror of the fields of the lower-level variant.

librad/tests/convergence.rs
xla: #344

xla:
@FintanH @kim This should be in a reasonable state now.

FintanH (approved):
:metal:

kim (approved):
lgtm

PR #340, opened 2020-09-28
Move Seed to bins

### LINKED ISSUE #339: Operationalise Seed
As we want to be living better through netzwerking together, it's vital that we make it simple to set up and operate seed nodes. For improved discovery, data availability and potentially future use-cases like hole punching/relaying. To do that there will likely be a surge in infrastructure around the Seed to run it on <INSERT_ENTERPRISE_SERVERLESS_SOLUTION>. The fact that the seed currently lives here in link has the nice side-effect that any breaking changes need to be immediately addressed in lockstep with the changes to librad. Yet we should be cautious further expanding the "monorepo", especially with non-essential code - especially when we wan this to be the place for the specification and implementation of the core protocol. Maintaining a high signal to noise ratio. I see three options going forward:

* we keep the seed in link and add auxilliary code into its sub-directory
  * lowest amount of fricition
  * non-essentials in the protocol repo
* we maintain the core seed functionality (a node of sort) in link and bootstrap `radicle-seed` which is a binary consuming the node
  * lockstep fixes to the seed node
  * extra infrastructure is kept outside of protocol repo
  * a distinct place to point to for people to look at docs and start running seeds
* we move all of the seed into a new repo `radicle-seed`
  * extra infrastructure is kept outside of protocol repo
  * a distinct place to point to for people to look at docs and start running seeds
  * maintainenance overhead from keeping the seed up-to-date

xla:
Closes #339


kim:
Oh - do you not think the `lib` half should stay here?

xla:
> Oh - do you not think the `lib` half should stay here?

Yeah, was too eager. Miss-understood the conversation in #339.

PR #342, opened 2020-10-01
Prevent potential cross-filesytem rename when creating include

### LINKED ISSUE #337: Cannot create include files on NixOS
The error:
```
Include(Io(Os { code: 18, kind: Other, message: "Invalid cross-device link" })))
```

Relevant documentation: https://doc.rust-lang.org/std/fs/fn.rename.html#platform-specific-behavior.

It's possible that my tmp is mounted on a different `dev` and so when we try to create a tmp directory and rename it to the git-include location it fails with the above.

kim:
Fixes #337

### DIFF
--- a/librad/src/git/include.rs
+++ b/librad/src/git/include.rs
@@ -21,6 +21,8 @@ use std::{
     path::{self, PathBuf},
 };
 
+use tempfile::NamedTempFile;
+
 use crate::{
     git::{
         local::url::LocalUrl,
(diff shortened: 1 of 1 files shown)


librad/src/git/include.rs
```diff
@@ -21,6 +21,8 @@ use std::{
     path::{self, PathBuf},
 };
 
+use tempfile::NamedTempFile;
```
FintanH, line 24: Ahhh that's what we were looking for @xla :smile: Remembering that we couldn't get the path of a `File`.

FintanH (approved):
Nice :metal:

PR #343, opened 2020-10-01
Bump ed25519-zebra to released version (2.2.0)

kim:
Funnily enough, I noticed both the source and the released version being built.
Maybe this combo of version + branch + git is a bit, errm, non-deterministic?


cloudhead (approved):
Nice

PR #345, opened 2020-10-02
Ban openssl crates

kim:
We don't want any openssl, because a. its license is not GPL-compatible
(even though the crates are licensed differently), and b. we need a
tight grip on what crypto we pull in.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -173,6 +173,8 @@ deny = [
     # Each entry the name of a crate and a version range. If version is
     # not specified, all versions will be matched.
     #{ name = "ansi_term", version = "=0.11.0" },
+    { name = "openssl-probe" },
+    { name = "openssl-sys" },
 ]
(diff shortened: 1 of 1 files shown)


kim:
Nb. `upstream` will likely need to review the features config for their `git2` dependency -- if `default-features` is enabled there, it will pull in `openssl-*` through the direct dep.

xla:
> Nb. `upstream` will likely need to review the features config for their `git2` dependency -- if `default-features` is enabled there, it will pull in `openssl-*` through the direct dep.

I believe Upstream depends on `git2` transitively through librad/surf.

xla (approved):
🔘 ➖ 📚 🏋

kim:
> I believe Upstream depends on `git2` transitively through librad/surf.

Ya seems so. I was convinced it needed at least a dev-dependency.

kim:
Closed via c29aa348b34addf148be0d0c328d35a4f887a3f9

PR #350, opened 2020-10-06
Networking: 0RTT stream upgrade

kim:
The is no reason for the roundtrip, so better make this change now.
Also, refine the CBOR encoding of UpgradeRequest.

Note that this is a breaking change.

### DIFF
--- a/librad/src/net/upgrade.rs
+++ b/librad/src/net/upgrade.rs
@@ -19,36 +19,63 @@
 //! (irrespective of ALPN, which applies per-connection). This module implements
 //! the negotiation protocol.
 
-use std::{io, marker::PhantomData, ops::Deref, pin::Pin};
+use std::{
+    fmt::{self, Debug, Display},
+    io,
+    marker::PhantomData,
+    ops::Deref,
+    pin::Pin,
+    time::Duration,
+};
 
 use futures::{
-    future::{BoxFuture, FutureExt},
-    io::{AsyncRead, AsyncWrite},
-    sink::SinkExt,
-    stream::TryStreamExt,
+    future::{self, TryFutureExt as _},
+    io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt},
     task::{Context, Poll},
 };
-use futures_codec::Framed;
-use minicbor::{Decode, Encode};
+use futures_timer::Delay;
 use thiserror::Error;
 
-use crate::{
-    git::p2p::transport::GitStream,
-    net::{
-        codec::{CborCodec, CborCodecError},
-        quic,
-    },
-};
+use crate::git::p2p::transport::GitStream;
 
+/// Timeout waiting for an [`UpgradeRequest`].
+// NOTE: This is a magic constant, which should account for very slow
+// links. May need to become a protocol config parameter if we see very busy
+// nodes time out a lot.
+const RECV_UPGRADE_TIMEOUT: Duration = Duration::from_secs(23);
+
+/// Length in bytes of the CBOR encoding of [`UpgradeRequest`].
+///
+/// We use this to allocate only a fixed-size buffer, and not deal with
+/// unconsumed bytes.
+// NOTE: Make sure to adjust in case [`UpgradeRequest`] gains larger variants.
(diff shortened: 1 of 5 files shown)


librad/src/net/upgrade.rs
```diff
@@ -19,36 +19,60 @@
...
-    git::p2p::transport::GitStream,
-    net::{
-        codec::{CborCodec, CborCodecError},
-        quic,
-    },
-};
+use crate::git::p2p::transport::GitStream;
+
+/// Timeout waiting for an [`UpgradeRequest`].
+const RECV_UPGRADE_TIMEOUT: Duration = Duration::from_secs(23);
```
FintanH, line 42: That's _very_ specific. What's special about 23?

librad/src/net/upgrade.rs
```diff
@@ -19,36 +19,60 @@
...
+
+/// Length in bytes of the CBOR encoding of [`UpgradeRequest`].
+///
+/// We use this to allocate only a fixed-size buffer, and not deal with
+/// unconsumed bytes. Make sure to adjust in case [`UpgradeRequest`] gains
+/// larger variants.
+const UPGRADE_REQUEST_ENCODING_LEN: usize = 4;
```
FintanH, line 49: Would the tests break if a new variant was added?

I'm wondering how we can catch this early if someone added a variant but didn't update this constant.

librad/src/net/upgrade.rs
kim: It is a magical constant handed down from generations and generations of networks programmors. Also, it is very convenient to type.

FintanH: :mage:

librad/src/net/upgrade.rs
kim: Yes, `test_upgrade` will break if 4 bytes is not enough

FintanH (approved):
gossip girl

librad/src/net/upgrade.rs
kim: The truth is, I don't know what that number should be. It should wait for long enough to account for very slow links, but then also not _that_ long, and who knows what that `async` runtime is up to.

When you see me using one of `23`, `69` or `42`, it is usually a "dunno number".

FintanH: Cool :+1: Do you think it would be worth tossing that fact in for anyone else that comes across this?

FintanH (dismissed):
:star2:

librad/src/net/upgrade.rs
```diff
@@ -19,36 +19,63 @@
...
+/// protocol upgrade, it shall simply close the stream.
+///
+/// # Wire Encoding
+///
+/// The message is encoded as a 2-element CBOR array, where the first element is
+/// the (major) version tag (currently `0` (zero)). The second element is of
+/// CBOR major type 0 (unsigned integer), with the value being the `u8`
+/// discriminator of the enum. This allows _compatible_ changes to
+/// [`UpgradeRequest`] (ie. both ends can handle the absence of a variant), as
+/// well as _incompatible_ evolution by incrementing the version tag.
```
xla, line 74: Flagging this for inclusion in the spec overhaul when expanding on #354

librad/src/net/upgrade.rs
xla: Is the upgrade timeout distinct from a normal time out to first byte? As we don't have any proof of work in the upgade I wonder how much signal is there.

xla (dismissed):
Nice improvement.

💻 🍨 🌶 🏠

librad/src/net/upgrade.rs
kim: That's a good point. I don't think there is any timing out on `quinn::RecvStream` -- which does make sense: in QUIC, a stream is created by writing data to it (essentially). So, we must have received a `STREAM` frame, but not necessarily the 4 bytes we're looking for. I assume it would also be legit that this frame is empty (has no data), e.g. when it falls on the boundary of a UDP packet.

Since we have determined that we need to send keep-alives, a peer could just tar pit us by not sending anything.

PR #355, opened 2020-10-07
Add a section about URNs

rudolfs:
I couldn't find a clear definition of URNs anywhere in the docs when doing my research for https://github.com/radicle-dev/radicle-upstream/issues/840, and thought it might be a good idea to write down my findings.

Note: the style might not match the rest of the document yet, I'm open for suggestions!

### DIFF
--- a/docs/spec/identities.md
+++ b/docs/spec/identities.md
@@ -88,7 +88,47 @@ sections.
 
 The `root` of a verified `Identity` is the stable identifier of the repository.
 
-> **TODO** Insert specification of URNs here
+### Radicle URNs
+
+Entities within the Radicle network are uniquely identified by a Radicle URN
+which fits the functional definition of a URN as defined in [rfc8141].
+
+Note: Radicle URNs are not prefixed by `urn:`, because they are not officially
+registered.
+
+This is what a typical Radicle URN looks like:
+```
+rad:git:hwd1yredksthny1hht3bkhtkxakuzfnjxd8dyk364prfkjxe4xpxsww3try
+```
+
+If we'd map it onto a [rfc8141] URN, then the Namespace Identifier (NID) part
+would be `rad` and the Namespace Specific String (NSS):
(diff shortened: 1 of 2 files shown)


docs/spec/identities.md
```diff
@@ -88,7 +88,47 @@ sections.
...
+If we'd map it onto a [rfc8141] URN, then the Namespace Identifier (NID) part
+would be `rad` and the Namespace Specific String (NSS):
+`git:hwd1yredksthny1hht3bkhtkxakuzfnjxd8dyk364prfkjxe4xpxsww3try`.
+
+The NSS part for our `rad` NID can be further broken down into:
+
+```
+<PROTOCOL>:<ID>/<PATH>
+```
+Where `<PROTOCOL>` can be either `git` or `pijul`, `<ID>` is a 59 character long
```
FintanH, line 113: I wouldn't mention `pijul` here since it can't actually be `pijul` (yet) :smile:

docs/spec/identities.md
```diff
@@ -88,7 +88,47 @@ sections.
...
+base64 encoded hash which uniquely identifies the resource and the `<PATH>`
+semantics depend on whether the Radicle URN represents a user or a project.
```
FintanH, line 115: I would say the semantics depends on what the protocol is. Giving the example that the `<PATH>` points to references in the case of `git`.

It might also be noteworthy that an empty path implies a path `/rad/id` which is a git ref that points to the identity document.

The difference between projects and users is only that we don't _expect_ users to have `heads` or `issues`, but this isn't enforced in any way.

FintanH (commented):
Thanks for helping out on adding this :grin:

kim:
Thanks @rudolfs. There's some more pedantry to apply, though, which I did in #356

rudolfs:
> Thanks @rudolfs. There's some more pedantry to apply, though, which I did in #356

Ya, #356 was supposed to just be a starting point & incentive : P

rudolfs:
Superseded by #356.

rudolfs:
Ah, didn't realise that #356 was supposed to be merged into this branch.

PR #356, opened 2020-10-08
Specify URNs

kim:
Thanks @rudolfs for beating me to it!

### DIFF
--- a/docs/spec/identities.md
+++ b/docs/spec/identities.md
@@ -90,45 +90,56 @@ The `root` of a verified `Identity` is the stable identifier of the repository.
 
 ### Radicle URNs
 
-Entities within the Radicle network are uniquely identified by a Radicle URN
-which fits the functional definition of a URN as defined in [rfc8141].
+Identities are addressable within the Radicle Network by their stable identifier,
+encoded as a URN. Radicle URNs are syntactically and functionally equivalent to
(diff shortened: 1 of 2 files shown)


FintanH (approved):
:kissing::metal:

rudolfs:
> Thanks @rudolfs for beating me to it!

Ya, #355 was supposed to just be a starting point & incentive : P

rudolfs:
I know that there is not supposed to be a distinction between projects and users, but in practice those are the only entities that can be addressed at the moment, would be nice to mention that somehow. Maybe in the examples section?

rudolfs (approved):

kim:
> there is not supposed to be a distinction between projects and users

The difference is specified in [Delegations](https://github.com/radicle-dev/radicle-link/blob/3a593a1a11398830f0da2f877a10d2303de1f380/docs/spec/identities.md#delegations), but it is not relevant for describing what the format of a URN is. Because, you don't know what you'll get when resolving a URN!

I agree that this is rather awkward, and I've proposed in the past to include the _kind_ of identity (content-type, if you wish) a URN is supposed to resolve to in the URN itself. The counter-argument is that you have to deal with the situation that there is a mismatch _in any case_ -- it is validated only after the fact, so technically it would even be possible to change the type if there are enough signers.

PR #359, opened 2020-10-08
Gossip: Quick fix to not remember local peer as "known"

kim:
We can't connect to ourselves, so there is no point in remembering. This
invariant should probably be maintained by the `KnownPeers` datastructure,
but this will do as a quick fix.


FintanH (approved):
:+1:

xla (approved):
😁 🌜 👈 💌

PR #360, opened 2020-10-09
Gossip: Fix shuffle random walk termination condition

kim:
Those booleans.


FintanH (approved):
True

xla (approved):
True || True

🔎 🚡 👨 💥

PR #361, opened 2020-10-09
Transport: Rely on destructors for wait(2)

kim:
Dreadful, but the wait thread may actually terminate when _everything
else_ has been dropped already. When using temporary directories for
tests, this leads to a race where the post-service hook is run after the
tempdir was already destroyed.

---

Fixes https://github.com/radicle-dev/radicle-upstream/pull/1010#issuecomment-705558553

### DIFF
--- a/librad/src/git/local/transport.rs
+++ b/librad/src/git/local/transport.rs
@@ -21,7 +21,6 @@ use std::{
     path::PathBuf,
     process::{Child, ChildStdin, ChildStdout, Command, ExitStatus, Stdio},
     sync::{Arc, Mutex, Once, RwLock},
-    thread,
 };
 
 use git2::transport::{Service, SmartSubtransport, SmartSubtransportStream, Transport};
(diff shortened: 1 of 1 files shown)


FintanH (dismissed):
Just to double-check I'm understanding correctly: we're tying `Connected` to `LocalStream` and ensuring that it waits and gets dropped when `LocalStream` is dropped?

kim:
drop on :drum:

FintanH:
Should I confirm that it works in the checkout branch or did you beat me to it?

kim:
The error disappears when pointing your branch to this one, but confirm is better.

kim:
Just to clarify how this works: we run `update_refs` as an "application-level" post-receive hook, because otherwise we would need to install an _actual_ git hook somewhere, and that hook needs to get at the credentials for the keystore. If this transport is used with the git helper, the user will see an error message if `update_refs` fails -- but the push itself succeeds. Just running push again should converge the state (unless something is completely wrong with the storage). If it is run programmatically (through `git2`), there is just no way to know if there was an error.

FintanH:
Looks good on my end :metal:

PR #362, opened 2020-10-09
Make PublicKey & PeerId Copy

### LINKED ISSUE #348: Make PublicKey Copy
It was a n00b oversight to not derive `Copy` for `PublicKey` (and, by extension,
`PeerId`). It is almost always impractical to pass those types around by
reference, and given it's only 32 bytes not really worth the trouble.

Deriving `Copy` will create screenfuls of linter warnings, though.

FintanH:
Fixes #348 

We derive Copy for PublicKey & PeerId since they're simply a slice of
bytes.
The majority of the changes are removing the clone calls on
PeerId/PublickKey and removing the ask by reference on PeerId in
function calls.

### DIFF
--- a/librad/src/git/p2p/url.rs
+++ b/librad/src/git/p2p/url.rs
@@ -191,7 +191,7 @@ impl<'a> GitUrlRef<'a> {
     where
         Addrs: AsRef<[SocketAddr]>,
     {
-        Self::from_rad_urn(url.urn, local_peer, url.authority, addr_hints)
+        Self::from_rad_urn(url.urn, local_peer, &url.authority, addr_hints)
     }
 
     pub fn from_rad_urn<Addrs>(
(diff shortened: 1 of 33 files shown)


librad/src/git/p2p/url.rs
```diff
@@ -165,27 +165,23 @@ impl Into<RadUrl> for GitUrl {
 
 #[derive(Debug, PartialEq)]
 pub struct GitUrlRef<'a> {
-    pub local_peer: &'a PeerId,
-    pub remote_peer: &'a PeerId,
+    pub local_peer: PeerId,
```
kim, line 168: Does clippiyeahyeah complain about using references here? I'd expect these `*Ref<'a>` types to be "zero-cost" and not do any cloning or copying.

FintanH: Nope, I was overzealous

FintanH: Reverted those changes

xla (dismissed):
(*´◒`*)

FintanH:
Something weird happened with the rebase and I have no idea what. It seemed to just apply cleanly but now all Kim's changes are in here o.O

FintanH:
Closing this in favour of #366 because I don't wanna mess around with git to figure out what happened :grimacing:

Issue #353, opened 2020-10-07
Split integration tests


kim:
`propagation_basic` was meant to test some, err, basic things.

Before it gets out of hand, we should split into a couple more modules. As those tests are becoming more heavy on the IOs, this also prepares for running them concurrently and/or behind a `slow` feature.

kim:
#508

PR #363, opened 2020-10-09
CI: Run tests verbosely

kim:
Strangely, `cargo` can be told to run only the unit or doc tests, but not to
only run the integration tests. Perhaps we should meditate over organising those
slow tests in a different way to work around the deficiencies of the tooling
(see also #353).

### DIFF
--- a/ci/build-test
+++ b/ci/build-test
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 set -eou pipefail
 
-echo '--- Build & Test'
-cargo test --all
+echo '--- Build'
+cargo build --verbose --workspace
+
+echo '--- Test'
+GIT_TRACE2=1 RUST_LOG=librad=trace cargo test --workspace


ci/build-test
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 set -eou pipefail
 
-echo '--- Build & Test'
-cargo test --all
+echo '--- Build'
+cargo build --verbose --workspace
+
+echo '--- Test'
+GIT_TRACE2=1 RUST_LOG=librad=trace cargo test --workspace
```
FintanH, line 8: So does this skip doc tests?

kim:
So does this skip doc tests?
>

Does it? I was looking for a way to only run the integration tests (which
doesn’t exist), and the help told me that `—all` is deprecated in favour of
`—workspace`. So it seems like it should do the same as before.

>

FintanH (approved):
Ah I just got confused with the commentary. Didn't realise all was deprecated 🐈

PR #364, opened 2020-10-09
Transport: Busy-wait for SIGCHLD

kim:
We may otherwise wrongly terminate the child, where we simply haven't received SIGCHLD yet.

---

Astonishing that a sytems programming language doesn't have better, cross-platform tools for these kinds of things in the standard library. Most likely, this will totally not work on Windows.


FintanH (approved):
Very good, very manly

PR #366, opened 2020-10-12
Make PublicKey & PeerId Copy

### LINKED ISSUE #348: Make PublicKey Copy

FintanH:
Fixes #348

Supersedes #362 

We derive Copy for PublicKey & PeerId since they're simply a slice of
bytes.
The majority of the changes are removing the clone calls on
PeerId/PublickKey and removing the ask by reference on PeerId in
function calls.

### DIFF
--- a/librad/src/net/gossip.rs
+++ b/librad/src/net/gossip.rs
@@ -182,26 +182,24 @@ where
         if !self.peers.contains_key(&peer_id) && self.peers.len() + 1 > self.max_peers {
             self.peers.insert(peer_id, sink);
 
-            let eject = self
-                .peers
-                .keys()
-                .choose(&mut self.rng)
-                .expect("Iterator must contain at least 1 element, as per the if condition. qed")
-                .clone();
-            self.remove(&eject)
+            let (eject, _) = self
+                .random()
+                .expect("Iterator must contain at least 1 element, as per the if condition. qed");
+            self.remove(eject)
         } else {
-            self.peers
-                .insert(peer_id.clone(), sink)
-                .map(|s| (peer_id, s))
+            self.peers.insert(peer_id, sink).map(|s| (peer_id, s))
         }
     }
 
-    fn remove(&mut self, peer_id: &PeerId) -> Option<(PeerId, S)> {
-        self.peers.remove(peer_id).map(|s| (peer_id.to_owned(), s))
+    fn remove(&mut self, peer_id: PeerId) -> Option<(PeerId, S)> {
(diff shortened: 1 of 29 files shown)


librad/src/net/gossip.rs
```diff
@@ -848,7 +846,7 @@ where
     async fn broadcast<'a, M, X>(&self, rpc: M, excluding: X)
```
kim, line 846: No complaint from clippy about unused lifetime or something?

librad/src/net/gossip.rs
```diff
@@ -880,9 +878,9 @@ where
         .await
     }
 
-    async fn reply<M: Into<Rpc<Addr, Broadcast>>>(&self, to: &PeerId, rpc: M) {
+    async fn reply<M: Into<Rpc<Addr, Broadcast>>>(&self, to: PeerId, rpc: M) {
```
kim, line 881: I think here we really only need a reference

librad/src/uri.rs
```diff
@@ -316,7 +316,7 @@ impl RadUrn {
 
     pub fn as_rad_url_ref<'a>(&'a self, authority: &'a PeerId) -> RadUrlRef<'a> {
         RadUrlRef {
-            authority,
+            authority: &authority,
```
kim, line 319: Wait what? Why do we need to borrow here?

librad/src/net/gossip.rs
FintanH: Nope :shrug: Good catch. I should do a little scour myself to review these.

FintanH: Tried adding https://rust-lang.github.io/rust-clippy/master/#extra_unused_lifetimes to `lib.rs` and the command line and it seems like it won't pick it up. I wonder if it's a bug in clippy :thinking:

librad/src/uri.rs
FintanH: Whoops, we don't. My guess is the compilers magic with referencing and dereferencing handled it for us.

FintanH:
I did an audit of pass by reference. I think it should be in good shape now :+1:

kim (approved):
:+1:

xla (approved):
🗯 ❇️ 👻 🈂️

PR #368, opened 2020-10-14
Fix insufficiently polymorphic ref namespacing

kim:
The attempt to emulate existential typing led to the type of namespace
effectively being fixed to either `PhantomData<!>`, or `Hash`. This impedes a
smooth transition to the NG identities, where `Hash` is replaced by
(essentially) a OID.


librad/src/git/types.rs
```diff
@@ -108,69 +100,59 @@ impl<R> SymbolicRef<R> {
     ///     is passed.
     pub fn create<'a>(&self, repo: &'a git2::Repository) -> Result<git2::Reference<'a>, git2::Error>
     where
-        R: Display + Clone,
+        S: Display,
```
FintanH, line 103: Pretty weird that it can be _any_ `S` that implements `Display` but I guess this makes it easier to use ergonomically. Wondering if we should seal it with a `trait Ref: Sealed` or something?

librad/src/git/types/reference.rs
```diff
@@ -20,24 +20,23 @@ use std::{
...
-};
+use super::{namespace::AsNamespace, Force, Refspec, SymbolicRef};
 
 /// Type witness for a [`Reference`] that should point to a single reference.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub struct Single;
+pub struct One;
+
+/// Alias for [`One`].
+pub type Single = One;
```
FintanH, line 32: Do you think we need this or are you just planning on doing this change over in phases?

kim: It's the Rust version of unison :grimacing:

FintanH: lol, well can we stick with `One`? Makes more sense tbh

kim: Actually, the thing is that `name` is not well-defined -- what we want is either a `ext::RefLike` (if it is not a pattern), or a corresponding type which passes the validation rules with `GIT_REFERENCE_FORMAT_PATTERN`. So, the cardinality parameter would actually become obsolete (since we can't have the type of `name` _depend_ on it).

Then we could also define meaningful traits for things which can be in a refspec or symref -- `AsRefspec` boils down to just `Display` on both sides (which was the case before this patch). What we want is to restrict `Refspec` to `Local: Into<Either<RefLike, RefPattern>>, Remote: Into<Either<RefLike, RefPattern>>` and `SymbolicRef` to `S: Into<RefLike>, T: Into<RefLike>`.

xla (dismissed):
utACK

🎸 👜 🍮 🍉

librad/src/git/types/reference.rs
FintanH: Ya, that makes sense :ok_hand: Shall we capture that in a follow up?

FintanH (dismissed):
:metal:

kim:
I'm ok-happy with this now, as it does away with stringly manipulation of those pesky refnames and -patterns.

Admittedly, the `for<'a> &'a Into<RefLike>` are very shady business, and in fact may send `rustc`'s constraint solver on a journey of no return occasionally -- I would consider this a compiler bug. There is, however, no way to express (for example) the overlapping impl:

```rust
impl<S, T> From<SymbolicRef<S, T>> for SymbolicRef<RefLike, RefLike>
where
    S: Into<RefLike>
    T: Into<RefLike>
```

The only alternative I can see is to store trait objects `Box<dyn Into<RefLike>>` in structs, instead of parametric polymorphism. I'm not sure, however, if this is what we want.

FintanH (dismissed):
Amazing :clap: This is much better :raised_hands:

xla (approved):
🤖 🍂 🐲 🎞

FintanH (approved):
:metal:

kim (commented):
That menagerie... I wonder if I now fixed something which only incidentally
worked before

FintanH:
@kim want me to take a look? I'd probably just kill the test early with my sleeper hold and inspect the monorepo manually.

FintanH:
So I checked the monerepo state of the third peer and it doesn't seem to have received the `heads/master` of the first peer. What do you think you might have fixed? :sweat_smile:

librad/src/git/types.rs
```diff
@@ -224,28 +282,51 @@ impl Refspec<PeerId, PeerId> {
             {
                 let their_singed_rad_refs = rad_signed_refs_of(tracked_peer)?;
                 for (name, target) in their_singed_rad_refs.heads {
-                    let name_namespaced = format!("refs/namespaces/{}/{}", namespace, name);
-                    if let Some(name) = name.strip_prefix("refs/heads/") {
```
FintanH, line 228: We were stripping the `refs/heads` before but it doesn't look like we're doing that in the new changes. Could be that we're ending up with `refs/heads/refs/heads/foobar`

kim (commented):
It's when you can't type the same thing twice lol

librad/src/git/types.rs
kim: `name` is of type `OneLevel` now -- no more strippin

FintanH: All the clubs are shut down at this level, strippin is verboten

FintanH (dismissed):
This is a strip free zone :no_entry:

PR #369, opened 2020-10-15
Refine types representing refnames + refspec patterns

kim:
Requires a lower bound on `git2` >= 0.13.12.


librad/src/git/ext/reference/name.rs
```diff
@@ -0,0 +1,504 @@
...
+            .0
+            .to_str()
+            .expect("cannot be constructed from invalid utf8")
+    }
+}
+
+/// A [`RefLike`] without a "refs/" prefix.
+///
+/// Conversion functions strip the first **two** path components iff the path
+/// starts with `refs/`.
```
FintanH, line 158: Just to clarify, if it's taking the first two components does that mean it's stripping `refs/<next_level>`? For example, `refs/heads/mistress` becomes `mistress` and `refs/remotes/origin/of-life` becomes `origin/of-life`.

FintanH (approved):
Sick :metal: I presume this is also in prep for using in the reference API?

FintanH:
> I presume this is also in prep for using in the reference API?

Just got my answer by seeing commits on the other PR :sweat_smile:

librad/src/git/ext/reference/name.rs
FintanH: Ya, I mean it makes sense in that I realise git does this a lot. I think since we're working with the raw strings it's fine to do this. Maybe just calling it out in the docs with examples will make it more clear.

kim: Made you some :guide_dog: s

FintanH (approved):
:kissing::ok_hand:

librad/src/git/ext/reference/name.rs
FintanH: image

PR #371, opened 2020-10-16
Add event emission to seed library

cloudhead:
This will be used to create the seed dashboard: https://www.figma.com/file/SQadExhzMpNYFziggQ1vDh/seed-node-ui?node-id=0%3A1


seed/src/lib.rs
```diff
@@ -23,6 +23,7 @@ use std::{collections::HashSet, io, net::SocketAddr, path::PathBuf, time::Durati
 
 use futures::stream::StreamExt;
 use thiserror::Error;
+use tokio::sync::mpsc as chan;
```
kim, line 26: I know it's a lie, but I do not want any tokio in library code unless it is absolutely required. The only exception to this rule are `main` functions, integration tests, and currently `spawn` points.

kim (changes requested):

seed/src/lib.rs
cloudhead: Understandable, but how should we solve this then? I can't use `std::mpsc`, can I?

cloudhead: Otherwise I can use a trait, but that seems a bit overkill.

kim: `futures::channel::mpsc`

seed/src/main.rs
```diff
@@ -81,5 +81,9 @@ async fn main() {
     };
     let node = Node::new(config).unwrap();
 
-    node.run().await.unwrap();
+    let (tx, mut rx) = tokio::sync::mpsc::channel(1);
+
+    tokio::spawn(async move { while rx.recv().await.is_some() {} });
```
kim, line 86: :thinking:

cloudhead: Yeah I mean, I guess I could log the events for now. Perhaps this `main` should be moved to `examples/`

kim: > moved to examples

Sounds sensible.

What's the idea eventually? Slap some prometheuses?

cloudhead: Ok had a chat with xla, we're gonna get #340 merged, so this file will be removed anyway. I think `radicle-bins/seed` will serve as the example then.

cloudhead: Create a fancy dashboard: https://www.figma.com/file/SQadExhzMpNYFziggQ1vDh/seed-node-ui?node-id=0%3A1 (updated pr description)

seed/src/main.rs
```diff
@@ -81,5 +82,13 @@ async fn main() {
     };
     let node = Node::new(config).unwrap();
 
-    node.run().await.unwrap();
+    let (tx, mut rx) = futures::channel::mpsc::channel(1);
+
+    tokio::spawn(async move {
+        while let Some(event) = rx.next().await {
+            log::debug!("{:?}", event);
```
kim, line 89: You set up `tracing` already, no need to sneak in `log` again :+1:

seed/src/lib.rs
```diff
@@ -21,7 +21,7 @@ extern crate async_trait;
 
 use std::{collections::HashSet, io, net::SocketAddr, path::PathBuf, time::Duration, vec};
 
-use futures::stream::StreamExt;
+use futures::{channel::mpsc as chan, sink::SinkExt, stream::StreamExt};
```
xla, line 24:
```suggestion
use futures::{channel::mpsc as chan, sink::SinkExt as _, stream::StreamExt as _};
```

xla: Why the rename to `chan`? `mpsc::Sender<>` is way more expressive.

kim (approved):

seed/src/lib.rs
cloudhead: It makes it easy to change the channel provider, which tends to happen, as proven by this PR :wink:

xla: jackie chan

xla (approved):
🐟 😶 ↘️ 🐁

seed/src/lib.rs
kim: What's the deal about renaming trait-only imports to `_`, actually? Is that something the idiotmatic polizei mandates?

xla: From the bible: [Underscore Imports](https://doc.rust-lang.org/reference/items/use-declarations.html#underscore-imports)

cloudhead: I actually like it because it also tells you it's a trait import. I generally find trait imports confusing.

PR #373, opened 2020-10-16
Expose connected peers

### LINKED ISSUE #209: PeerApi: Read currently-connected PeerIds
Not having this was simply an oversight.

Note that this is both easier to implement, and to reason about than the events
proposed in #205. Perhaps we can agree to only implement this.

xla:
Closes #209

### DIFF
--- a/librad/src/net/protocol.rs
+++ b/librad/src/net/protocol.rs
@@ -279,6 +279,17 @@ where
         self.gossip.announce(have).await
     }
 
+    /// Mapping of currently connected [`PeerId`]s and their remote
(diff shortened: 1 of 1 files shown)


kim (approved):

cloudhead (approved):
I would usually prefer an iterator to be returned, but this is also fine.

FintanH:
> I would usually prefer an iterator to be returned, but this is also fine.

Iterator also seems better to me :eyes:

xla:
> Iterator also seems better to me eyes

Is a PITA with locks, will not attempt at this hour.

PR #374, opened 2020-10-19
Add a check that the certifier is fetched

FintanH:
This currently shows that it isn't working as understood but instead
panics at the call to `some_metadata`.

Continuing on the conversation from https://github.com/radicle-dev/radicle-upstream/pull/1072

This comment in being of particular interest: https://github.com/radicle-dev/radicle-upstream/pull/1072#issuecomment-711848479


FintanH:
So in this test case this line https://github.com/radicle-dev/radicle-link/blob/dc23136019e5885e7599441c90c7e8ed5747253b/librad/src/git/types.rs#L308 is saying there are no certifiers when I print to stdout.

FintanH:
I found why the set of peers was always empty. The following code is trying to parse the `Hash` into a `RadUrn` instead of constructing it.
https://github.com/radicle-dev/radicle-link/blob/master/librad/src/git/storage.rs#L1069-L1080

FintanH:
With that fixed, I'm looking at a refspec of the form:
```
refs/namespaces/hwd1yreroji6nd5n6otd8ku3qzyba9ud5344obc3wi46mjgy983459tko9h/refs/rad/id:refs/namespaces/hwd1yreroji6nd5n6otd8ku3qzyba9ud5344obc3wi46mjgy983459tko9h/refs/remotes/hyy4qcyud9wetojdmc9u7wg1u4njcnktabduyhtwsap61bijazxw7c/rad/id
```

which seems right to me. I would imagine that it fetches the ref `refs/namespaces/hwd1yreroji6nd5n6otd8ku3qzyba9ud5344obc3wi46mjgy983459tko9h/refs/rad/id` from the peer and this would end up in the fetching peer's monorepo as `refs/namespaces/hwd1yreroji6nd5n6otd8ku3qzyba9ud5344obc3wi46mjgy983459tko9h/refs/remotes/hyy4qcyud9wetojdmc9u7wg1u4njcnktabduyhtwsap61bijazxw7c/rad/id`, I'm not witnessing this though :thinking:

FintanH:
Hmmm so I confirmed there is a remote for the certifier URN, but what I'm thinking is that we're using the wrong remote to fetch it. I think when we fetch we use the remote for the URN that's the project.

FintanH:
To clarify my comment above, we set up the following refspecs when cloning:
https://github.com/radicle-dev/radicle-link/blob/master/librad/src/git/types.rs#L298-L306

You may notice that the namespace is different, i.e. it's the `<ceritifier>` namespace -- as opposed to `<namespace>` in any of the other refspecs.

Our remotes are based on specific `RadUrn`s so the remote used for the project based on `<namespace>`, mentioned above, won't pick up the references under the `RadUrn` for `<certifier>`.

So we need a way to switch to the `<certifier>` remote and fetch the refspecs using it. My solution here was to call `certifiers` and `fetch_repo` for each `RadUrn`. @kim do you have any other ideas because this seems quite brute force? I also don't think the code linked above in this comment is even used at all, so it might be the case that we can remove it.

kim:
You're on the wrong track here, I'm afraid: there is no different remote, the p2p transport is crafted carefully such that top-level namespaces can be fetched from the remote end in **one** fetch operation. The issue is that, for the `can_clone` test, the `certifiers_of` function returns 0 (zero) certifiers. Hence that refspec is not used.

FintanH:
I'm confused when I was inspecting `certifiers_of` down in `fetch_heads`, calling the `can_clone`, I did see a certifier `refs/namespaces/hwd1yreroji6nd5n6otd8ku3qzyba9ud5344obc3wi46mjgy983459tko9h/refs/rad/id`. 

The reason I thought it was the fetcher is because the result of `fetch_heads` is then used in the `fetcher` built for the `clone_repo` based on the project's URN.

librad/src/git/storage.rs
```diff
@@ -747,6 +747,8 @@ where
             .copied()
             .collect::<HashSet<PeerId>>();
 
+        // FIXME(haxpenny): wtf is a resolver... we need to verify the certifier
+        // delegations that are a retrieved from this fetch.
```
kim, line 751: Do we not want to ensure `refs/namespaces/<certifier>/rad/{id,ids/*}` exists, even though we didn't actually verify?

FintanH: Yup, strong point well made :metal:

FintanH: Although, wait a minute. It's not `refs/namespaces/<certifier>/rad/{id,ids/*} ` but instead `refs/namespaces/<certifier>/refs/remotes/<tracked_peer>/rad/{id,ids/*} ` because the refspec in `fetch_heads` doesn't map it to a top level.

Which I suppose begs the question, does it make sense to add a fetch so that we have `refs/namespaces/<certifier>/rad/{id,ids/*}`? I guess I don't see why not.

FintanH: Actually, came to see a reason why not. We would have to come to a decision on which is the "canonical" version if we have multiple peers under the same certifier. So we probably shouldn't make this arbitrary choice and instead just have them under each remote.

And so the validation for existence is for `refs/namespaces/<certifier>/refs/remotes/<tracked_peer>/rad/{id,ids/*}`

xla: My mental model would match a layout similar to how projects work as well where potentially different versions of identity materialise under a remote of a tracked peer. When viewing such an identity, i.e. another users profile in app, one would select which peers view to look at similar again to how projects are interacted with.

kim: Yes. It may at this point that we either _already have_ `refs/namespaces/<xxx>/rad/id` (because we cloned that independently), or that we don't. In the former case, what the code currently does is correct (merging updates is an interactive process), whereas the latter would require proper verification. Since we don't yet have that, we can just pretend for now that the remote view is fine, and adopt this as our own view.

FintanH: So if an "Entity" is signed by `Signatory::OwnedKey` it doesn't have any `rad/ids/*` :facepalm: 

So my thinking to fulfil this is: I can verify that `refs/namespaces/<certifier>/refs/remotes/<remote_peer>/rad/id` exists and adopt that as our own. Commit incoming.

xla: > So my thinking to fulfil this is: I can verify that `refs/namespaces/<certifier>/refs/remotes/<remote_peer>/rad/id` exists and adopt that as our own. Commit incoming.

What is meant by adopting as our own here?

xla: > So if an "Entity" is signed by `Signatory::OwnedKey` it doesn't have any `rad/ids/*`

Should that be addressed?

FintanH: > What is meant by adopting as our own here?

We create a `symref` of `refs/namespaces/<certifier>/refs/remotes/<remote_peer>/rad/id` under `refs/namespaces/<certifier>/refs/rad/id` where `remote_peer` is the peer we're cloning from. This is the same behaviour as with projects.

> Should that be addressed?

I'm not sure it's worth it because Entities won't be around much longer and I don't think this necessarily breaks things.

librad/src/git/storage.rs
```diff
@@ -755,6 +755,32 @@ where
...
+                if !self.has_ref(&remote_certifier)? {
+                    return Err(Error::MissingCertifier { certifier, urn });
+                }
+
+                let local_certifier = remote_certifier.clone().with_remote(None);
+
+                // We may have already cloned this certifier at some point
+                if !self.has_ref(&local_certifier)? {
+                    remote_certifier
+                        .symbolic_ref(local_certifier, Force::False)
```
kim, line 777: If I didn't get `SymbolicRef` backwards, this will create:

    r/nss/<certifier>/refs/rad/id -> r/nss/<project>/refs/remotes/<remote peer>/rad/ids/<certifier>

Below we're creating the symref in the opposite direction (but from our own namespace):

    r/nss/<project>/refs/rad/ids/<certifier> -> r/nss/<certifier>/refs/rad/id

Because we want `<project>` to always point to the latest of `<certifier>`.

Now, creating the first symref is not technically _wrong_, because fetching `<certifier>` will replace it with a direct ref when there are updates. It may be a bit confusing, though.

librad/src/git/storage.rs
```diff
@@ -755,6 +755,32 @@ where
             |peer| self.certifiers_of(&urn, &peer),
         )?;
 
+        // Ensure refs/namespaces/<certifier>/remotes/<tracked>/rad/id exists
+        // and adopt it as our own.
+        //
+        // FIXME(haxpenny): Currently rad/ids/* doesn't exist if something is signed by
```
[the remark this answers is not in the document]
FintanH: Ah ok, so will a `user`-typed identity ever have `rad/ids/*`? :thinking: I suppose it _could_ if the person creates multiple identities that vouch for each other?

If they are present, should we iterate through them and create symrefs in the same way we're now doing with `rad/id`?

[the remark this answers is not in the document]
kim: > will a user-typed identity ever have rad/ids/*

Nope

> if the person creates multiple identities that vouch for each other

Nooooooooo....! This is why I don't want to continue using "certifiers" -- there's no such thing as certifying. The terminology suggests that some random person can come along and "certify" that this identity is valid, or trusted, or whatever. We are not doing that at all, neither is there a web-of-trust, nor a certificate authority.

librad/src/git/storage.rs
FintanH: Well, I'm glad you brought this up because that code is also `parse`'ing the wrong thing :sweat_smile: It's trying to get a `Hash` from `refs/namespaces/hwd1yre8itytmdeqt8ezb1f5sbtcz1p8z3cqy66nbyh73yh9idqc43k3ujy/refs/remotes/hydneauugdxtsrpg5o636gizgmoxoo4ymbkk74am5qf1uxkastoydo/rad/ids/hwd1yregdu7n8rqpsmft43ut9k3b1h9j4bxpaaw9whzk8sua89zn4zqgruc`

librad/src/git/storage.rs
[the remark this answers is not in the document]
FintanH: Ah, sorry :sweat: There's a lot to track here and I'm with you now :sweat_smile: I'll remove the FIXME

xla (approved):
💳 🌓 🔬 👄

librad/src/git/storage.rs
kim: I'm a little bit lost now -- don't we need to ensure that the `rad/id` (that is, not `remotes/xx/rad/id`) of the top-level namespace exists before symlinking to it?

FintanH: So I believe it goes like this (where `<namespace>` is what we're cloning/fetching and `<remote_peer>` is the peer we're cloning/fetching from):
* [For each tracked `<peer>` and for each `<certifier>` under that tracked peer we fetch](https://github.com/radicle-dev/radicle-link/blob/fintan/top-level-certifier/librad/src/git/types.rs#L298-L306)
  * From: `refs/namespaces/<certifier>/refs[/remotes/<peer>]/rad/id`
  * Into: `refs/namespaces/<certifier>/refs/remotes/<peer>/rad/id`
* [Then we iterate over `refs/namespaces/<namespace>/refs/remotes/<remote_peer>/rad/ids/*` which gives us these identities](https://github.com/radicle-dev/radicle-link/blob/fintan/top-level-certifier/librad/src/git/storage.rs#L762-L770)
* [For each `<certifier>` we create a symref](https://github.com/radicle-dev/radicle-link/blob/fintan/top-level-certifier/librad/src/git/storage.rs#L779-L786):
  * New symbolic ref: `refs/namespaces/<certifier>/rad/id`
  * Points to: `refs/namespaces/<namespace>/refs/remotes/<remote_peer>/rad/ids/<certifier>`

Writing this out, I get the feeling I'm mixing this last step up. Should it be that we create the `refs/namespaces/<certifier>/rad/id` from the `refs/namespaces/<certifier>/refs/remotes/<remote_peer>/rad/id`? I guess what's confusing me is that [original block of code that wasn't firing](https://github.com/radicle-dev/radicle-link/blob/fintan/top-level-certifier/librad/src/git/storage.rs#L758-L790) because of the parsing issue.

FintanH: This seems arse-ways now that I think about. The `rad/ids/*` are always just symref'd, correct? 

So, if I'm getting this right, the original [block here](https://github.com/radicle-dev/radicle-link/blob/master/librad/src/git/storage.rs#L762-L795) was creating a symref to the `refs/namespaces/<certifier>/rad/id`. The issue I'm having here is that we don't actually have that reference, instead we have `refs/namespaces/<certifier>/refs/remotes/<peer>/rad/id`.

Sorry that this is taking so long to sink in for me, it's pretty frustrating :disappointed:

kim: Ok, so in the last step, if we **already have** `refs/namespaces/<certifier>/rad/id`, what we want is symrefs

    refs/namespaces/<current project>/rad/ids/<certifier> -> refs/namespaces/<certifier>/rad/id

This ensures that forall `<project>`s, we get the updated tips by fetching `refs/namespaces/<certifier>` automatically. Note that we're not touching any `refs/remotes/`!

If we **do not have** `refs/namespaces/<certifier>/rad/id`, we **should** run the verification for `refs/namespaces/<certifier>/refs/remotes/<remote peer>/rad/id`, and **iff** that checks out, create a **direct** ref `refs/namespaces/<certifier>/rad/id`. Because there is no proper verification, we just pretend it's valid and create that direct ref.

FintanH: 7th times a charm :pray:

kim (approved):
I _think_ this is good now

 _mumble mumble algebras mumble mumble_

FintanH:
Honestly would love some algebraic understanding here because I'm staring at refs and monorepos for a good half hour before I get the right incantation 🤣😭

xla (approved):
🏖 🔊 🖋 🚾

PR #375, opened 2020-10-19
Local transport: re-instantiate threaded wait

kim:
We observe that spawned `git-receive-pack` and/or `git-upload-pack`
processes may outlive the `stdio` handles owned by the `libgit2` custom
transport. Killing the child in a drop finaliser is, however, premature
-- what we would want is to kill it after some timeout has elapsed.

Unfortunately, Rust makes it impossible to cancel OS threads or
processes in a non-blocking way -- unless non-blocking IO is used all
the way down. We can make use of this in the p2p transport, because a.
it is used in a long-running event loop, and b. only `git-upload-pack`
is used there, which allows to set an upper bound on the child running
time (via `--timeout`). Not so for the local transport, which is always
used in a blocking context.

To work around the problem, we spawn the `wait` call onto a separate
thread, and maintain a global counter of such waiters. If a maximum
number of blocked waiters is exceeded, no more `libgit2` transports can
be created. This is to defend against the pathological case of child
processes not terminating for some reason, and so the program leaking
waiter threads.

The tradeoff is that short-lived processes may terminate too early, and
thus terminating the child before it has finished.

### DIFF
--- a/librad/src/git/local/transport.rs
+++ b/librad/src/git/local/transport.rs
@@ -18,9 +18,16 @@
 use std::{
     collections::HashMap,
     io::{self, Cursor, Read, Write},
+    panic::{self, UnwindSafe},
     path::PathBuf,
     process::{Child, ChildStdin, ChildStdout, Command, ExitStatus, Stdio},
-    sync::{Arc, Mutex, Once, RwLock},
+    sync::{
+        atomic::{AtomicUsize, Ordering},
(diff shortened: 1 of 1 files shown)


FintanH:
> The tradeoff is that short-lived processes may terminate too early, and
> thus terminating the child before it has finished.

What kind of actions would surface this? Would they be rare cases?

PR #377, opened 2020-10-19
Local transport: side-channel for blocking on local transport results

kim:
`libgit2` may drop all references to the local transport before the
child process exited. We must, however, `wait(2)` on the children on a
(non-cancellable) thread, which leads to the dilemma of either risking
to block the parent process (by joining on this thread), or exiting it
prematurely.

We thus establish a side-band channel which allows non-longrunning
processes to fully await completion of their local transport operations
(e.g. in a `Drop` finaliser).

---

Supersedes #375

Note: one might think that it would be possible to simply use a `std::sync::mpsc`
channel for this, but that is not the case due to caught panics not being `Sync`.

### DIFF
--- a/librad/src/git/local/transport.rs
+++ b/librad/src/git/local/transport.rs
@@ -16,12 +16,14 @@
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
 use std::{
-    collections::HashMap,
+    collections::{HashMap, VecDeque},
     io::{self, Cursor, Read, Write},
+    panic::{self, UnwindSafe},
     path::PathBuf,
     process::{Child, ChildStdin, ChildStdout, Command, ExitStatus, Stdio},
-    sync::{Arc, Mutex, Once, RwLock},
+    sync::{Arc, Condvar, Mutex, Once, RwLock},
     thread,
+    time::Duration,
 };
 
 use git2::transport::{Service, SmartSubtransport, SmartSubtransportStream, Transport};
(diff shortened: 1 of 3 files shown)


librad/tests/working_copy.rs
```diff
@@ -204,6 +208,13 @@ async fn can_fetch() {
                 remote.fetch(&[&name], None, None).unwrap();
             }
         }
+        for res in local_transport_results_peer2
```
FintanH, line 211: So this list of results will be an exact correspondence to the `remote.fetch` actions above? Which means that like parentheses balancing, we need to ensure that every `git2` action has a corresponding `wait`?

FintanH (approved):
Well, _this_ is terrifying me.

On one hand, I'm glad you found a solution. On the other, I'm terrified of how I'm going to shoot myself in the foot by not remembering to call `wait`. Luckily, I believe there should be only a few places where we need to do this.

librad/tests/working_copy.rs
kim: No unfortunately not -- `fetch` is a compound action:

> Convenience function to connect to a remote, download the data, disconnect and update the remote-tracking branches.

One `fetch` seems to correspond to three calls to `SmartSubtransport::action` -- which doesn't make sense, as one would expect two, but :man_shrugging:

kim:
>  how I'm going to shoot myself in the foot

Well, it's not that you _must_ `wait` -- that's what the spawned threads are for. If you remember, this whole chore started because some tests were flaky -- we had a wait thread in place, but the assertions got called before the thread finished. So, if you need to ensure that you get a consistent read-after write, you need to wait, otherwise it's enough to implant a `Drop` somewhere which attempts to drain this queue for a while before the process gets shutdown (gracefully).

I considered "unregistering" the transport for the current `PeerId` when `Results` gets dropped -- but then realised that this would be racey: one could totally call `LocalTransportFactory::configure` _before_ the finaliser ran, and then end up with an error. If you can think of a way to wrap this whole dance register, configure, wait-results-on-drop in a datatype which `Deref`s to `&git2::Remote`, that would be great. I think that it would need to key the `SETTINGS` map differently (as in, a random key) to make it thread-safe.

librad/src/git/local/transport.rs
```diff
@@ -16,12 +16,14 @@
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
 use std::{
-    collections::HashMap,
+    collections::{HashMap, VecDeque},
```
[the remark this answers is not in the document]
kim: I always think of Eurocheque :writing_hand:

librad/tests/working_copy.rs
FintanH: Even more terrifying :scream:

xla (approved):
utAck

👁 🚕 👹 🕴

PR #379, opened 2020-10-20
Allow untrack of non-existing remote

FintanH:
If we call untrack on a non-existing remote we get an error back from
git2 saying that it doesn't exist. It would make more sense to match on
the remote not existing and return a bool whether the remote used to
exist or not.


kim (approved):

xla (approved):
🤖 🍬 🎱 👑

PR #383, opened 2020-10-22
Restrict the types which can be used as a remote for `Reference`

kim:
This should prevent the typechecker from recursing, at the expense of a loss of generality.

### DIFF
--- a/librad/src/git/sealed.rs
+++ b/librad/src/git/sealed.rs
@@ -15,7 +15,12 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+use crate::peer::PeerId;
+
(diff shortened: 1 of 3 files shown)


FintanH (approved):
Out of interest, did you check this worked? 

Also, admittedly I like the fact that this is `AsRemote` because I got confused when I was looking at the docs and couldn't remember what the `R` was :sweat_smile:

kim:
> Out of interest, did you check this worked?

Ya, uhm, can you remember a repro? The weird thing about this recursion was that it didn't _always_ trigger.

FintanH:
One easy way for me was to forget to import `TryFrom`. I think another way I triggered it was giving `FlatRef` the wrong annotation, for example, `let foo: RefLike<String, _>`

PR #384, opened 2020-10-22
Check certifier existence

FintanH:
It could be the case that we already have the certifier -- for example,
we have cloned the project but now we are fetching. In this case we
should check if it exists before creating the symbolic reference,
otherwise it's a no-op.

### DIFF
--- a/librad/src/git/storage.rs
+++ b/librad/src/git/storage.rs
@@ -787,11 +787,15 @@ where
                             self.adopt_as_ours(&remote_certifier, &certifier_id)?;
                         }
 
-                        certifier_id
-                            .symbolic_ref(certifier_here, Force::False)
-                            .create(&self.backend)
-                            .map_err(Error::from)
-                            .and(Ok(()))
+                        if !self.has_ref(&certifier_here)? {
+                            certifier_id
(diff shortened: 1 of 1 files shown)


librad/src/git/storage.rs
```diff
@@ -787,11 +787,15 @@ where
                             self.adopt_as_ours(&remote_certifier, &certifier_id)?;
                         }
 
-                        certifier_id
-                            .symbolic_ref(certifier_here, Force::False)
-                            .create(&self.backend)
-                            .map_err(Error::from)
-                            .and(Ok(()))
+                        if !self.has_ref(&certifier_here)? {
```
FintanH, line 790: This _seems_ right, but I'm now second-guessing my knowledge of monorepo :grimacing:

kim:
`libgit2` uses [locking](https://github.com/libgit2/libgit2/blob/master/src/refdb_fs.c#L1228) to make the operation atomic (across processes / repository handles). So I'd prefer my solution, as something else could attempt to symref between `has_ref` and `create`.

FintanH:
Appreciate that you looked into that :bow:

PR #385, opened 2020-10-22
Fix: ignore errors if certifier symref already exists

kim:
Somewhat counter-intuitively, `Force::False` will still return an error
if the symbolic reference already exists. We ignore this error, and
proceed attempting to symref the remaining certifier ids.

### DIFF
--- a/librad/src/git/storage.rs
+++ b/librad/src/git/storage.rs
@@ -790,8 +790,8 @@ where
                         certifier_id
                             .symbolic_ref(certifier_here, Force::False)
                             .create(&self.backend)
-                            .map_err(Error::from)
                             .and(Ok(()))
+                            .or_matches::<Error, _, _>(is_exists_err, || Ok(()))
                     },
                 }
             })?;


librad/src/git/storage.rs
```diff
@@ -790,8 +790,8 @@ where
                         certifier_id
                             .symbolic_ref(certifier_here, Force::False)
                             .create(&self.backend)
-                            .map_err(Error::from)
                             .and(Ok(()))
+                            .or_matches::<Error, _, _>(is_exists_err, || Ok(()))
```
FintanH, line 794: Could we move this into the `create`? :thinking:

kim: I'm unsure if this would be correct in the general case. Suppose you have:

    A -> B

Now, you want to create:

    A -> C

Without applying force, I'd expect this to also return `GIT_EEXISTS`. Here, we "know" that we'd always try `A -> B`, so the operation is idempotent, but the semantics of the above are different methinks.

FintanH (approved):
:metal:

PR #387, opened 2020-10-22
Seed handle

cloudhead:
Allows external code to ask the seed for projects and peers.


seed/src/lib.rs
```diff
@@ -206,21 +206,60 @@ impl Default for NodeConfig {
     }
 }
 
+pub struct NodeHandle {
+    channel: chan::UnboundedSender<Request>,
+}
+
+impl NodeHandle {
+    pub async fn get_projects(&mut self) -> Result<Vec<()>, chan::SendError> {
```
xla, line 214: The one thing I'd change about the proposed API is to not leak this `SendError` as it defeats the purpose of hiding channel semantics from consumers.

cloudhead: That's a :ok_hand: point

cloudhead: 03fa6bd

FintanH (commented):
I'm finding it a bit hard to review this PR. It seems like there's a lot of different types and methods tossed in `lib.rs` at this stage. Could we start creating sub-modules to better organise this?

FintanH (approved):
SO much easier! LGTM :clap:

seed/src/event.rs
```diff
@@ -0,0 +1,61 @@
...
+        name: Option<String>,
+    },
+    /// A peer has disconnected.
+    PeerDisconnected(PeerId),
+    /// A project has been tracked from a peer.
+    ProjectTracked(Project, PeerId),
+}
+
+impl Event {
+    pub(crate) async fn peer_connected(
```
xla, line 25: What's the reason for the excessive use of crate visibility? Why not instead control the re-exports?

xla (approved):
🕌 ⤵️ 📞 👈

seed/src/event.rs
cloudhead: Well the thing is I need to export `Event`, but having this as simply `pub` will allow users to call `Event::peer_connected` afaict.

PR #388, opened 2020-10-23
Refname macros

kim:
I got a bit fed up with the `try_from(&str).unwrap()` dance, so went ahead with this. For reasons, `proc-macro`s must go into their own crate, so it required shuffling some stuff around in order to not create cyclic dependencies.


macros/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+
+use radicle_git_ext::reference::name::{RefLike, RefspecPattern};
+
+/// Create `RefLike` from a string literal.
+///
+/// The string is validated at compile time, and an unsafe conversion is
+/// emitted.
+///
+/// ```rust
+/// use std::convert::TryFrom;
```
kim, line 34: Not sure if there is a way to avoid having to import `TryFrom` at the use site -- it needs to be in scope at runtime. Maybe just provide `RefLike::unsafe_from(s: &str)`?

macros/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+            TokenStream::from(expand)
+        },
+
+        Err(e) => {
+            lit.span()
+                .unwrap()
+                .error(format!("invalid RefLike literal: {}", e))
+                .emit();
+
+            TokenStream::from(quote! { unimplemented!() })
```
kim, line 59: This is a bit funky -- `rustc` won't stop reporting errors after we emitted ours. If we just emit an empty `TokenStream`, there will be the unhelpful message "expected expression, found end of macro arguments". This way, a series of unused import warnings, and perhaps a type inference error will follow.

FintanH: Could we use `std::compile_error` instead possibly?

kim: That's worse -- the span (the wobbly line indicating where the malformed input was) we just emitted will be lost, and instead we'll get a wobbly line below `compile_error`, which takes only a `&'static str`.

FintanH: Fair enough, it was a shot in the dark :sweat_smile: I'll take a closer look now :monocle_face:

xla (approved):
📪 🎩 ⬅️ 🛩

macros/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+#![feature(proc_macro_diagnostic)]
+
+use std::convert::TryFrom;
```
FintanH, line 20: Reading through https://gist.github.com/Koxiaet/8c05ebd4e0e9347eb05f265dfb7252e1#use-unambiguous-paths I can't tell if these imports should be imported with the `::` prefix. What do you think?

FintanH (commented):
Do we care about `no_std`? https://gist.github.com/Koxiaet/8c05ebd4e0e9347eb05f265dfb7252e1#re-export-core-and-alloc

Do you think it'd be helpful to add a hygiene test like here? https://gist.github.com/Koxiaet/8c05ebd4e0e9347eb05f265dfb7252e1#test-your-macro-with-no_implicit_prelude

FintanH (approved):
I'm happy otherwise :)

macros/src/lib.rs
kim: No you misread that: the _generated_ code should use unambiguous paths.

kim:
> Do we care about no_std

No.

> hygiene test like here

This would be more ritual than anything at this point imho -- we have all fully-qualified paths, it's like 3 lines of code **and** we use those macros in `librad`, which will complain soon enough.

FintanH:
Sweet, go go go!

PR #389, opened 2020-10-23
Macros: transmute into target type

kim:
Turns out we can actually avoid the runtime conversion (and allocation).
The only potential drawback is that we emit `unsafe` code, which linters
may dislike.


FintanH (approved):
Slick. Level :up:

kim:
I will be done only once I have an actual use for one of those `unsafe { &*(x as *const _ as *const _) }` :smiling_imp:

PR #390, opened 2020-10-23
Re-export git-ext, std-ext and macro crates through librad

kim:
Otherwise this is all going to end in **cargo hell**.

Note about the aliasing: I know people have **opinions** about aliasing, but **my opinion** is that this particular aliasing the most **gentle on the eyes**.


FintanH (approved):
No opinions from me :innocent:

PR #391, opened 2020-10-23
protocol: handle disconnect when conn is done

xla:
Adds explicit disconnect handling in the two places where they seemed to
lack so far. This should make it possible for a protocol user to observe
all disconnections, not only the ones initiated by the gossip layer.
Might be worthwhile at a later stage to bubble up the connect and
disconnect control events from gossip. Also aligned the variant names.

### DIFF
--- a/librad/src/net/protocol.rs
+++ b/librad/src/net/protocol.rs
@@ -61,7 +61,7 @@ use crate::{
 #[derive(Clone, Debug)]
 pub enum ProtocolEvent<A> {
     Connected(PeerId),
-    Disconnecting(PeerId),
+    Disconnected(PeerId),
     Listening(SocketAddr),
     Gossip(gossip::Info<IpAddr, A>),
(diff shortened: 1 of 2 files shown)


cloudhead (commented):
:parasol_on_ground: :parasol_on_ground: :+1:

librad/src/net/protocol.rs
```diff
@@ -61,7 +61,7 @@ use crate::{
 #[derive(Clone, Debug)]
 pub enum ProtocolEvent<A> {
     Connected(PeerId),
-    Disconnecting(PeerId),
+    Disconnected(PeerId),
```
kim, line 64: Nit: I'd leave it as `Disconnecting` (progressive form), because that is what is happening -- the disconnect is only complete once all references to the connection are dropped. This might not be the case when the event is emitted -- f.ex. there could still be a `git` stream in-flight.

kim (approved):

cloudhead (approved):

PR #394, opened 2020-10-28
seed: fetch and track when existing

xla:
Ensure replication of project updates coming from different peers the the one first observed.

### DIFF
--- a/seed/src/lib.rs
+++ b/seed/src/lib.rs
@@ -42,7 +42,7 @@ use librad::{
     },
     paths,
     peer::PeerId,
-    uri::RadUrn,
+    uri::{self, RadUrn},
 };
 
 pub use crate::{
@@ -270,6 +270,10 @@ impl Node {
     ) -> Result<(), Error> {
         let peer_id = peer_info.peer_id;
         let url = urn.clone().into_rad_url(peer_id);
+        let proejct_urn = RadUrn {
+            path: uri::Path::new(),
+            ..urn.clone()
+        };
         let port = peer_info.advertised_info.listen_port;
         let addr_hints = peer_info
             .seen_addrs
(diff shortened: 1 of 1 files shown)


seed/src/lib.rs
```diff
@@ -270,30 +270,50 @@ impl Node {
     ) -> Result<(), Error> {
         let peer_id = peer_info.peer_id;
         let url = urn.clone().into_rad_url(peer_id);
+        let proejct_urn = RadUrn {
```
kim, line 273: I, for one, support this spelling (localised alternatives: prøjct, pröjct).

seed/src/lib.rs
```diff
@@ -270,30 +270,50 @@ impl Node {
...
+        };
         let port = peer_info.advertised_info.listen_port;
         let addr_hints = peer_info
             .seen_addrs
             .iter()
             .map(|a: &std::net::IpAddr| (*a, port).into())
             .collect::<Vec<_>>();
 
-        let result = {
+        // FIXME(xla): There should be a saner way to test.
```
kim, line 284: I hear you, and am in the process of removing the distinction between `clone` and `fetch`.

That being said, you should be able to condense the code to track (unconditionally), then either fetch or clone depending on the result of `has_urn`.

seed/src/lib.rs
```diff
@@ -270,30 +270,50 @@ impl Node {
...
+        let exists = {
+            let urn = proejct_urn.clone();
+            api.with_storage(move |storage| storage.has_urn(&urn))
+                .await??
+        };
+
```
cloudhead, line 290: Shouldn't we try to clone, and if that fails, try to fetch? I'm wondering if we get in a situation where the pre-condition changes in between calls to `with_storage`. Eg. it exists, then it's deleted, then we try to fetch. Or are we always assuming there is only one writer to the monorepo on the device where this runs?

xla: What would delete the project? And changing the order won't help if the precondition changes in between.

cloudhead: > What would delete the project?

That's my question, but for eg. some other radicle process running, eg. upstream etc.

> And changing the order won't help if the precondition changes in between.

That's not what I'm suggesting, I'm suggesting getting rid of the `exist` check so that there's no race condition possible.

seed/src/lib.rs
```diff
@@ -270,30 +270,51 @@ impl Node {
...
+        {
             let urn = urn.clone();
+            api.with_storage(move |storage| storage.track(&urn, &peer_id))
+                .await??;
+        }
+
+        // FIXME(xla): There should be a saner way to test.
+        let exists = {
+            let urn = project_urn.clone();
+            api.with_storage(move |storage| storage.has_urn(&urn))
```
FintanH, line 294: This `with_storage` can do combined with the two below, i.e.

```rust
with_storage(move |storage| {
  let exists = storage.har_urn(&urn)?;
  if exists { clone } else { fetch }
})

seed/src/lib.rs
xla: > That's not what I'm suggesting, I'm suggesting getting rid of the `exist` check so that there's no race condition possible.

The race condition is still possible, if we fetch after a clone has failed and the project goes away in between the clone and fetch, it still produces an error.

xla: > That's my question, but for eg. some other radicle process running, eg. upstream etc.

A seed and an upstream would not share the same state.

FintanH (approved):
:metal:

cloudhead (approved):

PR #395, opened 2020-10-28
protocol: fix handle disconnect for incoming

### LINKED ISSUE #393: `Disconnecting` event not firing
I'm still not receiving `Disconnecting` events reliably after a connection times out, even after #391 

Wondering if it has to do with incoming vs. outgoing connections, since it seems to work for the proxy quite well.

xla:
The try operator in the loop would return the outter method, preventing
that the `handle_disconenct`, necessary to fire the disconnecting event,
is ever reached.

Closes #393

### DIFF
--- a/librad/src/net/protocol.rs
+++ b/librad/src/net/protocol.rs
@@ -495,7 +495,7 @@ where
     async fn handle_incoming<Incoming>(
         &self,
         conn: quic::Connection,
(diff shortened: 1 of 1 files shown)


FintanH (approved):
:metal: :electric_plug:

cloudhead (approved):

PR #396, opened 2020-10-28
seed: Distinguish project updates

cloudhead:


seed/src/event.rs
```diff
@@ -36,6 +41,15 @@ pub enum Event {
     PeerDisconnected(PeerId),
     /// A project has been tracked from a peer.
     ProjectTracked(Project, PeerId),
+    /// A project has been updated from a peer.
+    ProjectUpdated {
+        /// The URN of the updated project, including the updated path.
+        urn: RadUrn,
+        /// The updated rev, if any.
+        rev: Option<Rev>,
```
cloudhead, line 49: Not sure in what cases this is `None` :thinking:

xla (approved):
🆔 💉 ❕ 🐻

seed/src/event.rs
FintanH: If it's `None` then it's basically just advertising the project. I'm not even sure what the protocol would do in that case. It'll possibly just mark it as uninteresting

seed/src/event.rs
```diff
@@ -53,6 +67,18 @@ impl Event {
         })
     }
 
+    pub(crate) async fn project_updated(
```
FintanH, line 70: Why is this `async`? Why is this `Result`? WHY ARE YOU LYING?!

seed/src/lib.rs
```diff
@@ -291,24 +294,31 @@ impl Node {
 
         let result = {
             let urn = project_urn.clone();
-            api.with_storage(move |storage| -> Result<(), librad::git::storage::Error> {
-                // FIXME(xla): There should be a saner way to test.
-                let exists = storage.has_urn(&urn)?;
+            api.with_storage(
+                move |storage| -> Result<bool, librad::git::storage::Error> {
```
FintanH, line 298: Oooh I didn't know about this syntax :o

kim:
Stale?

xla:
> Stale?

Yeah, not relevant anymore afaiu. We can reopen if we need it.

PR #400, opened 2020-11-02
Qualify path in urn_context

FintanH:
If we don't have a `refs/` prefix when calculating the urn_context, then
we assume that the reference name is expecting a `heads/` prefix.


kim (approved):

xla (approved):
☕️ 🖲 🛥 ⏫

PR #403, opened 2020-11-04
Next

kim:
[Making Of (until point of no return)](https://github.com/radicle-dev/radicle-link/compare/identities/integration)


kim:
Instead of merging this into `master` straight away, I'm inclined to set `next` as the default `HEAD` for this repo. New patches would go against `next`, but may require a backport against `master`.

FintanH:
> Instead of merging this into master straight away, I'm inclined to set next as the default HEAD for this repo. New patches would go against next, but may require a backport against master.

_cough_ Did someone say Pijul _cough_... Seems reasonable to me. We shouldn't be in this state of affairs too long would we?

kim:
Idk, depends on how much work it is to upgrade `upstream`. I bet people have it rather conservative and broken, than bleeding edge (and possibly broken).

kim:
Alright, so `git` is failing me on trying to rebase this onto `master`, so I'll need to squash it into one patch.

Since more changes are upcoming which should go against `next`, I propose the following old-school branching model:

* `next` becomes the new default branch
* `master` is thought of as the "stable" version. It stays as-is for now
* `maint` is forked off `master` in case there are urgent fixes/backports which would not apply against `next`

@FintanH @xla Thoughts?

FintanH:
Ya, sounds good :+1: I'm hoping https://github.com/radicle-dev/radicle-link/pull/401 is the last fix we need for the beta anyway. And we can focus on integrating after that. Sound good to you?

xla:
> @FintanH @xla Thoughts?

Sound strategy.

> Ya, sounds good +1 I'm hoping #401 is the last fix we need for the beta anyway. And we can focus on integrating after that. Sound good to you?

Hahaaha.

kim:
Alright, `next` is now the default branch. Changed merge bases of open PRs to `maint` where applicable.

The full history of this PR is preserved under `historical/making-of-next`

PR #404, opened 2020-11-06
Expose a verify function on Signed

FintanH:


kim (commented):
Not opposed, but: it's not possible to construct an unverified `Signed` -- what's the motivation for this change? Maybe you want to split the types into `RefsWithUnverifiedSignature` and `SignedAndVerified` or something?

FintanH:
> what's the motivation for this change?

The motivation was to build the [sign/verify tool](https://github.com/radicle-dev/radicle-bins/pull/24) when debugging.

> it's not possible to construct an unverified Signed

We can construct `Refs` via `Deserialize`, which is exactly what I needed in the tool :)

> Maybe you want to split the types into RefsWithUnverifiedSignature and SignedAndVerified or something?

I'm not sure I understand this. Isn't that basically what we have now? :thinking:

kim:
Well yes, `Signed` has `Deserialize`, so "it's not possible" is only true if you don't side-step `Signed::from_json`. If we do allow this, we no longer need `Signed::from_json`, but we can also not distinguish between "well-formed" and "has valid signature". What we would probably want is a `Deserialize` impl for `Signed<Unknown>`, a `Serialize` impl for `Signed<Valid>`, and perhaps a method `impl<T> Signed<T> { fn verify(&self) -> Option<Signed<Valid>> }`

FintanH:
Ah right, I see what you're saying. When in doubt, slap a type witness on it.

FintanH:
I left the `from_json` because it seems like it's still useful for composing the deserialization and verification. Lemme know whatya think :)

kim (approved):
:+1:

PR #406, opened 2020-11-09
Use BTreeSet instead of HashSet

FintanH:
Using a HashSet the iter method will return items in an arbitrary order.
Since this is what serde uses under the hood to turn HashSet into an
array, it destabilises the signing and verification of refs.

Using a BTreeSet gives us a guaranteed order for the array, meaning
consistent de/serialization across peers.

### DIFF
--- a/librad/src/git/refs.rs
+++ b/librad/src/git/refs.rs
@@ -16,7 +16,7 @@
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
 use std::{
-    collections::{BTreeMap, HashMap, HashSet},
+    collections::{BTreeMap, BTreeSet, HashMap},
     fmt::Debug,
     hash::Hash,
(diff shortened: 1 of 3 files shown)


xla (approved):
`Ord`erly

🐭 🚀 🐠 ⛄️

kim (commented):
While we're at it, maybe we could just define it like so:

```rust
pub struct Remotes<A>(BTreeMap<A, BTreeMap<A, BTreeMap<A, ()>>>
```

While `BTreeSet` is implemented as `BTreeMap<A, ()>`, `serde` can't see that. If we use a map, the json will look like:

```json
{"a": {"b": {"c": null}}}
```

which would allow to more easily transition to a better type or more degrees of tracking.

kim:
Like this: #409

PR #407, opened 2020-11-09
Use BTreeSet instead of HashSet

FintanH:
A forward patch of https://github.com/radicle-dev/radicle-link/pull/406

Using a HashSet the iter method will return items in an arbitrary order.
Since this is what serde uses under the hood to turn HashSet into an
array, it destabilises the signing and verification of refs.

Using a BTreeSet gives us a guaranteed order for the array, meaning
consistent de/serialization across peers.

### DIFF
--- a/librad/src/git/fetch.rs
+++ b/librad/src/git/fetch.rs
@@ -557,7 +557,6 @@ mod tests {
     #[test]
     fn replicate_looks_legit() {
         use crate::git::refs::{Refs, Remotes};
-        use std::collections::HashMap;
 
         lazy_static! {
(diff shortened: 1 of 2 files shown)


FintanH:
Seems to be a flake :( https://buildkite.com/monadic/radicle-link-seedling/builds/1502#263e6b25-8015-4f64-87c4-e414c15e13f0

FintanH:
> Seems to be a flake :(

Worked on a re-run. I'm guessing we're still suffering from the local transport

kim (approved):

PR #409, opened 2020-11-09
Terminate remotes graph with a map

kim:
Such that serde sees it as a JSON object, which allows for better
evolvability in the future.


FintanH:
Works for me :100:

PR #411, opened 2020-11-10
Add Info::Applied

FintanH:
When having tracked a peer we don't immediately have their changes. We
can know that we do have the peer's changes when we have fetched their
changes or if we have applied them from a gossip message.

We need a way to surface that a result was successfully applied. In this
patch we do this by adding it to the `Info` enum and emit the protocol
message to the subscribers.

### DIFF
--- a/librad/src/net/gossip.rs
+++ b/librad/src/net/gossip.rs
@@ -101,6 +101,7 @@ where
     Addr: Clone + PartialEq + Eq + Hash,
 {
     Has(Has<Addr, A>),
+    Applied(A),
 }
 
 #[derive(Clone, Debug)]
@@ -683,11 +684,15 @@ where
                             self.broadcast(
                                 Have {
                                     origin: self.local_peer_info(),
-                                    val,
+                                    val: val.clone(),
                                 },
                                 remote_id,
                             )
-                            .await
+                            .await;
+
+                            self.subscribers
+                                .emit(ProtocolEvent::Info(Info::Applied(val)))
+                                .await;
                         },
 
                         // Meh. Request retransmission.
--- a/seed/src/lib.rs
+++ b/seed/src/lib.rs
@@ -251,6 +251,15 @@ impl Node {
                     }
                 }
             },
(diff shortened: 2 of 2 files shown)


seed/src/lib.rs
```diff
@@ -251,6 +251,15 @@ impl Node {
...
             },
+            ProtocolEvent::Gossip(gossip::Info::Applied(peer::Gossip { urn, rev, origin })) => {
+                let peer_id = origin.map(|peer| peer.to_string());
+                tracing::info!(
+                    "Applied change for URN `{}`, rev = {:?}, peer.id = {:?}",
+                    urn,
+                    rev,
+                    peer_id
+                );
+            },
```
cloudhead, line 262: This is gonna be super nice to have!

librad/src/net/gossip.rs
```diff
@@ -101,6 +101,7 @@ where
     Addr: Clone + PartialEq + Eq + Hash,
 {
     Has(Has<Addr, A>),
+    Applied(A),
```
cloudhead, line 104: Does this belong here? Shouldn't this only be for messages that are actually gossiped between peers?

FintanH: I'm not sure I understand the question, could you elaborate? :)

cloudhead: Should the `Applied` variant be here, since it won't be broadcast to peers?

FintanH: `Info` isn't for broadcasting to peers. It's used for announcing protocol events, afaict.

FintanH: Where protocol events are internal for the current peer

kim:
Technically, this is not really _surfacing_ anything -- would you not want to be notified via the `PeerEvent`s?

FintanH:
> Technically, this is not really surfacing anything -- would you not want to be notified via the PeerEvents?

Ah, well it seems I'm confused between the two :sweat_smile: It seems like this may not actually be needed because I surfaced the result in my previous set of changes: https://github.com/radicle-dev/radicle-link/blob/maint/librad/src/net/peer.rs#L124. Would that be right?

PR #412, opened 2020-11-10
Include fixes

FintanH:
If no fetch refspec was provided we would incorrectly use our own PeerId
in the refspec. This is essentially useless because we shouldn't have
our own PeerId under our remotes.

Instead, we change the interface so that remotes are added by providing
the PeerId and handle, then building the Remote with the fetch refspec
internally.

### DIFF
--- a/librad/src/git/include.rs
+++ b/librad/src/git/include.rs
@@ -28,7 +28,7 @@ use tempfile::NamedTempFile;
 use crate::{
     git::{
         local::url::LocalUrl,
-        types::{remote::Remote, AsRefspec, FlatRef, Force},
+        types::{remote::Remote, FlatRef, Force},
     },
     meta::user::User,
     peer::PeerId,
@@ -45,6 +45,9 @@ pub enum Error {
 
     #[error(transparent)]
     Git(#[from] git2::Error),
+
+    #[error("a remote must be set with a fetch refspec when creating an include entry")]
+    MissingRefspec,
 
     #[error(transparent)]
     Persist(#[from] tempfile::PersistError),
@@ -70,7 +73,7 @@ pub enum Error {
 /// ```
 pub struct Include<Path> {
     /// The list of remotes that will be generated for this include file.
-    pub remotes: Vec<Remote<LocalUrl>>,
+    remotes: Vec<Remote<LocalUrl>>,
     /// The directory path where the include file will be stored.
     pub path: Path,
     /// The namespace and `PeerId` this include file is interested in. In other
@@ -92,6 +95,12 @@ impl<Path> Include<Path> {
         }
     }
 
+    pub fn add_remote(&mut self, url: LocalUrl, peer: PeerId, handle: &str) -> Result<(), Error> {
+        let remote = Self::build_remote(url, peer, handle)?;
+        self.remotes.push(remote);
+        Ok(())
+    }
+
     /// Writes the contents of the [`git2::Config`] of the include file to disk.
     pub fn save(self) -> Result<(), Error>
     where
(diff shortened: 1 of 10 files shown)


kim:
> we would incorrectly use our own PeerId in the refspec

Where is this happening? I can't seem to spot it in the code here

FintanH:
> Where is this happening? I can't seem to spot it in the code here

[Here](https://github.com/radicle-dev/radicle-link/pull/412/files#diff-b2cdbd6dedfb9c88ab61da2cb96e6adc3c867e1b9662fb3f94d34436bb484f54L171-L174), particularly the line `remote.url.local_peer_id`. I think I was blindly following the types but not thinking of the semantics.

librad/src/git/types/remote.rs
```diff
@@ -100,6 +100,11 @@ impl<Url> Remote<Url> {
         }
     }
 
+    pub fn with_refspec(mut self, refspec: Box<dyn AsRefspec>) -> Self {
```
kim, line 103: I feel like we have all permutations of `with_x(mut self) -> Self`, `set_x(&mut self) -> &mut Self`, `with_x(self) -> Self`, `x(&mut self) -> &mut Self`, `set_x(&mut self) -> ()`, etc. somewhere in the code. I'll book a `C_IDIOTMATIC` workshop with the styleguide polizei for next year, ok?

librad/src/git/include.rs
```diff
@@ -92,6 +95,12 @@ impl<Path> Include<Path> {
         }
     }
 
+    pub fn add_remote(&mut self, url: LocalUrl, peer: PeerId, handle: &str) -> Result<(), Error> {
```
kim, line 98: If the `handle` is rejected as a refname, we would fail the entire `from_tracked_users`, right? 

Perhaps we should actually be more restrictive about what can be used as a name, so we can define this as `impl Into<RefLike>`... :thinking:

kim (approved):
lgtm

librad/src/git/types/remote.rs
FintanH: Yaw, we have them here https://github.com/radicle-dev/radicle-link/blob/next/librad/src/git/types/reference.rs#L127-L140.

We need polymorphic ownership lenses :pray:

librad/src/git/include.rs
FintanH: Ya, I had a similar thought. Although the name is actually `<handle>@<peer>`. If `handle` is `RefLike` would that imply `<handle>@<peer>` is `RefLike`? Probably yes, right? It'd be `handle.into().join("@").join(peer)`.

I'd prefer create a follow up ticket for making the remote name `RefLike`, if that's ok with you? :)

FintanH: Created an issue here: https://github.com/radicle-dev/radicle-link/issues/414

kim: > handle.into().join("@").join(peer)

Except as per commandment 9 of [git-check-ref-format](https://git-scm.com/docs/git-check-ref-format), the single character "@" is not a valid refname. :low_brightness:

librad/src/git/types/remote.rs
kim: Who needs polymorphic anything if there are CODING CONVENTIONS?!

librad/src/git/include.rs
FintanH: That's #414's problem :grimacing:

FintanH:
@kim qualms with the current state of this?

xla (approved):
🧀 🚾 🌾 💶

kim (approved):

librad/src/git/include.rs
```diff
@@ -101,11 +110,22 @@ impl<Path> Include<Path> {
...
             let mut config = git2::Config::open(tmp.path())?;
             for remote in &self.remotes {
-                let (key, url) = url_entry(&remote);
-                config.set_str(&key, &url.to_string())?;
-
-                let (key, fetch) = fetch_entry(&remote)?;
-                config.set_str(&key, &fetch)?;
+                let (url_key, url) = url_entry(&remote);
+
+                let (fetch_key, fetch) = match fetch_entry(&remote) {
```
kim, line 115: I'm having second thoughts here: why are we ignoring misconfigured remotes instead of returning an error? @FintanH

FintanH: Theoretically, we can only add correctly configured remotes using https://github.com/radicle-dev/radicle-link/pull/412/files#diff-b2cdbd6dedfb9c88ab61da2cb96e6adc3c867e1b9662fb3f94d34436bb484f54R168. But this is really a poor excuse.

Should we `expect` or return an error instead?

PR #415, opened 2020-11-11
Upgrade toolchain + switch to target caching instead of sccache

kim:
sccache has proven quite ineffective unless more than one workspace can benefit from it.

Also, reorganise the build steps for better parallelism and improved workflow when upgrading the toolchain.

_Also_, FRESH new clippy pedantery now builtin! For elevated Freude, it now complains about the exact opposite than before!


FintanH (approved):
OKR: `ok_or`

PR #417, opened 2020-11-11
Use in_scope and instrument

FintanH:


FintanH:
Turns out using `enter` in `async` is very bad :grimacing: [RTFM](https://docs.rs/tracing/0.1.21/tracing/struct.Span.html#method.enter) -- Read `tracing` for me

FintanH:
I don't believe this needs a forward port because of the `instrument` macro [being used](https://github.com/radicle-dev/radicle-link/blob/next/librad/src/net/peer.rs#L506).

As long as:
> If tracing_futures is specified as a dependency in Cargo.toml, async fns may also be instrumented:[0]
[0]https://docs.rs/tracing/0.1.21/tracing/attr.instrument.html#examples

[which it is](https://github.com/radicle-dev/radicle-link/blob/next/librad/Cargo.toml#L36)

kim (approved):

xla (approved):
Sicko

🗝 🤗 🌸 🐄

PR #421, opened 2020-11-16
Bracket local transport

### LINKED ISSUE #381: Safer API for programmatic local transport
In order to make using the local transport as a custom `libgit2` transport a bit
less obscure (see #377), we should constrain ourselves to using it in a
disciplined way (as opposed to relying on the URL).

This would be a breaking change, though:

Instead of keying the per-peer global registration by `PeerId`, we would need to
use a unique token, such that we could define a bracket-style function:

```rust
pub fn with_local_transport<F, A>(url: GitUrl, settings: Settings, f: F) -> A
where
    F: FnMut(&git2::Remote) -> A
```

which would ensure `register` is called, the settings are registered, and after
`f` completes polls the `Results` for some time. The function should panic if
`Results::wait` returns a `thread::Result::Err`.

This could also be formulated as a `struct` with `Deref<Target = git2::Remote>`
and a `Drop` impl which does the waiting.

kim:
Fixes #381 

Collaterals:

* Use trait object for `Storage` signer

  Turns out to be the less painful workaround for the impossible arrow `&Storage<S: Signer> -> &Storage<BoxedSigner>`

* Refine push and fetch spec types

  A push spec MUST resolve to a single ref (ie. it can not contain any patterns). `libgit2` doesn't seem to support expressions (e.g. `HEAD~2`), and neither do we. Mirroring `git push --all`, we do, however, allow resolving matching refs from the local repo. 

  It is also rather confusing to talk about a "local" and "remote" end of a `Refspec`, as that depends on whether it is used as a fetch or push spec -- instead, use git terminology (`src`, `dst`), and force "named args" throughout.

* Demote namespacedness of the `Reference` type

  It turns out that there are quote surprising interactions with impl resolution and inference, due to [fundamental types](https://doc.rust-lang.org/stable/reference/glossary.html#fundamental-type-constructors) and `PhantomData` not being a [local type](https://doc.rust-lang.org/stable/reference/glossary.html#local-type). As we don't have a use case (that I'm aware of) which would require distinguishing namespaced and non-namespaced refs _statically_, just storing an `Option` (again) seems way more, errr, errrgonomic.


FintanH (approved):
Great stuff. I love the extra work put into `Remote` and super glad the local transport stuff is cleaned up :bow:

kim:
"quote suprising" -- quote me on that!

xla (approved):
utAck

🐨 👩 🔫 ♍️

kim:
Closed via f3c5cc0f0368c75c0c5dd0352d3f9526237fb394

PR #423, opened 2020-11-16
Users are persons

### LINKED ISSUE #419: Rename "User" to "Person"
All occurences of the term "User" in the context of identities should be replaced by "Person".

Rationale:

* user of what? Applications have users, but protocols don't (not in the same sense at least)
* There's a derogative connotation to the term: users are these mostly clueless beings who shallt be shielded from the ways of the komputermachine
* We are indeed talking about "personal identities" in the spec, which is more precise terminology: in the same sense RFC4880 describes how User IDs (there: only name + email) can be _bound_ to an identity, extensibility of the identity document payload admits _binding_ more than one external identifier (including _user_ identifiers of other systems) to the identity.

kim:
Fixes #419

### DIFF
--- a/librad/src/identities/git.rs
+++ b/librad/src/identities/git.rs
@@ -26,7 +26,7 @@ use crate::{
     identities::{
         delegation::{self, Delegations},
         generic::{self, Signed, Verified},
-        payload::{self, ProjectPayload, UserPayload},
+        payload::{self, PersonPayload, ProjectPayload},
         sign::{Signature, Signatures},
         urn,
     },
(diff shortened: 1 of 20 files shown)


librad/src/identities/payload.rs
```diff
@@ -655,9 +655,9 @@ mod tests {
         }
     }
 
-    fn gen_user_payload() -> impl Strategy<Value = UserPayload> {
-        (any::<User>(), proptest::option::of(any::<UpstreamUser>())).prop_map(|(user, up)| {
-            let mut u = UserPayload::new(user);
+    fn gen_user_payload() -> impl Strategy<Value = PersonPayload> {
```
FintanH, line 658: Dunno how far you want to go with renaming, but the function names still mention `user`
```suggestion
    fn gen_person_payload() -> impl Strategy<Value = PersonPayload> {
```

librad/src/identities/payload.rs
```diff
@@ -681,13 +681,13 @@ mod tests {
...
         prop_oneof![
-            gen_user_payload().prop_map(SomePayload::User),
+            gen_user_payload().prop_map(SomePayload::Person),
             gen_project_payload().prop_map(SomePayload::Project)
         ]
     }
 
-    fn gen_user_delegations() -> impl Strategy<Value = UserDelegations> {
-        proptest::collection::btree_set(gen_public_key(), 1..32).prop_map(UserDelegations)
+    fn gen_user_delegations() -> impl Strategy<Value = PersonDelegations> {
```
FintanH, line 689: Here too:
```suggestion
    fn gen_person_delegations() -> impl Strategy<Value = PersonDelegations> {
```

librad/src/identities/git.rs
```diff
@@ -98,8 +98,8 @@ impl<'a, T: 'a> From<&'a git2::Repository> for Identities<'a, T> {
 }
 
 impl<'a, T: 'a> Identities<'a, T> {
-    /// Convenience to specialise `T` to [`User`].
-    pub fn as_user(&self) -> Identities<'_, User> {
+    /// Convenience to specialise `T` to [`Person`].
+    pub fn as_user(&self) -> Identities<'_, Person> {
```
FintanH, line 102:
```suggestion
    pub fn as_person(&self) -> Identities<'_, Person> {
```

librad/src/identities/git.rs
```diff
@@ -708,16 +708,16 @@ impl<'a> Identities<'a, Project> {
 
     fn updated_user(
```
FintanH, line 709:
```suggestion
    fn updated_person(
```

FintanH (changes requested):
There's mention of "users" in the identity resolution doc as well, for example in the second paragraph here https://github.com/radicle-dev/radicle-link/blob/user-is-person/docs/rfc/identity_resolution.md#motivation. It would make sense to keep that in line with the change in language here.

kim:
> in the identity resolution doc

Yaah, but that also uses "entities" -- as that is not "official specification", I'd rather leave it as is. I've updated [identities.md](./docs/spec/identities.md)

kim:
I also left seed stuff untouched. Not sure... that seems more like an "upstream server" at the moment.

FintanH:
> Yaah, but that also uses "entities" -- as that is not "official specification", I'd rather leave it as is. I've updated identities.md

Fair enough :+1: I can rework it to make it more official at some point :)

FintanH (dismissed):
I persist therefore I am

xla (approved):
🏮 🕷 🔧 ☑️

FintanH (approved):
I'm persistent therefore I am

PR #424, opened 2020-11-17
sigrefs: define remotes (tracking graph) as a recursive type

### LINKED ISSUE #408: Make git::refs::Remotes recursive
Our `Remotes` data structure is defined with an arbitrary cutoff of 3-degrees and is basically a new type over the std hash collections. The real definition of remotes can be better defined as a recursive data structure (but Rust's recursive data structures are PITA). This would give us more power over the semantics of the structure, one benefit being that we can make the cutoff configurable.

kim:
Closes #408


librad/src/git/refs.rs
```diff
@@ -50,68 +49,106 @@ use crate::{
...
+                v.cutoff_mut(depth)
+            }
+        } else {
+            self.clear()
+        }
+    }
 
-    fn deref(&self) -> &Self::Target {
-        &self.0
+    /// Yield all `A` values in an unspecified order.
```
FintanH, line 117: It would be good to note that they may not be unique as well. I didn't think of that until I saw the test below :)

FintanH (commented):
Nice :+1: One note that would be useful :point_down:

xla (approved):
utACK

👋 📛 🌊 🔓

FintanH (approved):

kim:
Closed via 66b70fb318d34ff82f661610655d3e7e9dfd4aab

PR #430, opened 2020-11-20
net: generate X509 certificates from signer only

### LINKED ISSUE #195: Implement rustls::ResolvesServerCert
In our quest to be able to delegate signing to a vault, agent, or whatever, we need to implement an appropriate `rustls::ResolvesServertCert`.

See also: radicle-dev/radicle-keystore#1

kim:
Removes the requirement for an actual, PKCS8-endoded secret key. That is, secret key material can now be kept in agents or HSMs.

~~Relies on a patch to `picky-asn1-x509` for ed25519 algorithm identifiers: https://github.com/Devolutions/picky-rs/pull/72~~

Closes #195


kim (commented):
@xla, @FintanH: let me know if this should be backported

kim:
Honestly, also, we should never have moved away from BTX.

librad/src/net/x509.rs
```diff
@@ -0,0 +1,238 @@
...
+        let subject = issuer.clone(); // self-signed
+        let subject_public_key_info = SubjectPublicKeyInfo {
+            algorithm: algorithm.clone(),
+            subject_public_key: PublicKey::Ed(
+                BitString::with_bytes(signer.public_key().as_ref()).into(),
+            ),
+        };
+        let validity = {
+            let not_before = GeneralizedTime::new(1970, 1, 1, 0, 0, 0).unwrap().into();
+            let not_after = GeneralizedTime::new(2070, 1, 1, 0, 0, 0).unwrap().into();
```
FintanH, line 100: Climate change will have taken us out by then

FintanH (dismissed):
Nice, looks good to me :metal: 

I don't think it needs back-porting.

librad/src/net/x509.rs
kim: This is actually a suprisingly non-trivial issue: it is _required_ to specify validity, and reasonable values would be approx. now - approx. a month or so (there is a marginal security benefit to that, as the certificate owner basically renews the proof of possessing the key). But that would mean that the network stack has to be restarted once a month.

FintanH: So we include `rad-cron-job`? :grimacing: Would be worth capturing this issue in a ticket to think about it in the long term.

xla (dismissed):
Sick!

📗 🏣 🍒 🌋

FintanH (approved):

librad/src/net/x509.rs
FintanH: How does the latest commit fix this now? So we make the cert valid for 3 months but does that get refreshed somehow?

kim: Sensei Wu likes to say: never put off until tomorrow what you can do today.

The only thing I'm wondering about is: re-generating a certificate can only fail due to the signer returning an error. This is actually a rather fatal situation -- perhaps a `panic` would be appropriate then?

librad/src/net/tls.rs
```diff
@@ -15,157 +15,156 @@
...
+        let now = OffsetDateTime::now_utc().date();
+        let read = self.cert.read().unwrap();
+        if now >= read.expires {
+            drop(read);
+            let fresh = x509::Certificate::generate(&self.signer)?;
+            let expires = x509::validity_time_as_date(&fresh.tbs_certificate.validity.not_after);
+            let der = rustls::Certificate(fresh.to_der());
+            {
+                let mut cert = self.cert.write().unwrap();
+                *cert = Cert {
```
FintanH, line 113: Ah we refresh it here! :)

librad/src/net/tls.rs
```diff
@@ -15,157 +15,156 @@
...
+        }
+    }
+
+    /// Get the certificate, or generate a fresh one if it has expired
+    fn cert(&self) -> Result<rustls::Certificate, BoxedSignError> {
+        let now = OffsetDateTime::now_utc().date();
+        let read = self.cert.read().unwrap();
+        if now >= read.expires {
+            drop(read);
+            let fresh = x509::Certificate::generate(&self.signer)?;
```
FintanH, line 108: @kim I presume this where you mean its fatal? Sounds reasonable to me to panic here. It would mean something is _terrrrrribly_ wrong.

kim: I went for logging the error -- the signer can fail all over the place, and it could be intermittent

PR #432, opened 2020-11-20
keys: remove Copy impl for SecretKey and zero memory on drop

### LINKED ISSUE #349: Make SecretKey **not** Copy
`SecretKey` should **not** be `Copy`, and should zero its memory when dropped
(which the `sodiumoxide` type does, but not the `ed25519-zebra` one).

kim:
Fixes #349


xla (approved):
🏏 ▪️ 💚 🗼

PR #433, opened 2020-11-20
Various guards against dancing with oneself

### LINKED ISSUE #410: Tracking own PeerId
While testing the Upstream application we noticed that it's possible that one can end up tracking their own remote. This was surfaced when collaborating between 3 peers and a seed.

kim:
- replication: do not track self
- replication: deny replicating from self
- net: deny ingress + egress connections to local peer id

Does NOT fix #410, but can't hurt either.

### DIFF
--- a/librad/src/git/replication.rs
+++ b/librad/src/git/replication.rs
@@ -43,6 +43,9 @@ pub use crate::identities::git::Urn;
 #[derive(Debug, Error)]
 #[non_exhaustive]
 pub enum Error {
+    #[error("cannot replicate from self")]
+    SelfReplication,
+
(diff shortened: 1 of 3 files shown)


xla (dismissed):
Can we backport this to `maint`?

🗻 🎉 🙉 ☣

FintanH (dismissed):

kim:
@xla: Actually, I don't think this is necessary. We have

https://github.com/radicle-dev/radicle-link/blob/maint/librad/src/git/storage.rs#L951

and on the connection level the TLS handshake should actually reject connections to/from the same peer id.

Issue #431, opened 2020-11-20
net: quick reconnect will terminate both old and new connection


xla:
When quickly reconnecting (inside the keep alive max timeout) the old and new connection will be forcefully disconnected, likely because the old `recvd` stream when terminating with timed_out will hit this:

https://github.com/radicle-dev/radicle-link/blob/a3a6be979218311bf8df029003d5a851ca1329c8/librad/src/net/gossip.rs#L555-L556

kim:
Interestingly, I don't think there is a solution to this unless we invert the logic and reject new connections from the same peer, instead of ejecting existing ones. The problem is that we would need to wait until all futures which process streams of an existing connection have completed before we start processing the new one -- but that is not possible.

Perhaps the inverted logic is better anyway.

xla:
> Interestingly, I don't think there is a solution to this unless we invert the logic and reject new connections from the same peer, instead of ejecting existing ones. The problem is that we would need to wait until all futures which process streams of an existing connection have completed before we start processing the new one -- but that is not possible.

Would it help if we store some form of coordination primitive like a oneshot channel. Which when used to select in the same loop where the gossip messages are worked on, can be used to signal shutdown of the existing recv loop and helps determining that it has completed.

While punishing quick reconnects is not an unsound strategy it might impede heavily with spotty connectivity. Is there any possibility to "handover" the existing stream between connections? So when a peer reconnects the old one is removed and business continues as usual.

kim:
Before I respond with more thoughts: could you explain how this reconnection happens? Like, are you dropping the networking state, and then re-initialising it? Or is this something you observe with peers in steady state?

xla:
This happens when people restart Upstream, mostly when things get stuck (e.g. replication not working as expected). So the networking state is completely dropped. How contrive that is I don't know, I also can't say if it happens for steady peers, as we didn't have enough stored logs to contextualize.Definitely something worse investigating.

How does quic behave with network (wifi et al) interruptions, when the network state is preserved? Will this trigger the same connection duplication occur? Or will it, due to it's UDP genome, just continue with the same connection once packets roll on within the keep-alive period?

kim:
[...]

PR #434, opened 2020-11-23
protocol: reject duplicate connections

kim:
The idea to eject previous connections from a remote peer is racy, as described in #431. Instead, we reject new incoming connections until the old one is either closed gracefully, or times out.

### DIFF
--- a/librad/src/net/protocol.rs
+++ b/librad/src/net/protocol.rs
@@ -370,10 +370,12 @@ where
                     "Run::Discovered",
                 );
 
-                if !self.connections.lock().await.contains_key(&peer) {
-                    if let Some((conn, incoming)) = connect(&self.endpoint, peer, addrs).await {
-                        self.handle_connect(conn, incoming.boxed(), None).await;
-                    }
+                // XXX: needs reconsideration: we leave it to the supplier of the
+                // disco stream to initiate a re-connect by yielding an element
+                // we actually saw before. The consequence is that existing
+                // connections will be terminated.
+                if let Some((conn, incoming)) = connect(&self.endpoint, peer, addrs).await {
+                    self.handle_connect(conn, incoming.boxed(), None).await;
                 }
             },
 
@@ -433,6 +435,8 @@ where
                             Some(conn) => match conn.open_stream().await {
                                 Ok(stream) => {
                                     drop(connections);
+                                    // The incoming future should still be running,
+                                    // so it's enough to drive an outgoing
                                     if let Err(e) = self.outgoing(stream, None).await {
                                         tracing::warn!("error handling outgoing stream: {}", e);
                                     }
@@ -471,7 +475,6 @@ where
 
         {
             let mut connections = self.connections.lock().await;
-            // Should not actually be possible, but let's handle the case
             if let Some(prev) = connections.insert(remote_id, conn.clone()) {
                 tracing::warn!(
                     "new outgoing ejects previous connection to {}@{}",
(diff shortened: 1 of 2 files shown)


FintanH (dismissed):
LGTM :metal:

librad/src/net/protocol.rs
```diff
@@ -503,18 +518,32 @@ where
...
+
+            conn.close(CloseReason::DuplicateConnection);
+            drop(incoming);
+
             self.subscribers
-                .emit(ProtocolEvent::Connected(remote_id))
+                .emit(ProtocolEvent::Disconnecting(remote_id))
                 .await;
-        }
+            drop(connections);
```
xla, line 531: drop

xla: What does this drop do? Is it to guarantee that the mutexguard is dropped?

xla (commented):
I dunno if it is relevant in the context of this PR, but what does when the gossip stream terminates, shouldn't the connection be disconnected as well?  Currently it's only logged:

https://github.com/radicle-dev/radicle-link/blob/c65030894d51a7ec39fdbeb24d901ade5adb09b7/librad/src/net/protocol.rs#L528-L532

Alternatively is there a way to bring up a new gossip stream? The reason I'm bringing this up is that we observed the gossip stream exiting without the protocol bubbling up a disconnect event.

librad/src/net/protocol.rs
kim: Yes

xla (approved):
🔱 🛀 🤑 🛤

kim:
>  when the gossip stream terminates, shouldn't the connection be disconnected as well

Yeah right -- it shouldn't. The connection itself is actually independent of gossip membership. There is currently code which terminates the connection when _gossip_ decides to eject a peer, but that should be removed.

> Alternatively is there a way to bring up a new gossip stream

That's indeed a bit of a problem currently, because we keep the connection alive even if it is completely idle (ie. there is no `git` stream either). This means that it can only become active again if either end promotes it to the active set.

xla:
Can we consider #431 fixed with this changeset?

FintanH (approved):

xla (approved):
📝 🍡 🗼 💁

xla (approved):
🌵 📣 🚇 🍛

kim:
I suspect I might be missing a `spawn` somewhere...

FintanH:
> I suspect I might be missing a spawn somewhere...

Here you go
image

FintanH (dismissed):
Looks good to me

kim:
That's a beautiful spawn.

Ya, so, I'm not sure. We only spawn incoming streams. Need to revisit that at some point.

kim:
> is there a way to bring up a new gossip stream?

This is not properly addressed yet.

xla:
@kim You wanna look at this before merging, or can go forward with this one?

kim:
This should be approximately correct now (watch out for `XXX`). Is there are way we can test this in production?

xla:
> This should be approximately correct now (watch out for `XXX`). Is there are way we can test this in production?

Yes, something specifically? Or just the quick reconnect of peers to the seed?

kim:
> the quick reconnect of peers to the seed

This

kim:
@xla does it seem alright?

xla:
> @xla does it seem alright?

Testing as we speak.

xla (approved):
🚌 👚 🦁 🐹

FintanH (approved):

PR #435, opened 2020-11-23
spec: Reorganise tree

kim:
Move the spec tree under `docs/`, and start splitting up into smaller
section documents. Also set up the docs directory for serving via GH
pages.


kim:
Possibly some more sophistication could be applied to generating the served site tree, contributions welcome

PR #436, opened 2020-11-23
seed: include guessed user for project tracked

xla:
To enable radicle-dev/radicle-bins#43 the additionl user metadata is
also guessed when a project is tracked. This will help get to know of
new identities without having to restart the seed.


rudolfs (approved):

rudolfs (approved):

FintanH (changes requested):
If we've tracked the project, we should be able to `get_rad_self_of` using that project `URN` and the `PeerId`, without going through this guessing business.

seed/src/event.rs
```diff
@@ -21,36 +21,52 @@ use librad::{meta::project::ProjectInfo, net::peer::PeerApi, peer::PeerId, uri::
 
 use crate::{guess_user, Error, Project, Signer};
 
+/// A peer with optional user metadata.
+#[derive(Clone, Debug)]
+pub struct User {
+    pub peer_id: PeerId,
+    pub urn: Option<RadUrn>,
+    pub name: Option<String>,
+}
```
cloudhead, line 30: Kind of weird that this is called `User` when the doc says "peer". Based on the fields, I think this should be called `Peer`, because a user without a URN or a name is not a user.

xla: `Peer` doesn't hae a name, so it cuts both ways.

xla:
> If we've tracked the project, we should be able to `get_rad_self_of` using that project `URN` and the `PeerId`, without going through this guessing business.

Could you leave a suggestion in place to contextualise your comment.

FintanH:
> Could you leave a suggestion in place to contextualise your comment.

Sure thing. My familiarity with the seed code isn't great so I wasn't sure exactly _where_ to add this comment, apologies.

seed/src/event.rs
```diff
@@ -72,7 +88,7 @@ impl Event {
                 name: proj.name().to_owned(),
                 description: proj.description().to_owned(),
             },
-            provider,
+            User::guess_from_peer(api, provider).await?,
```
FintanH, line 91: So here we've tracked the `Project` and we have a `provider`. So we can actually do:
```rust
storage.get_rad_self_of(urn, provider)
```
in the above and return `proj` and the `user` for the `proivder`

cloudhead:
I'm not sure this PR solves the issue at hand  -- we need to guess the users of the connected peers, and we should try to do that everytime a new project is tracked (since it contains user information). But this PR seems to only guess the user of the provider for a given tracked project, which could be the same for all projects.

xla:
> I'm not sure this PR solves the issue at hand -- we need to guess the users of the connected peers, and we should try to do that everytime a new project is tracked (since it contains user information). But this PR seems to only guess the user of the _provider_ for a given tracked project, which could be the same for all projects.

So what are you proposing?

xla:
> I'm not sure this PR solves the issue at hand -- we need to guess the users of the connected peers, and we should try to do that everytime a new project is tracked (since it contains user information). But this PR seems to only guess the user of the _provider_ for a given tracked project, which could be the same for all projects.

@cloudhead Also the only other user the seed can currently learn about is the maintainer, if it's no the one who announced the project.

cloudhead:
Hmm well, let me think outloud a bit:

1. A bunch of peers connect to the node, and the node tracks projects from them, that they are **no**t contributors to. These are anonymous peers as far as the seed node is concerned.
2. Some new peers now connect, and they have some projects created by the anonymous peers above. The seed nodes replicate them. Now the seed should essentially go back and check all anonymous peers to see if these new projects have a match.
3. Since a single project tracked could de-anonymize an unlimited number of peers, the seed node would then have to send a list of `(PeerId, User)` mappings upon tracking a new project. Either in the `ProjectTracked` event, or via a separate event (cleaner). Eg.:

```rust
let proj_tracked = ProjectTracked(proj, provider);
let users = get_users_from(proj);

channel.send(proj_tracked);
channel.send(UsersDiscovered(users));
```

The user of the lib can then update its `peer -> user` mapping based on newly discovered users..

FintanH:
> I'm not sure this PR solves the issue at hand -- we need to guess the users of the connected peers, and we should try to do that everytime a new project is tracked (since it contains user information). But this PR seems to only guess the user of the provider for a given tracked project, which could be the same for all projects.

So are you saying that we have a list of connected `PeerId`s and that we want to fill in `User` information for them? I suppose when you hear about a project you could try `get_rad_self_of` using any of the `PeerId`s that haven't been filled in. They may not exist which will be an error so you can skip over them.

I wonder if it's interesting for the protocol, in the future, to slowly introduce this discovery type information. So not only do you say your `PeerId`, but you also advertise what your `User` is.

seed/src/event.rs
```diff
@@ -65,14 +81,25 @@ impl Event {
             })
             .await??;
 
+        let user = api
+            .with_storage({
```
FintanH, line 85: Why not use the same `with_storage` call above and reduce the fetches from the pool?

xla: It makes error handling more involved and awkward.

rudolfs (approved):

xla:
> I would say it would be better discussed how to get advertised user info in an issue and decide on a better solution there.

Part of `next`.

xla:
@cloudhead I think what you suggesting with the independent event for discovered users makes a lot of sense, it's more scalable no matter how we gonna discover users in the future or how many. Will rework.

cloudhead:
> @cloudhead I think what you suggesting with the independent event for discovered users makes a lot of sense, it's more scalable no matter how we gonna discover users in the future or how many. Will rework.

Awesome, also happy to merge as-is and re-work in the future, up to you!

seed/src/event.rs
cloudhead: :thinking:

kim:
A few things:

* There are no users, there are [persons](https://github.com/radicle-dev/radicle-link/pull/423). It might be a good time to start thinking in this terminology, in order to avoid misunderstandings.
* There is no 1:1 mapping between peers and personal identities — it is only meaningful to talk about what a peer’s personal identity is in the context of a _project_ (or the top-level identity itself). That is what the `self` ref is for — there is no need for another protocol primitive.
* It is correct that the `self` ref is not compulsory. It would be a wrong conclusion though that instead the whole repo space should be searched for all or any identity which happens to delegate to the peer at hand. Even if I cannot convince you that this isn’t a good idea, the relationship is still 1:n.
* It isn’t clear to me where this stateless, event-based approach came from, but it clearly needs to be revisited. Again, the only meaningful way to show metadata is project-local.

kim (changes requested):
In other words: I really don’t get whose experience is enhanced by showing this “deanonymised” information publicly. How about including the IP address, just to enhance trust in peer-to-peer networks?

xla:
> There are no users, there are [persons](https://github.com/radicle-dev/radicle-link/pull/423). It might be a good time to start thinking in this terminology, in order to avoid misunderstandings.

Good reminder, you reckon this should be established even outside of `next`?

> * There is no 1:1 mapping between peers and personal identities — it is only meaningful to talk about what a peer’s personal identity is in the context of a _project_ (or the top-level identity itself). That is what the `self` ref is for — there is no need for another protocol primitive.
> 
> * It is correct that the `self` ref is not compulsory. It would be a wrong conclusion though that instead the whole repo space should be searched for all or any identity which happens to delegate to the peer at hand. Even if I cannot convince you that this isn’t a good idea, the relationship is still 1:n.
> 
> * It isn’t clear to me where this stateless, event-based approach came from, but it clearly needs to be revisited. Again, the only meaningful way to show metadata is project-local.

xla:
Thanks for the reviews and the push-back @kim - this will be adequately addressed in radicle-dev/radicle-bins#44.

kim:
What I'm trying to get across is: this sends the wrong message.

Through decades of attempts to criminalise participants of peer-to-peer filesharing networks, everyone should be aware of the fact that p2p is fundamentally the opposite of a privacy-enhancing technology. Our answer to this is not yet fleshed out -- because the only way to counter it is by indirection, which complicates a lot of things.

This screen is served through a public website, and intended to showcase the network. By resolving public keys to user profiles, we are making it very explicit that there is no anonymity, unless you are competent enough to take your own measures. That is honest, but the astute observer will quickly realise that it is preferrable to not use a `name` which is recognisable in other contexts. So how much is won?

It is a different thing to resolve contributor identities (`rad/self`), because contribution is a conciously public action. The whole onboarding flow of `upstream`, however, makes it deceivingly so as if one is signing up for some website, and doesn't account for the fact that essentially just starting the application makes all your actions public. This should be thought through properly.

Until then, I would find it more appropriate to only show peer ids on connection events (again: this is a public website).

PR #438, opened 2020-11-24
Apply Have’s for tracked URNs even if no rev is set

geigerzaehler:
When we receive a `Have` messages for a URN we are tracking we will fetch the repo even if no `branch` is set in the `Have` message. This allows us to receive ref updates and solves https://github.com/radicle-dev/radicle-upstream/issues/1242.

### DIFF
--- a/librad/src/net/peer.rs
+++ b/librad/src/net/peer.rs
@@ -549,47 +549,41 @@ where
                         return PutResult::Error;
                     },
                 };
-                let res = match has.rev {
-                    // TODO: may need to fetch eagerly if we tracked while offline (#141)
-                    Some(Rev::Git(head)) if is_tracked => {
-                        let res = {
-                            let this = self.clone();
-                            this.git_fetch(provider, urn, head)
-                                .instrument(span.clone())
-                                .await
-                        };
+                let res = if is_tracked {
+                    let res = self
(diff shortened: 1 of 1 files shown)


librad/src/net/peer.rs
```diff
@@ -549,47 +549,41 @@ where
...
+                                PutResult::Applied(has.clone())
+                            } else {
+                                span.in_scope(|| {
+                                    tracing::warn!(
+                                        provider = %provider,
+                                        has.origin = ?has.origin,
+                                        has.urn = %has.urn,
+                                        "Provider announced non-existent rev"
+                                    );
+                                    PutResult::Stale
```
FintanH, line 570: @kim Shouldn't this really be an error? We did a successful fetch but for some reason, it wasn't in our store.

FintanH (approved):
:metal:

librad/src/net/peer.rs
kim: Yeaaaahh... I'm not sure. `Error` makes an effort to request retransmission. So if a peer announces things it doesn't actually have, it can cause a lot of traffic. It is probably better to just stop disseminating gossip here.

kim (approved):
lgtm

librad/src/net/peer.rs
FintanH: Ok, makes sense :+1:

PR #439, opened 2020-11-25
protocol: filter out unroutable IP addresses for connect

kim:
The advertised address might be `0.0.0.0`. This is a quick fix, we may want to catch this earlier in the future.

### DIFF
--- a/librad/src/lib.rs
+++ b/librad/src/lib.rs
@@ -20,6 +20,7 @@
 #![feature(bool_to_option)]
 #![feature(btree_drain_filter)]
 #![feature(core_intrinsics)]
+#![feature(ip)]
 #![feature(never_type)]
 
 #[macro_use]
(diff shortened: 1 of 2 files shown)


xla:
Does this affect if two peers are started on the same machine and connect to each other?

xla (approved):
🔸 🎾 🔺 🈁

FintanH (approved):

kim:
@xla: I don't think a connect to `0.0.0.0` can ever succeed -- the receiver of a `Join` rpc sets the `seed_addrs` to what it sees.

Since binding to `0.0.0.0` means "listen on all interfaces", I'm wondering if there are APIs in The Language to determine what "al interfaces" is. Or rather, if we can convince `quinn` to give us access to that. I believe we would need to [pre-fabricate a socket](https://docs.rs/quinn/0.6.1/quinn/generic/struct.Endpoint.html#method.rebind).

PR #442, opened 2020-11-25
forward-port: Apply Have’s for tracked URNs even if no rev is set (#438)

kim:
When we receive a `Have` messages for a URN we are tracking we will
fetch the repo even if no `branch` is set in the `Have` message. This
allows us to receive ref updates and solves https://github.com/radicle-dev/radicle-upstream/issues/1242


xla (dismissed):
🌉 🎫 🍴 🎤

FintanH (dismissed):
Nice :metal:

PR #443, opened 2020-11-25
forward-port: protocol: reject duplicate connections (#434)

kim:
The idea to eject previous connections from a remote peer is racy, as
described in #431. Instead, we reject new incoming connections until the
old one is either closed gracefully, or times out.

This also entails to not automatically terminate connections when a peer
is ejected from the active set of gossip membership.


xla (approved):
👺 👻 🌮 🕵

PR #445, opened 2020-11-25
net: revisit transport params

kim:
* Set keep alive interval to recommended 30s
* Make only one end send keep alive probes at all
* Set idle timeout (on the other end) to tolerate loss of 1-2 keep-alive
  probes

### DIFF
--- a/librad/src/net/quic.rs
+++ b/librad/src/net/quic.rs
@@ -36,17 +36,26 @@ use crate::{
     signer::Signer,
 };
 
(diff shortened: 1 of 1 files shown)


FintanH (dismissed):
Looks good :+1: XXX hot stuff :fire:

xla (dismissed):
🚖 🕴 🖋 📭

xla:
Would be useful for `maint` as well, can whip it up, or do  you have reservations?

PR #446, opened 2020-11-25
net: guard against a panic when connecting

kim:
It is unclear if there is a condition where `seen_addrs` can be empty,
or if `quinn::Connection::remote_address` can return a non-routable
address. This is a quick fix to to prevent `future::select_ok` from
panicking when given an empty iterator.


FintanH (approved):
Alles klar

xla (approved):
🎯 🎱 🍤 🐞

PR #449, opened 2020-11-25
WIP conntrack fixes

kim:


kim:
Yeah, so, I think this is hitting a roadblock: I remember cursing at the fact that we can not get at internal connection and stream IDs, which is exactly what we need here.

PR #458, opened 2020-12-01
next is current

kim:
It may never have been necessary to keep `next` around alongside `maint`, but now we should retire it.

**This is your chance to petition for ~~coloring the bikeshed~~renaming the default branch**


xla:
> This is your chance to petition for ~coloring the bikeshed~renaming the default branch

Says in the title: `current`

Is that a name we can live with? Otherwise I don't mind `main`.

PR #468, opened 2020-12-04
Windows compilation fixes

### LINKED ISSUE #467: Running tests on windows fail due to path separators on RefLike urn tests
It looks like the RefLike and other structures on `name.rs` use interally a `PathBuf` which changes how it behaves when running on windows.

```rust
pub struct RefLike(PathBuf);
```

```
---- identities::urn::tests::is_reflike_with_path stdout ----
thread 'identities::urn::tests::is_reflike_with_path' panicked at 'assertion failed: `(left == right)`
  left: `"hnrkyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy/refs/heads/lolek/bolek"`,
 right: `"hnrkyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy\\refs/heads\\lolek/bolek"`', librad\src\identities\urn.rs:485:9
```

All the tests failing on the repo are due to using the `/` separator on tests directly, but further inspecting the code, we have a few other places which assumes the `/` separator, even tho it's a pathbuf internally.

I'm not sure what the best course of action is here - there is a lot of places using PathBuf internally and I'm not sure if we only need to change the `RefLike` join would be enough

bltavares:
More detail on each change on the commit messages.

All tests are passing, both on Windows 10 and WSL Debian.

Closes #467


git-ext/src/reference/name.rs
```diff
@@ -76,15 +75,15 @@ impl RefLike {
     /// If `base` is not a prefix of `self`, or `base` equals the path in `self`
     /// (ie. the result would be the empty path, which is not a valid
     /// [`RefLike`]).
-    pub fn strip_prefix<P: AsRef<Path>>(&self, base: P) -> Result<Self, StripPrefixError> {
+    pub fn strip_prefix<P: AsRef<str>>(&self, base: P) -> Result<Self, StripPrefixError> {
```
kim, line 78: I'm a bit uncomfortable with having to remember to add a trailing slash to `base`. How about something like this?:

```rust
    let suffix = iter::repeat("").take(self.0.len().saturating_sub(base.len()));

    let path = self.0
        .split('/')
        .zip(base.split('/').chain(suffix))
        .skip_while(|(a, b)| a == b)
        .map(|(x, _)| x)
        .collect::<Vec<&str>>()
        .join("/");
```

librad/src/git/tracking.rs
```diff
@@ -181,7 +181,7 @@ mod tests {
     fn track_is_tracked() {
         let tmp = tempfile::tempdir().unwrap();
         {
-            let paths = Paths::from_root(tmp).unwrap();
+            let paths = Paths::from_root(&tmp).unwrap();
```
kim, line 184: :+1:

macros/src/lib.rs
```diff
@@ -28,13 +28,11 @@ pub fn reflike(input: TokenStream) -> TokenStream {
...
-        Ok(safe) => {
-            let safe_str = safe.as_str();
+        Ok(_) => {
             let expand = quote! {
-                unsafe {
-                    ::std::mem::transmute::<::std::path::PathBuf, ::radicle_git_ext::RefLike>(
-                        ::std::convert::From::from(#safe_str)
-                    )
+                {
+                    use ::std::convert::TryFrom;
```
kim, line 34: What's the reason you removed the `transmute`? We have already computed a valid `RefLike`, so don't have to do it again at runtime.

kim (commented):
Thanks @bltavares for putting the work into this!

Do you have some experience with windows CI services? Might be worth setting up a a basic job.

Also, I need to bug you about our [DCO](./CONTRIBUTING.md)

macros/src/lib.rs
bltavares: The transmute was converting from a PathBuf, so I thought we could remove it, as it still has a From. Maybe it could could straight into the RefLike but I didn't think about it at first, only about removing the PathBuf from the code

git-ext/src/reference/name.rs
bltavares: I agree with the concern on forgetting the trailing slash, but I was also concerned of doing too much work as well to fix it.

What about:

```rust
let base = format!("{}/", base.strip_suffix('/').unwrap_or(base))
```

bltavares:
Yup - I often use [GitHub Actions](https://github.com/actions-rs/cargo) as it has windows support. It would require some extra steps to download the new library to compile the code,

bltavares:
DCO added to commits

macros/src/lib.rs
kim: Yeah, it can just transmute the `safe_str` into a `RefLike`. Alternatively, we can also add a constructor marked `unsafe` on `RefLike` itself, which doesn't validate the `String` argument.

git-ext/src/reference/name.rs
kim: Mmya, although this is not technically doing _less work_...

I'm wary about the allocations introduced by `format!` everywhere -- I suspect this will be noticable eventually, as we're creating a lot of refnames all the time. But I'd be willing to accept this for now, and spend some time optimising in a separate patch.

FintanH (commented):
This looks good to me :) There is a merge conflict @bltavares before I can give the full approval on my side 👍

bltavares:
Merge conflicts resolved, tests passing, signed off 👍

bltavares:
BTW - should I open the PR against `master` or `next`?

kim:
Thanks @bltavares, this is now in `master`. I amended the changes to macros, though.

PR #473, opened 2020-12-11
Return the git remote when saving

FintanH:
Since this function is aware of creating the Remote it's useful to
return it. The cloning behaviour in Upstream used it for setting up the
cloned working copy's remote.

### DIFF
--- a/librad/src/git/types/remote.rs
+++ b/librad/src/git/types/remote.rs
@@ -125,7 +125,7 @@ impl<Url> Remote<Url> {
     /// untouched, if present.
     #[allow(clippy::unit_arg)]
     #[tracing::instrument(skip(self, repo), fields(name = self.name.as_str()), err)]
-    pub fn save(&mut self, repo: &git2::Repository) -> Result<(), git2::Error>
+    pub fn save<'a>(&mut self, repo: &'a git2::Repository) -> Result<git2::Remote<'a>, git2::Error>
     where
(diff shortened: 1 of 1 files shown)


xla (approved):
🏷 🏩 🐩 😜

kim (changes requested):
Ah no, I removed that on purpose -- for the local transport hax0ry to work, you're supposed to not get hold of a raw `git2::Remote` if possible. Since `save` takes `&mut self`, you should own a `Self`, so all should be good, no?

FintanH:
> Ah no, I removed that on purpose -- for the local transport hax0ry to work, you're supposed to not get hold of a raw git2::Remote if possible. Since save takes &mut self, you should own a Self, so all should be good, no?

In this case, the raw remote is _needed_ because of https://docs.rs/git2/0.13.12/git2/build/struct.RepoBuilder.html#method.remote_create. We can just do `repo.find_remote` after in the call since this is probably something that's libgit2 idiotsyncratic

kim:
I wasn't aware you're using `RepoBuilder`. Pretty sure this is broken since #421, the way this works now doesn't work with a raw remote

FintanH:
Ya, I missed it. And ya, the `RepoBuilder` `clone` functionality -- essentially project creation -- should probably be living as a sub-crate in `radicle-link`. I'll close this anyway because we can work around it for this case.

PR #478, opened 2020-12-16
Replicate remotes

FintanH:
* Updated `menage_a_troi` to check the existence of the identity references as well as the commit
* Fetch the identities of the tracked peers

@kim not sure how far off base I am with this :grimacing:


librad/src/git/fetch.rs
```diff
@@ -115,6 +115,42 @@ pub mod refspecs {
         ]
     }
 
+    pub fn peek_of<P, R>(urn: &Urn<R>, remote_peer: P) -> Vec<Fetchspec>
```
kim, line 118: Ah yes.

Maybe we can collapse this into one method which takes a parameter `local_mapping: impl Into<Option<PeerId>>` or something? Also, how do you feel about adjusting the (pure) tests as well?

kim (commented):
Good start. 

I think the more severe issue is that `replicate` falsely assumes a `clone`-like pattern, where it is not known who's `rad/id` is the authoritative one. It thus maps the one of `remote_peer` onto the local one, while it _must not_ do that if we already know the delegates. Iow, if we already have `urn` locally, we know the delegates, and should fetch `Peek { delegates: Vec<PeerId> }`, mapping them to their corresponding remotes.

librad/src/git/fetch.rs
FintanH: Ya, was going to see what you thought of the placement of the refs before cleaning up :smile:

kim:
Alright, wait, maybe we can simplify this.

So, say `alice` wants project `P`. She does not know anything about it except its URN. `alice` has a git connection to `bob`, where `bob` may or may not be in the delegates of `P`. The sensible thing to do for `alice` is to `Peek` at `refs/namespaces/P/refs/rad/id` (ie. `bob`'s view of `rad/id`), and run verification on that. If that passes, she learned that `carol` and `dylan` are in the delegates, so she needs to go and fetch `refs/namespaces/P/refs/remotes/{carol,dylan}/rad/id` from `bob`, run verification on these, and determine if they describe the same history (ie. they're not forks). Once _that_ passes, `alice` can pick either one of `carol` or `bob`'s `rad/id` and create that at the top of _her_ namespace.

Next time `alice` fetches `P` from `bob` (or any other peer), only `carol`, `dylan` + any other tracked peers are interesting. What `bob` thinks is irrelevant.

So, the first thing to fix is that `Peek` maps to `rad/id` instead of `remotes/R/rad/id` -- `rad/id` gets created anyways at the end. Iow, `Peek` needs to _always_ behave like `peek_of`, but _may_ require one _or more_ `PeerId`s (`replicate` decides whether that is `remote_peer` or `carol` & `dylan`).

librad/src/git/replication.rs
```diff
@@ -181,119 +224,248 @@ where
...
+    }
 
-    // Verify + symref the delegates first
-    for delegate in proj.delegations().iter().indirect() {
-        let delegate_urn = delegate.urn();
-        // Find in `refs/namespaces/<urn>/refs/remotes/<remote
-        // peer>/rad/ids/<delegate.urn>`
-        let in_rad_ids: Urn = Reference::rad_delegate(Namespace::from(&urn), &delegate_urn)
+    pub fn verify(storage: &Storage, urn: Urn, remote: PeerId) -> Result<VerifiedProject, Error> {
+        let urn: Urn = Reference::rad_id(Namespace::from(urn.clone()))
```
kim: I think it would be nicer to have a separate function for the `Urn` conversion, without any side effect. The function could then have a doctest.

librad/src/git/replication.rs
```diff
@@ -181,119 +224,248 @@ where
...
+    pub fn verify_delegates(
+        storage: &Storage,
+        urn: Urn,
+        delegates: BTreeMap<PeerId, Urn>,
+    ) -> Result<BTreeMap<PeerId, DelegateProject>, Error> {
+        let mut confirmed_delegates = BTreeMap::new();
+        for remote in delegates.keys() {
+            match verify(&storage, urn.clone(), *remote) {
+                // It's possible that this PeerId for the delegate didn't live on the machine we
+                // fetched from
```
kim, line 296: But that's a protocol violation, and should abort the fetch

librad/src/git/replication.rs
```diff
@@ -181,119 +224,248 @@ where
...
+                    );
+                },
+                Err(err) => return Err(err),
+            }
+        }
+        Ok(confirmed_delegates)
+    }
+
+    #[tracing::instrument(level = "trace", skip(storage), err)]
+    pub fn adopt_delegate(
```
kim, line 317: Not sure -- after having checked for forks, we should also be able to tell which `content_id` is the most recent one.

librad/src/git/identities/project.rs
```diff
@@ -143,6 +151,12 @@ pub fn merge(storage: &Storage, urn: &Urn, from: PeerId) -> Result<Project, Erro
     Ok(next)
 }
 
+pub fn is_fork(storage: &Storage, left: &Urn, right: &Urn) -> Result<Fork, Error> {
+    let left = get(storage, left)?.ok_or_else(|| Error::NotFound(left.clone()))?;
+    let right = get(storage, right)?.ok_or_else(|| Error::NotFound(right.clone()))?;
+    Ok(identities(&storage).is_fork(left.revision.into(), right.revision.into())?)
```
kim, line 157: This passes two `tree` oids, but `Identities::is_in_ancestry_path` expects the first to be a `commit`, and the second to be a `tree` oid.

librad/src/identities/git.rs
```diff
@@ -174,6 +182,25 @@ impl<'a, T: 'a> Identities<'a, T> {
...
 
+    pub fn is_fork_generic(&self, left: git2::Oid, right: git2::Oid) -> Result<Fork, git2::Error> {
+        if left == right {
+            return Ok(Fork::Parity);
+        }
+
+        let left_path = self.is_in_ancestry_path(left, right)?;
+        let right_path = self.is_in_ancestry_path(right, left)?;
+
+        if left_path && right_path {
```
kim, line 193: This is the same as `left == right`.

We can avoid grinding through the history by returning early if `left_path` is `true`, else if `right_path`. If both are `false`, we have a fork. 

There isn't really a way to tell which side was forked, because there is no main side here.

librad/src/git/replication.rs
FintanH: Sorry, this isn't clear. The issue here is that a `Person` can have multiple `Direct` delegates. Only _one_ of these may be the remote `PeerId` on the machine we're replicating from. 

So the invariant is that there must be at least one `PeerId` for `Urn` but we need to find out which one. I suppose we could check that when building the map?

librad/src/git/replication.rs
FintanH: Makes sense :+1:

librad/src/git/identities/project.rs
FintanH: Ah, confusion on my part :+1:  Is `content_id` the commit oid?

librad/src/git/replication.rs
```diff
@@ -127,17 +132,51 @@ where
...
-                Ok(Some(delegates.collect()))
+                let proj = project::ensure_setup(storage, proj, remote_peer)?;
+                Ok(Some(project::delegates(storage, proj)?))
             },
         },
     }?;
 
     let tracked = tracking::tracked(storage, &urn)?.collect::<BTreeSet<_>>();
+
+    // We have fetched the remote's identities, now we fetch the delegate identities
```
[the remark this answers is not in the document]
kim: Oh I think I get it -- the first `Peek` should already consider delegates if we have the project locally, but that doesn't work if we haven't.

I think want to combine `Peek` and `SignedRefs` in that case. Something like:

```rust
enum Fetchspecs {
    Peek,
    SignedRefs,
    Replicate,
    Combo(Vec<Fetchspecs>),
}
```

FintanH:
This should be in better shape. It turned into more of a rewrite so scrutiny is welcome.

librad/src/identities/git.rs
```diff
@@ -174,8 +175,33 @@ impl<'a, T: 'a> Identities<'a, T> {
...
+    }
+
     //// Helpers ////
 
+    /// Assumes that the bag of commits are in a related history.
+    fn latest_generic(&self, bag: NonEmpty<git2::Oid>) -> Result<git2::Oid, git2::Error> {
+        let mut latest = bag.head;
+
+        for commit in bag.into_iter() {
+            if self.repo.graph_descendant_of(latest, commit)? {
```
kim, line 197: I think the argument order of `graph_descendant_of` is assumed-to-be-more-recent, ancestor. So this would find the oldest commit.

git-ext/src/reference/name.rs
```diff
@@ -69,6 +69,15 @@ impl RefLike {
         RefspecPattern(self.0.join(suf.into().0))
     }
 
+    /// Prepend a [`RefspecPattern`], yielding a [`RefspecPattern`]
```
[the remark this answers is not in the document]
FintanH: Ya, I like this a lot. It's a more broken down form. Good shout :ok_hand:

librad/src/identities/git.rs
```diff
@@ -174,8 +175,33 @@ impl<'a, T: 'a> Identities<'a, T> {
         root.verify(progeny)
     }
 
+    pub fn is_fork_generic(&self, left: git2::Oid, right: git2::Oid) -> Result<bool, git2::Error> {
```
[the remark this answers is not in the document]
FintanH: I fundamentally don't understand this, so I'm going to stop guessing :sweat_smile: Can we walk through detecting a fork in the history of an identity?

librad/src/identities/git.rs
FintanH: I want to ban all booleans from programming

librad/src/git/fetch.rs
```diff
@@ -403,8 +504,14 @@ impl<'a> DefaultFetcher<'a> {
 
             let mut callbacks = git2::RemoteCallbacks::new();
             callbacks.transfer_progress(|prog| {
-                tracing::trace!("Fetch: received {} bytes", prog.received_bytes());
-                true
+                let received_bytes = prog.received_bytes();
+                tracing::trace!("Fetch: received {} bytes", received_bytes);
+                if received_bytes < MAX_FETCH {
```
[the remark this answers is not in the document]
kim: > One question that came to mind though is: if a malkovich doesn't update a project for _a long time_ could they potentially be stuck in being unable to update the project because the delta is too big to fetch?

Ya... another reason to have different limits. The `rad/id` branches are not supposed to change very often. The `rad/signed_refs`, otoh, change all the time -- they will also not compress that well, because they contain hashes. Still, **a lot** of updates should fit into a few hundred KiB of packfile. Everything else... well, Linus' kernel tree is something like 2.5G.

librad/src/identities/git.rs
kim: :white_check_mark:

librad/src/identities/git.rs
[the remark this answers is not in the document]
kim: We have `content_id`, which is the commit hash, and `revision`, which is the tree hash. The _document_ blob described by `revision` contains the previous `revision`.

The reason for this is that it gives more flexibility for implementing in-band voting on updates -- if we had just commits, then everyone participating in the quorum must end up with the exact same commit. Which implies that either a leader is chosen who creates the final multisig commit, or that everyone performs (redundant) criss-cross merges with everyone else until all arrive at a fast-forward. Using trees, everyone can maintain their own commit history, while still arriving at the same _tree_ history.

librad/src/git/replication.rs
```diff
@@ -106,194 +128,599 @@ pub fn replicate<Addrs>(
...
+    #[derive(Clone, Debug)]
+    pub struct DelegateView {
+        pub urn: Urn,
+        pub delegate: VerifiedPerson,
+        pub project: VerifiedProject,
+    }
+
+    /// Process the setup of a `Project` by:
+    ///   * Verifying the identity
+    ///   * Ensuring there are no forks between the remote we are replicating
```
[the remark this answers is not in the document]
FintanH: This is a good point, and it's the switch in perspective I needed for improving a bit of logic that seems a bit "corner-case"-y to me. That is this scenario where we're adding the remote to the set regardless https://github.com/radicle-dev/radicle-link/pull/478/files#diff-fe1d1a45efacfdf43a7b387a93ab0db0b6eeab3e68b613e5f190fcbcdafb6ff0R271-R273.

The last thing to solve to remove this is the case of getting the state of the delegates https://github.com/radicle-dev/radicle-link/pull/478/files#diff-fe1d1a45efacfdf43a7b387a93ab0db0b6eeab3e68b613e5f190fcbcdafb6ff0R575. In the case of cloning we're relying on the remote's view to get setup the delegates in our monorepo. As you pointed out above, we shouldn't find the remote interesting (unless we track them). So in the question is **who** do we look at to get the (possibly) new state of the delegates? 

I'm thinking it's pretty arbitrary, we could just pick _one_ delegate remote to understand the new state of the world. How does that sit with you?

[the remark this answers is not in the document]
kim: > So in the question is **who** do we look at to get the (possibly) new state of the delegates?

You mean for the fetch case? This is where the TODO at the end of `replicate` kicks in: we know _some_ set of delegates, which we'd all update. Because they bring their own delegates (`rad/ids/*`), it will be possible to verify updates without knowing any **new** delegates yet. We can also check if the known set agrees (doesn't fork), and determine a most-recent one (to which we can update our own `rad/id`, **unless** we are ourselves part of the delegates set, in which case we need to prompt the user ... somehow).

However, if new delegates where introduced, we now need to track them. And technically also fetch, although this could be deferred to later.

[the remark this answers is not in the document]
FintanH: > You mean for the fetch case?

Yes, exactly. 

> This is where the TODO at the end of replicate kicks in: we know some set of delegates, which we'd all update.

Ahhh, I see I thought that was something to do with I/O interactions and being out of sync.

This PR is making an attempt to solve this case. I'll try to lay out the steps it takes:
1. Peek at known delegates and tracked peers
2. Choose one of those peers, the `trustee`, to get the delegate views -- doing the adopt and track dance
3. Ensure there's no forking against our view of the project
4. Track direct peers
5. Fetch signed refs
6. Track the signed ref peers
7. Adopt the latest 
8. Prune any delegates or tracked peers that were removed

The `trustee` part was what I was getting at in my question, and I added it in the latest commit https://github.com/radicle-dev/radicle-link/pull/478/commits/c8d33950536a6ec59bcb0b21ede86ad11930ec6d.

What you're saying above seems to be that we don't need step 2. in that list. We just go off of the delegates that we know about. Seems like I just need to rework that logic to work with the known delegates and then extend from there with the new/old delegate case (with the tricky case of being a delegate yourself).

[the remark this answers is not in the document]
FintanH: This should be closer to what we want https://github.com/radicle-dev/radicle-link/pull/478/commits/5414f964ac8ddd4e46668e250eac4318f7b71370. We only fetch what we know, and I added an `enum` for signaling when we're in the case of being a delegate and there's a mismatch between the latest tip and our tip. However, I defaulted to saying we're "behind", but I think we could be "ahead" as well. I still don't have a good mental model for this **tree** and **commit** business. I'd appreciate a session on this to flesh it out in my head a bit more if you have the time. I'm looking over your comments and the docs to see if I can piece it together.

FintanH:
Pong @kim

librad/src/git/replication.rs
```diff
@@ -102,198 +136,623 @@ pub fn replicate<Addrs>(
...
+        match tracking::untrack(storage, urn, *peer) {
+            Ok(removed) => {
+                if removed {
+                    tracing::info!("pruned `{}`", peer);
+                } else {
+                    tracing::trace!("attempted to prune `{}` but it did not exist", peer);
+                }
+            },
+            Err(err) => {
+                tracing::warn!("failed to prune `{}`\nreason: {}", peer, err);
```
kim, line 343: I think the loop should be aborted on the first error, as it's somewhat unlikely that the next iteration will succeed.

Also nit: newlines in log messages are probably not improving readability. I've settled on `xyz!(err = ?e, "somethingsomething error")` elsewhere.

librad/src/identities/git.rs
```diff
@@ -546,6 +572,17 @@ impl<'a> Identities<'a, Project> {
             .verified(parent.as_ref())?)
     }
 
+    pub fn is_fork(&self, left: &Project, right: &Project) -> Result<bool, error::Store> {
```
kim, line 575: Wondering if this needs to be monomorphic in `Person` / `Project` -- there is an implied invariant which is that `left` and `right` are actually `VerifiedIdentity<T>`.

librad/src/git/fetch.rs
```diff
@@ -30,13 +30,26 @@ use crate::{
...
+/// limiting the fetch of the `rad/*` references.
+pub const ONE_MB: usize = 1024 * 1024;
+
 /// Seed value to compute the fetchspecs for the desired fetch phase from.
 ///
 /// See also: [`super::replication::replicate`]
 #[derive(Debug)]
 pub enum Fetchspecs<P, R> {
+    /// Request all identity documents
+    All { max_fetch: usize },
```
kim, line 46:
```suggestion
    PeekAll { max_fetch: usize },
```

librad/src/git/fetch.rs
```diff
@@ -30,13 +30,26 @@ use crate::{
     peer::PeerId,
 };
 
+/// Upper bound of 1GB for fetching data. This upper bound is useful for when
+/// fetching from an unknown (and possibly untrusted) peer.
+pub const ONE_GB: usize = 1024 * 1024 * 1024;
+/// Upper bound of 1MB for fetching data. This upper bound is an estimate for
+/// limiting the fetch of the `rad/*` references.
+pub const ONE_MB: usize = 1024 * 1024;
```
kim, line 38: I think those numbers are too big -- the branches considered by `All` / `Peek` are supposed to be _**mostly**_ identical, so the resulting packfile should be a lot smaller than `ONE_MB * number_of_rad_id_branches` (or else, something's fishy).

Any chance we could bubble these up as runtime parameters?

librad/src/git/fetch.rs
```diff
@@ -30,13 +30,26 @@ use crate::{
...
 #[derive(Debug)]
 pub enum Fetchspecs<P, R> {
+    /// Request all identity documents
+    All { max_fetch: usize },
+
     /// Only request the branches necessary for identity verification.
-    Peek,
+    Peek {
+        remotes: BTreeSet<P>,
+        max_fetch: usize,
```
kim, line 51: I think we'd want a `max_fetch` for the other two variants as well.

kim (commented):
Mostly minor things left, but I think the verification precondition for the fork check should receive another pass.

librad/src/identities/git.rs
```diff
@@ -103,6 +103,22 @@ impl<'a, T: 'a> Identities<'a, T> {
...
+    }
+
+    /// Convenience to specialise `T` to [`VerifiedProject`].
+    pub fn as_verified_project(&self) -> Identities<'_, VerifiedProject> {
+        Identities {
+            repo: self.repo,
+            _marker: PhantomData,
+        }
+    }
+
```
kim, line 121: Nope -- casting to verified must go through verify!

kim: Oh wait, nevermind. I got the diff context wrong.

FintanH: :innocent:

librad/src/git/fetch.rs
```diff
@@ -30,24 +30,56 @@ use crate::{
...
+/// fetching from an unknown (and possibly untrusted) peer.
+pub const FIVE_HUNDRED_KB: usize = 1024 * 500;
+/// Relative upper bound of 1KB for fetching data. This bound is used in tandem
+/// with the number references that it will be fetching.
+pub const ONE_KB: usize = 1024;
+
+/// Limits used for guarding against fetching large amounts of data from the
+/// network.
+#[derive(Clone, Copy, Debug)]
+pub struct Limit {
```
kim, line 43: Hm, the `Replicate` variant needs a much higher limit (like, 3GB, or even 5). I think if we have a peeking limit (regardless of the number of references), a sigrefs limit, and a data limit, it would be easier to reason about it.

librad/src/identities/git.rs
```diff
@@ -379,6 +410,24 @@ where
     }
 }
 
+impl<'a, T: 'a> Identities<'a, VerifiedIdentity<T>> {
+    #[tracing::instrument(level = "debug", skip(self, mine, theirs), err)]
```
kim, line 414: Suggestion: use `fields()` in the `instrument` macro to establish the `content_id`s / `revision`s in the span scope, then log just the strings.

librad/src/git/fetch.rs
FintanH: So I just wanna make sure I got this right, and also follow up with a couple of questions.

We have 3 limits:
1. `peek` that is used in `PeekAll` and `Peek`.
2. `signed_refs` that is used in `PeekAll` -- since we fetch signed refs here too -- and `SignedRefs`
3. `data` that is used in `Replicate`

One question that I have is how should we scale the factor of `peek` and `signed_refs` in the case of `PeekAll`? Do we just have a constant factor of "expect at most 20 remotes"? Should this also be a runtime parameter?

In the case of `data` should it be scaled by the tracked refs and delegates or is it a flat limit, i.e. is it 5GB or 5GB * # of refs? I guess the answer boils down to "Are we potentially fetching 5GB from each of these delegates?", which I don't have a good intuition for.

kim: >     2\. `signed_refs` that is used in `PeekAll` -- since we fetch signed refs here too -- and `SignedRefs`

Ah, hm. So why are we fetching `SignedRefs` separately still? :thinking: 

> should we scale the factor

What I'm thinking is that we shouldn't do this at all -- every fetch makes the other side assemble a single packfile, but it is not predictable how large it will be. In the case of `id`-related branches, those should only ever differ by at most a few commits. The same is more or less true for `data` -- say you have a repo with a long history, where fetching the `master` branch would clock at 1GB already. Even if it has a lot of concurrent contributions, it is rather unlikely that those would diverge by gigabytes (or else, something is a bit pathological about that project -- if you still want it, we should provide a knob to raise the limit for that specific repo).

The remaining is `signed_refs`, which is more tricky: every peer has their own, and they may differ substantially between peers. Otoh, every update should be very small, and a single peer's history should delta-compress well. So idk, maybe that'd be a reason to keep fetching `signed_refs` in a separate step, so we can adjust the limit acc. to "real world" numbers.

FintanH: > Ah, hm. So why are we fetching SignedRefs separately still? thinking

Actually, just realising that I added fetching the `singed_refs` to `Peek` and `SignedRefs` isn't actually used :facepalm: I removed this in the latest change.

> if you still want it, we should provide a knob to raise the limit for that specific repo

Right, this makes it more clear to me. Imagining a field where you can jack up the limit and try again.

I set a 5KB limit for peek and 5GB limit for data in the latest change.

FintanH:
Any other improvements I could make here? I feel confident in what's there now :)

librad/src/git/replication.rs
```diff
@@ -102,198 +136,678 @@ pub fn replicate<Addrs>(
...
+    identities::common::IdRef::from(urn)
+        .create(storage, tip)
+        .map_err(|e| Error::Store(e.into()))
+}
+
+/// Untrack the list of `PeerId`s, which also has the side-effect of removing
+/// that peer's remote references in the storage.
+///
+/// In the case of an error occurring during the untrack call we emit a warning
+/// and append it to a `Vec` of unpruned peers. The unpruned peers are returned.
```
kim, line 338:
```suggestion
```

This is no longer true. We can introduce it once we need it.

kim (approved):
Alright, let's get this in. Thanks a lot for the effort put into this?

(Do you want to put FIXMEs into issues?)

FintanH:
> Alright, let's get this in. Thanks a lot for the effort put into this!

Appreciate it, and your guidance :)

> (Do you want to put FIXMEs into issues?)

Yup, can do!

FintanH:
Closed via https://github.com/radicle-dev/radicle-link/commit/087a1bf48f5b5802878c42986b874fff7cf9b4c0

PR #483, opened 2021-01-13
Fix broadcast local storage provider

kim:
Instead of passing the peer id of the sender, we need to pass the origin, which
preserves relaying semantics. This also means that we actually have address
hints available.

### DIFF
--- a/librad/src/net/peer/storage.rs
+++ b/librad/src/net/peer/storage.rs
@@ -3,8 +3,9 @@
 // This file is part of radicle-link, distributed under the GPLv3 with Radicle
 // Linking Exception. For full terms see the included LICENSE file.
 
-use either::Either::{self, Left, Right};
+use std::net::SocketAddr;
 
+use either::Either::{self, Left, Right};
 use git_ext::{self as ext, reference};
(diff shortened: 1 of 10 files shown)


kim:
ping @FintanH @xla

librad/src/net/peer/storage.rs
```diff
@@ -131,39 +144,36 @@ impl broadcast::LocalStorage for Storage {
...
+            match self
+                .git_fetch((provider, addr_hints), urn.clone(), head)
+                .await
+            {
                 Ok(()) => {
+                    // Verify that the announced data is stored locally now.
+                    //
+                    // If it is, rewrite the gossip message to use the `origin`
+                    // we determined -- everyone down the line may now fetch
+                    // the that remote from us.
```
FintanH, line 162:
```suggestion
                    // the remote from us.
```

FintanH (approved):
Looks good :+1: One grammar fix to the comment :)

kim (commented):
TFW your rebase undoes the exact change the patch is all about :facepalm:

PR #484, opened 2021-01-13
Include more refs categories in signed_refs

kim:
Namely:

* refs/rad
* refs/tags
* refs/notes

The first one is necessary for implementing "radicle applications",
while the other two are more or less for the giggles for now.

### DIFF
--- a/librad/src/git/refs.rs
+++ b/librad/src/git/refs.rs
@@ -201,10 +201,25 @@ pub mod stored {
     }
 }
 
-/// The current `refs/heads` and [`Remotes`] (transitive tracking graph)
+/// The published state of a local repository.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Refs {
+    /// `refs/heads/*`
     pub heads: BTreeMap<reference::OneLevel, Oid>,
+
+    /// `refs/rad/*`, exluding `refs/rad/signed_refs`
+    pub rad: BTreeMap<reference::OneLevel, Oid>,
+
+    /// `refs/tags/*`
+    pub tags: BTreeMap<reference::OneLevel, Oid>,
+
+    /// `refs/notes/*`
+    pub notes: BTreeMap<reference::OneLevel, Oid>,
+
+    /// The [`Remotes`], ie. tracking graph.
+    ///
+    /// Note that this does does not include the oids, as they can be determined
+    /// by inspecting the `rad/signed_refs` of the respective remote.
     pub remotes: Remotes<PeerId>,
 }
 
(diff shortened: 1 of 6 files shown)


librad/src/git/refs.rs
```diff
@@ -214,29 +229,44 @@ impl Refs {
...
-            })?;
+        let heads = storage
+            .references(&Reference::heads(namespace.clone(), None))?
+            .filter_map(peeled)
+            .map(refined)
+            .collect::<Result<_, _>>()?;
+        let rad = storage
+            .references(&Reference::rads(namespace.clone(), None))?
+            .filter_map(peeled)
+            .filter(|(name, _)| !name.ends_with("rad/signed_refs"))
```
kim, line 257: I'm unsure if this should also filter out `rad/id`. `rad/ids/*` :thinking:

librad/src/git/types/reference.rs
```diff
@@ -367,6 +370,39 @@ impl<N, R> Reference<N, R, Many> {
             namespace: namespace.into(),
         }
     }
+
+    /// Build a reference that points to:
+    ///     * `refs[/namespaces/<namespace>]/refs[/remotes/<remote>]/rad/*`
```
FintanH, line 375: The same suggestions go for the rest of them, I believe. We would have `refs/refs` if the namespace was optional :)

librad/src/git/refs.rs
```diff
@@ -201,10 +201,25 @@ pub mod stored {
...
 }
 
-/// The current `refs/heads` and [`Remotes`] (transitive tracking graph)
+/// The published state of a local repository.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Refs {
+    /// `refs/heads/*`
     pub heads: BTreeMap<reference::OneLevel, Oid>,
+
+    /// `refs/rad/*`, exluding `refs/rad/signed_refs`
```
FintanH, line 210:
```suggestion
    /// `refs/rad/*`, excluding `refs/rad/signed_refs`
```

librad/src/git/refs.rs
FintanH: What's the usage of `Refs`? How useful is it for those to be included or would always just access them by building the reference ourselves, i.e. `Reference::rad_id`, `Reference::rad_ids`?

Or in another light, what's the purpose of the `rad/*` references in here? Is it for the extension of new application level branches, e.g. the revisions?

FintanH (commented):
Looks great, I appreciate the splitting out into functions.

There are some minor doc fixes

librad/src/git/types/reference.rs
```diff
@@ -346,8 +349,8 @@ impl<N, R> Reference<N, R, Many> {
         ext::References::from_globs(repo, &[self.to_string()])
     }
 
-    /// Build a reference that points to
-    /// `refs/namespaces/<namespace>/refs/rad/ids/*`
+    /// Build a reference that points to:
+    ///     * `refs[/namespaces/<namespace>]/refs/rad/ids/*`
```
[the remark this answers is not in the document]
kim: oof. I tried adding doctests, but UNFORTUNATELY they don't have access to `#[cfg(test)]` code (ie. the `FakeId` for creating URNs).

librad/src/git/refs.rs
kim: We only ever fetch the refs of peer `A` which are listed in the `signed_refs`, which must be signed by `A`, and where the Oids match the advertised ones. Recall: we could be fetching `A`s refs from untrusted `B` (ie. `refs/remotes/A/*`), so if we didn't do this, `B` could give us anything.

I'm including `rad/*` here because if we want to namespace "special" refs there, they need to get replicated. Clearly, we can't include `rad/signed_refs`, because the hash will change after we committed the result of `compute`. For everything else, the question is would it interfere with anything? We fetch `rad/id` and `rad/ids/*` first (without looking at `rad/signed_refs`), so it should not have any effect. _Strictly speaking_, we could omit them, thus. But I'm wondering if it would make things more robust against future changes if we have less exceptions from the rule.

librad/src/git/types/reference.rs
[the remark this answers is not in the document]
FintanH: Doesn't seem to be getting anywhere either https://github.com/rust-lang/rust/issues/45599. I think I just ended up repeating myself in `radicle-surf` for the sake of having comprehensive examples

librad/src/git/refs.rs
FintanH: > But I'm wondering if it would make things more robust against future changes if we have less exceptions from the rule.

I'm gravitating towards this sentiment more. Let's not have edge-cases that we will have to keep track of and workaround. It's easier to filter any unwanted `rad` refs downstream than it is to add them back in.

FintanH (approved):
`refs/approvals/fintohaps`

librad/src/git/refs.rs
kim: For correctness, we would need to trigger a `Refs::update` after messing with the identity, tho :thinking:

PR #485, opened 2021-01-13
Fix URN in merge code

FintanH:
The urn::DEFAULT_PATH has the value of `refs/rad/id`. When we want to
look at a remote's `rad/id` we have to remove the refs` prefix.

### DIFF
--- a/git-ext/src/reference/name.rs
+++ b/git-ext/src/reference/name.rs
@@ -281,18 +281,14 @@ impl OneLevel {
     }
 
     pub fn from_qualified(Qualified(path): Qualified) -> (Self, Option<RefLike>) {
-        if path.starts_with("refs/") {
-            let mut path = path.iter().skip(1);
-            match path.next() {
-                Some(category) => {
-                    let mut cat = PathBuf::new();
-                    cat.push(category);
-                    (Self(path.collect()), Some(RefLike(cat)))
-                },
-                None => (Self(path.collect()), None),
-            }
-        } else {
-            (Self(path), None)
+        // skip "refs/"
+        let mut path = path.iter().skip(1);
+        match path.next() {
+            Some(category) => {
+                let category = Path::new(category).to_path_buf();
+                (Self(path.collect()), Some(RefLike(category)))
+            },
+            None => (Self(path.collect()), None),
         }
     }
 
--- a/git-ext/src/reference/name.rs
+++ b/git-ext/src/reference/name.rs
@@ -263,6 +263,11 @@ impl Display for RefLike {
 /// );
 ///
 /// assert_eq!(
+///     OneLevel::from_qualified(Qualified::from(RefLike::try_from("refs/HEAD").unwrap())),
+///     (OneLevel::from(RefLike::try_from("HEAD").unwrap()), None)
+/// );
+///
+/// assert_eq!(
 ///     &*OneLevel::from(RefLike::try_from("origin/hopper").unwrap()).into_qualified(
 ///         RefLike::try_from("remotes").unwrap()
 ///     ),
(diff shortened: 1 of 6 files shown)


kim (commented):
Yeah, I think the path also needs to start with `reflike!("refs/remotes")`, as you said.

Wondering if we could refine the types a bit. For example, `urn::DEFAULT_PATH` could actually be typed `Qualified`, requiring explicit conversion at the use site. Conversion to a `OneLevel` could also be typed `a -> (OneLevel, Maybe String)`, preserving the category which has been stripped (if any). Symmetrically, we could also provide construction of `Qualified` which supplies the category (eg. `OneLevel -> cat -> Qualified`). Wdyt?

FintanH:
> Conversion to a OneLevel could also be typed a -> (OneLevel, Maybe String), preserving the category which has been stripped (if any).

Even better would be to re-use `RefsCategory`, maybe adding a case for an ad-hoc category appearing. 

> Symmetrically, we could also provide construction of Qualified which supplies the category (eg. OneLevel -> cat -> Qualified).

I like symmetry :)

kim:
> Even better would be to re-use `RefsCategory`, maybe adding a case for
> an ad-hoc category appearing.

Ye-aah... `RefsCategory` is link-specific, while `OneLevel` is not. Yolo is
easy, while modularity is not.

FintanH:
> Ye-aah... `RefsCategory` is link-specific, while `OneLevel` is not. Yolo is
easy, while modularity is not.

Ah good point, well we should at least guarantee it to be `RefLike` rather than `String` :innocent:

git-ext/src/reference/name.rs
```diff
@@ -256,6 +279,26 @@ impl OneLevel {
     pub fn as_str(&self) -> &str {
         self.as_ref()
     }
+
+    pub fn from_qualified(Qualified(path): Qualified) -> (Self, Option<RefLike>) {
+        if path.starts_with("refs/") {
+            let mut path = path.iter().skip(1);
+            match path.next() {
```
FintanH, line 286: This match highlights the fact that `Qualified` can still be in a weird form. It could be that it has `refs/mistress` so there's a category but no path. Any thoughts on resolving this or do you think it's worth resolving? :thinking:

kim: > /me jumps around, hysterically chanting "do what I mean! do what I mean!"

I think this is great as-is, because the semantics of git refs is really ... in the eye of the beholder. For example `OneLevel("HEAD")` is valid, as is `Qualified("refs/HEAD")`, but its semantics depend on how you look at it (conventionally, `HEAD` would be expected to be a symref). If we wanted to capture all permutations in the type system in a general way, we would need to be able to express something like this:

```haskell
type QualifiedAndCategorised a = Ref '["refs"] ++ Category a ++ (xs :: '(NonEmpty PathComponent));
type QualifiedSym = Ref '["refs", Uppercase];
type Category a = a | '["remotes", Remote, a];
type GitCategory = "heads" | "tags" | "notes" | "replace" | "bisect" | SomeCategory;
type RadCategory = (GitCategory ∉ "replace" | "bisect" | SomeCategory) | "rad"
```

Not sure it's possible to approximate that in the type system we have, and even in a more powerful one it wouldn't be trivial. So I think we should probably look into refining types in `librad` where we understand what forms we expect / allow.

You don't need the `if else`, though: a `Qualified` _always_ starts with "refs/".

FintanH: > hysterically chanting "do what I mean! do what I mean!"

Ahahaha fair enough. But then WDYM when you have a `Qualified(refs/HEAD)` and we call `from_qualified`? Should it be `(OneLevel(HEAD), None)`? If that's the case my function is wrong.

Really, we shouldn't be able to hit the `None` case because we should have `refs/` followed by some value. Do we prefer a `panic!` or `unreachable!` for these cases?

kim: The version in 4297476 looks good to me

kim (approved):
:+1:

FintanH:
Closed via https://github.com/radicle-dev/radicle-link/commit/285bdfee6956084c955c079a3789b4ac6468f7d2

PR #487, opened 2021-01-15
net: permit logical network identifiers in ALPN

kim:
For operating "devnets" without physical network isolation.

Also, spell out the protocol versioning policy, and bump the version as
`net/next` is incompatible.

### DIFF
--- a/librad/src/net.rs
+++ b/librad/src/net.rs
@@ -13,3 +13,43 @@ pub mod upgrade;
 pub mod x509;
 
 mod codec;
+
+/// The network protocol version.
+///
+/// The version number denotes compatibility, and as such is subject to change
+/// very infrequently: if two peers advertise the same version number, they MUST
+/// be able to communicate with each other (ie. their implementations must be
+/// compatible both forward and backward).
+///
+/// The protocol version is negotiated during the handshake via [ALPN], which
+/// permits gradual rollout scenarios of major network upgrades.
+///
+/// For the negotiation of optional (compatible _per definitionem_) protocol
+/// features, the [`PeerAdvertisement`] reserves space for advertising
+/// [`info::Capability`]s.
+///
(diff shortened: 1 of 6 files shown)


kim:
Note that this is against `net/next`. I'm procrastinating a backport to `next` until I hear that the "Dev seed node" card is actively being worked on (because, maybe `net/next` is in `next` already until then). /cc @rudolfs

xla:
> Note that this is against `net/next`. I'm procrastinating a backport to `next` until I hear that the "Dev seed node" card is actively being worked on (because, maybe `net/next` is in `next` already until then). /cc @rudolfs

Looking at the timeline given current priority next will land before the dev seed nodes will even start, so a backport doesn't seem to be impactful.

librad/src/net.rs
```diff
@@ -13,3 +13,42 @@ pub mod upgrade;
...
+///
+/// The protocol version is negotiated during the handshake via [ALPN], which
+/// permits gradual rollout scenarios of major network upgrades.
+///
+/// For the negotiation of optional (compatible _per definitionem_) protocol
+/// features, the [`PeerAdvertisement`] reserves space for advertising
+/// [`info::Capability`]s.
+///
+/// [ALPN]: https://tools.ietf.org/html/rfc7301
+pub const PROTOCOL_VERSION: u8 = 2;
```
xla, line 32: Vorsprung durch Technik.

xla (approved):
LGTM. I wonder if there is a necessity for a "default" (i.e. `Main`) network or if node creation always expect a custom identifier.

🍣 🗑 🍬 🗝

kim:
Maybe I misread what the devseed is about: network identifiers are useful for "field testing". As long as everything binds to the loopback interface, it's not needed. Does this change anything?

> if node creation always expect a custom identifier.

@xla could you expand on this?

kim (commented):
Rebased onto `net/next`

PR #489, opened 2021-01-19
Introduce Storage Tests

### LINKED ISSUE #475: Listing project with PublicKey delegation
When using `project::create` with a `PublicKey` and following that up with a call to `any::list`, that `Project` will fail to deserialize.
This was found while working on `radicle-upstream` so it would be good to create a test case for this to confirm that it is the case and, if it is, fix the bug.

FintanH:
Fixes #475 

Feeding two birds with one scone.

1. Introducing the place for tests that involve `Storage` and the `identities` APIs, since this is the main entry point for applications.
2. Fixes a case where a project can be set up using only `PublicKey`s but cannot be retrieced through the `Any` deserializer.

@kim I held off on sharing some of the resources from the other set of `identities::git::tests` (as opposed to these new `git::tests`). What are your thoughts on moving some of the common stuff upwards in the module system?

### DIFF
--- a/librad/src/git.rs
+++ b/librad/src/git.rs
@@ -17,6 +17,9 @@ pub mod types;
 
 mod sealed;
 
+#[cfg(test)]
+pub(crate) mod tests;
+
 pub use crate::identities::git::Urn;
 
 /// Initialise the git backend.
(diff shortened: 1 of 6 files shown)


kim (approved):

FintanH:
Closed via https://github.com/radicle-dev/radicle-link/commit/3e56f9852d212833d089a99b0975cae562a0949f

PR #490, opened 2021-01-19
net: tickle connections after evaluating periodic tasks

kim:
See commentary inline. We need to keep passive peers from hitting the
idle timeout, on a best-effort basis.


FintanH (approved):

kim:
Closed via 68900e9

PR #493, opened 2021-01-21
e2e: basic setup

kim:
Rudimentary automation for spawning networks onto the loopback
interface.

PR #494, opened 2021-01-21
net: tickle connection in AsyncWrite::poll_close

kim:
Not exactly sure why that is, but otherwise the network will become
disconnected after about 10min due to collecting idle connections
(detected via #493).


xla (approved):
💪 🖨 🕑 🍄

FintanH (approved):

Issue #488, opened 2021-01-18
Profiles and profile-scoped paths


geigerzaehler:
We want to support multiple user profiles with different identities and Git data.

Profiles are identified by an opaque string. (We will use UUIDs as identifiers). The active profile is stored as a string in `$XDG_CONFIG_HOME/radicle/active_profile`on Linux. Profile specific data is stored in `$XDG_{CONFIG|DATA}_HOME/radicle/<profile_id>`. For example, the key directory will be  `$XDG_{CONFIG|DATA}_HOME/radicle/<profile_id>/keys`. For other platforms the location is similarly scoped by inserting the profile identifier.

If the `active_profile` file does not exist, it is created, a new profile identifier is generated and written to the file.

Profiles and the new paths will be used by `git-remote-rad` and `radicle-upstream`.

If possible, we will provide functionality to migrate the paths.

### APIs

We introduce a new `Profile` struct with the following methods
~~~rust
impl Profile {
   /// Tries the profile identifier from a well known path on disk. If the file does not exist, it
   /// is created, a new identifier is generated, and written to the file.
   fn init() -> Result<Self, io::Error>  {
       todo!()
   }

   /// Returns the profile identifier
   fn id(&self) -> &str {
       todo!()
   }

   /// Returns [`Paths`] for this profile and creates all the directories.
   fn paths(&self) -> Result<Paths, io::Error> {
       todo!()
   }
}
~~~

We remove the `Paths::new()` from the public API so that `Paths` can only be obtained through `Profile::paths`.

The directory paths returned by `Paths` will include the profile identifier as specified above.

### Alternatives

We considered using the identity as the profile identifier but this has a couple of shortcomings: It requires us to generate an identity before we can initialize a profile which complicates a lot of code. This approach also precludes us from having multiple profiles with the same identity. One advantage of using identities to identify profiles is that it would make profile selection easier for the user. But this can also be solved through better tooling, especially given that it is not clear yet how the user would select different profiles.

cloudhead:
Hey could you say a few words about how this will be used?

dragon788:
How does this work on a multi-user system, each user if they have multiple profiles would have them exist under their own XDG directory, but there wouldn't be coordination sharing via the system level XDG (/usr/share/ or similar?).

geigerzaehler:
> Hey could you say a few words about how this will be used?

[...]

PR #496, opened 2021-01-25
Add profiles and profile-scoped paths

geigerzaehler:
To address #488 we add the `Profile` struct.

This pull request is against `next`. I will try to write automatic migration code for this so we can include it in `maint` without a breaking change.

We’ve also simplified some methods on `Paths` and made them private.

### DIFF
--- a/librad/src/paths.rs
+++ b/librad/src/paths.rs
@@ -4,8 +4,6 @@
 // Linking Exception. For full terms see the included LICENSE file.
 
 use std::{
-    collections::HashMap,
-    env,
     fs,
     io,
     path::{Path, PathBuf},
@@ -13,7 +11,12 @@ use std::{
 
 use directories::ProjectDirs;
 
-#[derive(Clone)]
+/// A set of paths to store application data.
+///
+/// The paths are either based on system specific directories when created
+/// with [`crate::profile::Profile::paths`] or all contained in a given
+/// directory when created with [`Paths::from_root`].
+#[derive(Debug, Clone)]
 pub struct Paths {
     keys_dir: PathBuf,
     git_dir: PathBuf,
@@ -21,16 +24,14 @@ pub struct Paths {
 }
 
 impl Paths {
-    pub fn new() -> Result<Self, io::Error> {
-        let proj = ProjectDirs::from("xyz", "radicle", "radicle").ok_or_else(|| {
-            io::Error::new(
-                io::ErrorKind::NotFound,
-                "Couldn't determine application directories.",
-            )
-        })?;
-
-        let config_dir = proj.config_dir();
-        let data_dir = proj.data_dir();
+    /// Returns based on [`ProjectDirs`] and scoped to `profile_id`.
+    ///
+    /// On Linux, all paths start with
+    /// `$XDG_{CONFIG|DATA}_HOME/radicle-link/<profile_id>`.
+    pub(crate) fn new(profile_id: &str) -> Result<Self, io::Error> {
+        let proj = project_dirs()?;
+        let config_dir = proj.config_dir().join(profile_id);
+        let data_dir = proj.data_dir().join(profile_id);
 
         Self {
             keys_dir: config_dir.join("keys"),
@@ -40,9 +41,7 @@ impl Paths {
         .init()
     }
 
-    // Don't use system paths, but the supplied directory as a root.
-    //
-    // For testing, you know.
+    /// All paths are contained in the given directory.
     pub fn from_root(root: impl AsRef<Path>) -> Result<Self, io::Error> {
         let root = root.as_ref();
         Self {
@@ -51,13 +50,6 @@ impl Paths {
             git_includes_dir: root.join("git-includes"),
         }
         .init()
-    }
-
-    pub fn from_env() -> Result<Self, io::Error> {
-        match env::var_os("RAD_HOME") {
-            None => Self::new(),
-            Some(root) => Self::from_root(root),
-        }
     }
 
     pub fn keys_dir(&self) -> &Path {
@@ -72,7 +64,7 @@ impl Paths {
         &self.git_includes_dir
     }
 
-    pub fn all_dirs(&self) -> HashMap<&str, &Path> {
+    fn all_dirs(&self) -> impl Iterator<Item = &Path> {
         // Nb. this pattern match is here to keep the map consistent with the
         // struct fields
         let Self {
@@ -81,20 +73,31 @@ impl Paths {
             git_includes_dir,
         } = self;
 
-        [
-            ("keys_dir", keys_dir.as_path()),
-            ("git_dir", git_dir.as_path()),
-            ("git_includes_dir", git_includes_dir.as_path()),
+        vec![
+            keys_dir.as_path(),
+            git_dir.as_path(),
+            git_includes_dir.as_path(),
         ]
-        .iter()
-        .cloned()
-        .collect()
+        .into_iter()
     }
 
     fn init(self) -> Result<Self, io::Error> {
-        self.all_dirs().values().try_for_each(fs::create_dir_all)?;
+        self.all_dirs().try_for_each(fs::create_dir_all)?;
         Ok(self)
     }
+}
+
+/// Returns [`ProjectDirs`] for this specific project (`radicle`).
+///
+/// Returns [`io::Error`] if the project directories could not be determined,
+/// most likely due to the `$HOME` environment variable missing
+pub(crate) fn project_dirs() -> Result<ProjectDirs, io::Error> {
+    ProjectDirs::from("xyz", "radicle", "radicle-link").ok_or_else(|| {
+        io::Error::new(
+            io::ErrorKind::NotFound,
+            "Couldn't determine application directories.",
+        )
+    })
 }
 
 #[cfg(test)]
@@ -108,7 +111,7 @@ mod tests {
     fn test_initialises_paths() {
         let tmp = tempdir().unwrap();
         let paths = Paths::from_root(tmp.path()).unwrap();
-        assert!(paths.all_dirs().values().all(|path| path.exists()))
+        assert!(paths.all_dirs().all(|path| path.exists()))
     }
 
     /// Test we indeed create everything under the root dir -
@@ -119,7 +122,6 @@ mod tests {
         let paths = Paths::from_root(tmp.path()).unwrap();
         assert!(paths
             .all_dirs()
-            .values()
             .all(|path| { path.ancestors().any(|parent| parent == tmp.path()) }))
     }
 }
--- a/librad/src/profile.rs
+++ b/librad/src/profile.rs
@@ -1,0 +1,245 @@
+// Copyright © 2019-2021 The Radicle Foundation <hello@radicle.foundation>
+//
+// This file is part of radicle-link, distributed under the GPLv3 with Radicle
+// Linking Exception. For full terms see the included LICENSE file.
+
+use std::{
+    env,
+    fs,
+    io,
+    path,
+    path::{Path, PathBuf},
+};
+use uuid::Uuid;
+
+use crate::paths::{project_dirs, Paths};
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("invalid profile ID in RAD_PROFILE environment variable: {id}")]
+    InvalidProfileIdFromEnv { id: String },
+    #[error("invalid profile ID loaded from {path}: {id}")]
+    InvalidProfileIdFromFile { id: String, path: PathBuf },
+    #[error(transparent)]
+    Io(#[from] io::Error),
+}
+
+/// A [`Profile`] provides [`Paths`] scoped by an identifier.
+///
+/// Profiles with different identifiers have distinct paths.
+#[derive(Debug, Clone)]
+pub struct Profile {
+    id: String,
+    paths: Paths,
+}
+
+impl Profile {
+    /// Creates a profile by loading the profile identifier and paths from
+    /// the environment variables or well-known file.
+    ///
+    /// By default, the profile identifier is read from the `active_profile`
+    /// file in a system specific location. The paths returned by
+    /// [`Profile::paths`] are based on system paths (see [`directories::
+    /// ProjectDirs`]) and include the profile identifier. If the file
+    /// containing the identifier does not exist, it is created and a new
+    /// identifier is generated and written to the file.
+    ///
+    /// If the `RAD_PROFILE` environment variable is set, its value is used as
+    /// the profile identifier. The `active_profile` file is ignored.
+    ///
+    /// If the `RAD_HOME` environment variable is set, its value is used instead
+    /// of system specific project directories. See also
+    /// [`Profile::from_root`].
+    ///
+    /// The profile identifier must not empty, must not contain path separators,
+    /// must not be a windows path prefix like `C:`, and must not be a
+    /// special component like `.` or `..`. Otherwise an error is returned
+    ///
+    /// On Linux, the path to the active profile is
+    /// `$XDG_CONFIG_HOME/radicle-link/active_profile` and profile specific
+    /// system paths are `$XDG_CONFIG_HOME/radicle-link/<profile_id>` and
+    /// `XDG_DATA_HOME/radicle-link/<profile-id>`.
+    pub fn load() -> Result<Self, Error> {
+        let env_profile_id = env::var("RAD_PROFILE").ok();
+
+        if let Some(ref id) = env_profile_id {
+            if !is_valid_profile_id(id) {
+                return Err(Error::InvalidProfileIdFromEnv { id: id.clone() });
+            }
+        }
+
+        if let Ok(rad_home) = env::var("RAD_HOME") {
+            Self::from_root(Path::new(&rad_home), env_profile_id)
+        } else {
+            let id = if let Some(id) = env_profile_id {
+                id
+            } else {
+                load_profile_id(project_dirs()?.config_dir())?
+            };
+
+            let paths = Paths::new(&id)?;
+
+            Ok(Self { id, paths })
+        }
+    }
+
+    /// Creates a profile where `<root>/<profile_id>` is used as the base path
+    /// for profile specific data.
+    ///
+    /// If `profile_id` is `None`, then the profile is read from
+    /// `<root>/active_profile` if the file exists. Otherwise, a new profile
+    /// ID is generated and written to `<root>/active_profile`.
+    ///
+    /// [`Paths::from_root`] for more information.
+    pub fn from_root(root: &Path, profile_id: Option<String>) -> Result<Self, Error> {
+        let id = if let Some(profile_id) = profile_id {
+            profile_id
+        } else {
+            load_profile_id(root)?
+        };
+
+        let paths = Paths::from_root(root.join(&id))?;
+
+        Ok(Self { id, paths })
+    }
+
+    /// Returns the profile identifier
+    pub fn id(&self) -> &str {
+        &self.id
+    }
+
+    /// Returns [`Paths`] for this profile.
+    pub fn paths(&self) -> &Paths {
+        &self.paths
+    }
+}
+
+/// Returns `true` if `id` is a valid profile ID.
+///
+/// A valid profile ID is not empty, does not contain path separators, is not
+/// a windows path prefix like `C:`, and is not a special component like `.` or
+/// `..`.
+fn is_valid_profile_id(id: &str) -> bool {
+    let mut components = Path::new(id).components();
+
+    match components.next() {
+        Some(path::Component::Normal(_)) => {},
+        _ => return false,
+    }
+
+    if components.next().is_some() {
+        return false;
+    }
+
+    true
+}
+
+/// Read the profile ID from a file or create the file and write a newly
+/// generated profile ID to it.
+///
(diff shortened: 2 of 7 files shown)


librad/src/profile.rs
```diff
@@ -0,0 +1,112 @@
...
+use uuid::Uuid;
+
+use crate::paths::{project_dirs, Paths};
+
+/// A [`Profile`] provides [`Paths`] scoped by an identifier.
+///
+/// Profiles with different identifiers have distinct paths.
+#[derive(Debug, Clone)]
+pub struct Profile {
+    id: String,
```
FintanH, line 16: Would there be any advantage to using the `Uuid` type here instead? Or even having a generic parameter for switching between different identity types in the future or for testing?

librad/src/profile.rs
```diff
@@ -0,0 +1,112 @@
...
+    paths: Paths,
+}
+
+impl Profile {
+    /// Creates a profile by loading the the profile identifier and paths from
```
FintanH, line 21:
```suggestion
    /// Creates a profile by loading the profile identifier and paths from
```

librad/src/profile.rs
```diff
@@ -0,0 +1,112 @@
...
+    /// the environment variables or well-known file.
+    ///
+    /// If `RAD_PROFILE` is not set, the value is used as the profile identifier
```
FintanH, line 24: Should this be, "If `RAD_PROFILE` is set"? Otherwise, what "value" are you referring to?

librad/src/paths.rs
```diff
@@ -72,7 +63,7 @@ impl Paths {
         &self.git_includes_dir
     }
 
-    pub fn all_dirs(&self) -> HashMap<&str, &Path> {
+    fn all_dirs(&self) -> impl Iterator<Item = &Path> {
```
FintanH, line 66: What's the reason for this change?

FintanH (commented):
Looking good. I just have a few details I'd like to clear up :)

librad/src/profile.rs
geigerzaehler: I don’t see a strong need to make this generic. At the moment he tests don’t benefit from it and I can’t imagine that in the future we need a profile with two different ID parameters. If that happens though, it’s not to late to change the code.

Regarding using `Uuid` instead of `String`. If we implement this than only UUIDs are valid profile identifiers. This is a limitation that may make the feature harder to user. For example, say I’m playing around with this using the `RAD_PROFILE` environment variable. Since UUIDs are quite opaque, I’m free to use `RAD_PROFILE=foo` and radicle will deal with this. We only use `UUIDs` to generate unspecified profiles to ensure they are unique.

FintanH: Makes sense :+1:

librad/src/paths.rs
geigerzaehler: I’ve removed `pub` because this method is unused by external consumers and keeping the public API small makes it easier to grasp and harder to inadvertently break.

I’ve replaced `HashMap` with the iterator because for consumers it provides all it needs and it simplifies the implementation.

I’m ok with reverting all those secondary changes. They are not related to the functionality but seemed like nice improvements.

librad/src/profile.rs
geigerzaehler: Yes, of course! I’ve fixed this.

librad/src/paths.rs
```diff
@@ -81,22 +72,33 @@ impl Paths {
...
-        self.all_dirs().values().try_for_each(fs::create_dir_all)?;
+        self.all_dirs().try_for_each(fs::create_dir_all)?;
         Ok(self)
     }
 }
 
+/// Returns [`ProjectDirs`] for this specific project (`radicle`).
+///
+/// Returns [`io::Error`] if the project directories could not be determined,
+/// most likely du to the `$HOME` environment variable missing
```
kim, line 92:
```suggestion
/// most likely due to the `$HOME` environment variable missing
```

librad/src/paths.rs
```diff
@@ -81,22 +72,33 @@ impl Paths {
...
+pub(crate) fn project_dirs() -> Result<ProjectDirs, io::Error> {
+    ProjectDirs::from("xyz", "radicle", "radicle").ok_or_else(|| {
```
kim, line 94: Do we want to take the opportunity to change this to `xyz`, `radicle`, `radicle-link`? I'd think that migration logic would be greatly simplified, too.

librad/src/profile.rs
```diff
@@ -0,0 +1,112 @@
...
+    /// If `RAD_PROFILE` is set, the value is used as the profile identifier
+    /// and system paths, scoped to the identifier, are used. See
+    /// [`Paths::new`] for details.
+    ///
+    /// Otherwise and if the `RAD_HOME` environment variable is set, the value
+    /// is used to construct the profile with [`Profile::from_root`].
+    ///
+    /// Otherwise, the profile identifier is read from a well-known file and
+    /// system paths scoped to the identifier are used. If the file containing
+    /// the identifier  does not exist, it is created and new identifier is
```
kim, line 33:
```suggestion
    /// the identifier  does not exist, it is created and a new identifier is
```

librad/src/paths.rs
FintanH: Ya, it seems like its uses were just getting the `values` of the `HashMap` anyway, and I believe we usually use the direct accessors in upstream.

FintanH (dismissed):
Looks good to me

librad/src/profile.rs
```diff
@@ -0,0 +1,112 @@
...
+    /// generated and written to the file.
+    ///
+    /// On Linux, the path to the active profile is
+    /// `$XDG_CONFIG_HOME/radicle/active_profile`.
+    pub fn load() -> Result<Self, io::Error> {
```
kim, line 38: Might be good to ensure that the profile id is only a path _component_, or otherwise ensure that it doesn't look funny (eg. `../../WORKGROUP`).

Maybe we can also hand out the id as a `Path`, to reduce suprises when it contains newlines or other funky characters.

kim (commented):
looks good, only minor remarks

librad/src/paths.rs
geigerzaehler: Good idea, since this is a breaking change anyway.

librad/src/profile.rs
geigerzaehler: > Might be good to ensure that the profile id is only a path _component_, or otherwise ensure that it doesn't look funny (eg. `../../WORKGROUP`).

Good point. Will add the validation logic.

> Maybe we can also hand out the id as a `Path`, to reduce suprises when it contains newlines or other funky characters.

Are your referring to `Profile::id`? Having it return `Path` might be confusing, especially since it can only be a path segment since the base directories are all different.

kim: Ya I was thinking of the "smart" `Display` impl for `Path`. If we sanitise the input properly, I'm ok with `&str`.

librad/src/paths.rs
geigerzaehler: Addressed in https://github.com/radicle-dev/radicle-link/pull/496/commits/79a81a3177810bf278b8ba73e644bd822aed573f

librad/src/profile.rs
geigerzaehler: Addressed in https://github.com/radicle-dev/radicle-link/pull/496/commits/79a81a3177810bf278b8ba73e644bd822aed573f

librad/src/profile.rs
```diff
@@ -0,0 +1,191 @@
...
+    path,
+    path::{Path, PathBuf},
+};
+use uuid::Uuid;
+
+use crate::paths::{project_dirs, Paths};
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("Invalid profile ID in RAD_PROFILE environment variable: {id}")]
```
FintanH, line 19: The convention for error messages is starting with lowercase, as mentioned in `radicle-decisions`

librad/src/profile.rs
```diff
@@ -0,0 +1,191 @@
...
+}
+
+/// Returns `true` if `id` is a valid profile ID.
+///
+/// A valid profile ID is not empty, does not contain path separators, is not
+/// a windows path prefix like `C:`, and is not a special component like `.` or
+/// `..`.
+fn is_valid_profile_id(id: &str) -> bool {
+    let mut components = Path::new(id).components();
+    if let Some(c) = components.next() {
```
FintanH, line 103: I think this could be cleaned up to:
```rust
match components.next() {
  Some(path::Component::Normal(_)) => {},
  _ => return false,
}
```

librad/src/profile.rs
```diff
@@ -0,0 +1,191 @@
...
+}
+
+/// Read the profile ID from a file or create the file and write a newly
+/// generated profile ID to it.
+fn load_profile_id(config_dir: &Path) -> Result<String, Error> {
+    let active_profile_path = config_dir.join("active_profile");
+
+    match fs::read_to_string(&active_profile_path) {
+        Ok(content) => {
+            let id = content.lines().next().unwrap();
```
FintanH, line 126: Is this `unwrap` safe?

librad/src/profile.rs
```diff
@@ -0,0 +1,191 @@
...
+    path::{Path, PathBuf},
+};
+use uuid::Uuid;
+
+use crate::paths::{project_dirs, Paths};
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("Invalid profile ID in RAD_PROFILE environment variable: {id}")]
+    InvalidProfileIdFromEnv { id: String },
```
FintanH, line 20: Is this used? :thinking:

librad/src/profile.rs
geigerzaehler: Yes. Will add a comment. Or is the preferred way to put the explanation as the argument to `expect`?

librad/src/profile.rs
geigerzaehler: Oops, forgot about this.

librad/src/profile.rs
```diff
@@ -0,0 +1,211 @@
...
+        if let Ok(id) = env::var("RAD_PROFILE") {
+            if !is_valid_profile_id(&id) {
+                return Err(Error::InvalidProfileIdFromEnv { id });
+            }
+            let paths = Paths::new(&id)?;
+            return Ok(Self { id, paths });
+        }
+
+        if let Ok(rad_home) = env::var("RAD_HOME") {
+            return Self::from_root(Path::new(&rad_home));
```
MeBrei, line 68: Shouldn't the `Paths` still contain the `profile id` here? Otherwise if you set `RAD_HOME` (eg when starting upstream) you wont have the multiple profile feature

librad/src/profile.rs
FintanH: Right, I'd have expected the contents would need to be non-empty :) Sure the file could be there but what happens if it's empty :smile:

MeBrei: I tested this and it threw an error for me. Maybe I removed the profile id in an incorrect way, but it doesn't seem safe to assume it being removed in a certain way. (What I did: opened the file with `vim`, used `dd` to remove the line, closed with `:x`)

librad/src/profile.rs
geigerzaehler: Good point. I was thinking about `RAD_HOME` as an exstension of profile selection where it would not only allow you to control the profile ID (like `RAD_PROFILE`) but also the location of all the data.

What you are suggesting would be to have `RAD_HOME` behave like setting all `XDG_*_HOME` variables all at once and having profiles work just like without `RAD_HOME` but under a different base directory.

I think this is the better approach as it decouples `RAD_HOME` and profiles and is a bit more predictable. @FintanH, @kim any thoughts on this?

FintanH: I'm not understanding the conversation here so I'll try and see how far my understanding goes.

So `RAD_HOME` decides where the root is which could contain multiple profiles and the active profile, i.e.
```
RAD_HOME
|--- active_profile
|--- 1
|    |--- git
|    |--- etc.
|--- 2
|    |---- git
```

But what @MeBrei is suggesting that `RAD_HOME` would point to _one_ of those profiles, for example, `RAD_HOME=<ROOT>/2`, is that correct?

MeBrei: > So `RAD_HOME` decides where the root is which could contain multiple profiles and the active profile, i.e.
> 
> ```
> RAD_HOME
> |--- active_profile
> |--- 1
> |    |--- git
> |    |--- etc.
> |--- 2
> |    |---- git
> ```
> 
@FintanH the above is what I was suggesting. 
Below is what is currently happening, if you set `RAD_HOME` to a profile directory, but it in my tests I set it to a non-profile directory and it resulted in a mixed version, where the proxy store was under a profile id, but the librad directories weren't
> But what @MeBrei is suggesting that `RAD_HOME` would point to _one_ of those profiles, for example, `RAD_HOME=<ROOT>/2`, is that correct?

librad/src/profile.rs
```diff
@@ -0,0 +1,211 @@
...
+
+        let profile_id_path = tempdir.path().join("active_profile");
+        let content = "foo\n";
+        std::fs::write(profile_id_path, content).unwrap();
+        let id = load_profile_id(tempdir.path()).unwrap();
+        assert_eq!(id, "foo");
+    }
+
+    #[test]
+    fn empty_profile() {
```
FintanH, line 202: This doesn't seem to be testing the empty case, it looks like the test case above? :eyes:

librad/src/paths.rs
```diff
@@ -40,9 +41,7 @@ impl Paths {
         .init()
     }
 
-    // Don't use system paths, but the supplied directory as a root.
-    //
-    // For testing, you know.
+    /// All paths are contained in the given directory.
     pub fn from_root(root: impl AsRef<Path>) -> Result<Self, io::Error> {
```
FintanH, line 45: In relation to https://github.com/radicle-dev/radicle-link/pull/496/files#r565931339, this doesn't take into account the `id`, so this https://github.com/radicle-dev/radicle-link/pull/496/files#diff-b5c21e30928026061fdc46f4a876c7bc8374ebe78fe68851375632dffe20d472R83 line seems off to me. You're creating the id, but not using it.

librad/src/profile.rs
FintanH: Mmmm I think I'm piecing it together now. Something seems off to me which I commented here https://github.com/radicle-dev/radicle-link/pull/496/files#r566020184

librad/src/profile.rs
geigerzaehler: I think it makes sense to treat an empty (but present) `active_profile` file like there is no ID yet. Will update the PR.

FintanH: What kind of scenario would that be useful for?

geigerzaehler: I think it’s just to avoid confusion on the users side. I can definitely foresee somebody deleting the content of the file and then wondering why profile is invalid. If you feel strongly otherwise, @FintanH, we can also leave as it is.

librad/src/paths.rs
geigerzaehler: You’re right, it does not take the profile ID into account. We could add the profile ID as a second argument and join it to root. But I’d rather prefer to do that at the call site, that is `Profile::from_root`. The latter isn’t implemented yet as @MeBrei  correctly pointed out.

librad/src/profile.rs
geigerzaehler: I’ve updated the PR (and added test) for to implement multi-profile support within `RAD_HOME`. I hope this should clarify things.

librad/src/profile.rs
MeBrei: The idea is to trigger a new Identity being created when no `active_profile` can be found. I think we should serve to both cases if the user removes the contents of the file or deletes the file completely.

FintanH: Ya, I would think this is an error variant. I wouldn't think that an empty profile is a valid one.

librad/src/profile.rs
kim: I'm good with what's in c8e6f5d43c8ac8a94e5ba3039420a147e19a4a63.

geigerzaehler: Sorry folks, didn’t actually push the updates. Could you have another look?

librad/src/profile.rs
geigerzaehler: There seems to be some disagreement here, @MeBrei, @FintanH. Either treat an empty file as an error or treat it as no active profile does exist.

After giving this some thought I believe it would be best to treat an empty file as an error. This makes the behavior more straight-forward and reduces the number of edge cases. If the user deletes the content of the file then we throw an error instead of working around this.

FintanH: Makes sense to me :ok_hand:

librad/src/profile.rs
kim: Hm ok, so `RAD_HOME` and `RAD_PROFILE` are effectively mutually exclusive -- if I set `RAD_PROFILE`, I'd need to set `XDG_*_HOME` to make data go into a non-standard location (which is only possible under Linux). If I set `RAD_HOME`, the only way to control profile selection is via the `active_profile` file, which must already exist.

I find this _more_ confusing, but we all seem to have different expectations there. Perhaps we should continue to consider `RAD_HOME` as a convenience for test setups (not "stable API"), and look into respecting `XDG_*` variables regardless of platform?

MeBrei: I also assumed both env variables could be combined. We already "released" `RAD_HOME` as a possibility to have an independent dev setup in upstream.

geigerzaehler: As discussed in https://github.com/radicle-dev/radicle-link/issues/488#issuecomment-770923997 we will go with @MeBrei’s suggestion. This is implemented in f36bfc9a8aad7acdc637f5abd913a1c4dba61915

librad/src/profile.rs
geigerzaehler: Fxied

librad/src/paths.rs
geigerzaehler: Fixed in f36bfc9a8aad7acdc637f5abd913a1c4dba61915

geigerzaehler:
Thanks for all the input @kim, @FintanH, @MeBrei. Should be ready for review again.

kim (approved):
:+1:

Do we not need a persistent `set_active_profile` yet?

librad/src/profile.rs
```diff
@@ -0,0 +1,234 @@
...
+
+impl Profile {
+    /// Creates a profile by loading the profile identifier and paths from
+    /// the environment variables or well-known file.
+    ///
+    /// By default, the profile identifier is read from the `active_profile`
+    /// file in a system specific location. The paths returned by
+    /// [`Profile::paths`] are based on system paths (see [`directories::
+    /// ProjectDirs`]) and include the profile identifier. If the file
+    /// containing the identifier  does not exist, it is created and a new
```
FintanH, line 44:
```suggestion
    /// containing the identifier does not exist, it is created and a new
```

geigerzaehler:
> Do we not need a persistent `set_active_profile` yet?

Not yet. In upstream you can only use `RAD_PROFILE` to switch between profiles.

librad/src/profile.rs
```diff
@@ -37,51 +37,68 @@ impl Profile {
...
         }
 
         if let Ok(rad_home) = env::var("RAD_HOME") {
-            return Self::from_root(Path::new(&rad_home));
-        }
-
-        let id = load_profile_id(project_dirs()?.config_dir())?;
+            Self::from_root(Path::new(&rad_home), env_profile_id)
+        } else {
+            let id = if let Some(id) = env_profile_id {
```
FintanH, line 74: This suspiciously looks like `from_root` :eyes: Would I be right in thinking that this is `Self::from_root(project_dirs()?.config_dir()), env_profile_id)`? And then ultimately these just differ on the `root` being passed, i.e. `RAD_HOME` or the project dirs

librad/src/profile.rs
```diff
@@ -219,15 +231,4 @@ mod test {
         let id = load_profile_id(tempdir.path()).unwrap();
         assert_eq!(id, "foo");
     }
-
-    #[test]
-    fn empty_profile_file() {
```
FintanH: Wait, is there a test for the empty file causing an error?

FintanH (commented):
Looks good :+1: Few nits/questions

librad/src/profile.rs
geigerzaehler: The difference between this and `Profile::from_root` is that here we call `Paths::new(id)` whereas in `Profile::from_root` we call `Paths::from_root(..)`; and both `Paths` constructors also differ. What this has in common with `Profile::from_root` is unwrapping the profile ID or loading it from file, but that’s not too much and doesn’t warrant extracting it.

librad/src/profile.rs
geigerzaehler: Added in https://github.com/radicle-dev/radicle-link/pull/496/commits/d33410cc76657db918e3ae259ee00623acaf1741

librad/src/profile.rs
FintanH: Ah you're right. Thanks for checking that :)

FintanH (approved):
Great! Thanks for the good work here. I'm glad we were able to work out the finer details :)

PR #502, opened 2021-01-27
Fix smallvec vulnerability

FintanH:
We force the smallvec dependency to be 1.6.1 due to the vulnerability
outlined in the issue here Amanieu/parking_lot#274. 
We depend on governor which in turn depends on parking_lot.

Also updating the field `db-url` to `db-urls`, and outputting the version of cargo deny for inspection sake.

### DIFF
--- a/librad/Cargo.toml
+++ b/librad/Cargo.toml
@@ -111,6 +111,12 @@ features = ["logging", "dangerous_configuration"]
 version = "1.0"
 features = ["derive"]
 
+# Note: this is not a direct dependency. governor depends on
+# parking_lot, which depends on smallvec. There is vulnerability
+# in v1.0 and so we're forcing cargo to use this version instead.
+[dependencies.smallvec]
+version = "1.6.1"
+
 [dependencies.tokio]
 version = "0.2"
 features = ["full"]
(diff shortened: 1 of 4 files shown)


librad/Cargo.toml
```diff
@@ -111,6 +111,12 @@ features = ["logging", "dangerous_configuration"]
 version = "1.0"
 features = ["derive"]
 
+# Note: this is not a direct dependency. governor depends on
+# parking_lot, which depends on smallvec. There is vulnerability
+# in v1.0 and so we're forcing cargo to use this version instead.
+[dependencies.smallvec]
+version = "1.6.1"
```
kim, line 118: Can we say `>= 1.6.1, 1.6`?

Also, how do we know this is no longer needed? Can we know? What can we know? Are we?

FintanH: > Can we say >= 1.6.1, 1.6?

Ya, `>= 1.6.1` would include the fix :+1: 

> Also, how do we know this is no longer needed? Can we know? What can we know? Are we?

lol, I'll include this link https://github.com/antifuchs/governor/issues/60. Once that's in and bumped we could drop this.

kim (approved):

FintanH:
Closed via https://github.com/radicle-dev/radicle-link/commit/41c216bc7ae3605442608ce852c8aef99aef4b76

PR #503, opened 2021-01-28
Update governor

FintanH:
There was a quick turnaround on the governor update. This means we can
remove the `smallvec` dependency and instead bump the `governor` one.


kim (approved):
o7

FintanH:
Thank you guvn'r

PR #504, opened 2021-01-29
Fixes found while testing seed

FintanH:
While testing the `radicle-bins` code on the `seed-next` branch I spotted some issues.
The details of the fixes are in the commit messages.

I'll probably keep this open while I possibly find anything else that can be improved.

### DIFF
--- a/librad/src/net/peer.rs
+++ b/librad/src/net/peer.rs
@@ -546,6 +546,11 @@ impl LocalStorage for PeerStorage {
                 // FIXME(fintohaps): Do we want to tell the malkovich when we are behind in the
                 // gossip message?
                 Ok(ReplicateResult::Latest) | Ok(ReplicateResult::Behind) => {
+                    // The peer we fetch from is always the `origin` we figured out.
+                    let urn = Right(Originates {
+                        from: origin,
+                        value: has.urn.clone(),
+                    });
                     if self.git_has(urn, head).await {
                         PutResult::Applied(Gossip {
                             // Ensure we propagate exactly the `origin` we
--- a/librad/src/net/peer.rs
+++ b/librad/src/net/peer.rs
@@ -533,24 +533,17 @@ impl LocalStorage for PeerStorage {
         };
 
         let res = if is_tracked {
-            let urn = match has.origin {
-                Some(origin) => Right(Originates {
-                    from: origin,
-                    value: has.urn.clone(),
-                }),
-                None => Left(has.urn.clone()),
-            };
+            // The peer we fetch from is always the `origin` we figured out.
+            let urn = Right(Originates {
+                from: origin,
+                value: has.urn.clone(),
+            });
             let head = has.rev.as_ref().map(|Rev::Git(head)| *head);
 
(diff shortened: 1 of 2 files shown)


librad/src/net/peer.rs
```diff
@@ -546,6 +546,11 @@ impl LocalStorage for PeerStorage {
                 // FIXME(fintohaps): Do we want to tell the malkovich when we are behind in the
                 // gossip message?
                 Ok(ReplicateResult::Latest) | Ok(ReplicateResult::Behind) => {
+                    // The peer we fetch from is always the `origin` we figured out.
+                    let urn = Right(Originates {
```
kim, line 550: Note that this is still not fully correct: we are passing the `PeerId` we got the gossip message from, but should actually pass the `origin` field (because that message might be relayed). This is fixed in #483

FintanH: Are you positive? Afaict they're both using the result of `origin.unwrap_or(provider)`

kim: That's the payload's `origin`

FintanH: I _think_ they are equivalent now, or else I'm missing something entirely

kim: Not at all. The payload origin refers the remote the sender is advertising, while the gossip envelope origin refers to the sender itself. If a payload is determined `Uninteresting`, the message is still relayed, but with the origin unchanged (otherwise, the origin is changed to the receiver). Hence, all receivers must attempt to fetch from the envelope origin, because the data may not be present at the peer who sent the message.

FintanH: I'm confused as to what the "payload" is and what the "gossip envelope" is. Looking at `net/next` the type of `has` is [this](https://github.com/radicle-dev/radicle-link/blob/net/next/librad/src/net/protocol/gossip.rs#L46-L67). So the `origin` is determined [here](https://github.com/radicle-dev/radicle-link/blob/net/next/librad/src/net/peer/storage.rs#L139), and the `urn` is built with that `origin` [here](https://github.com/radicle-dev/radicle-link/blob/net/next/librad/src/net/peer/storage.rs#L149-L152).

On this branch, `has` is this [type](https://github.com/radicle-dev/radicle-link/blob/fixes/librad/src/net/peer/types.rs#L46-L67), looks pretty much the same as `Payload`, I believe. The `origin` is determined [here](https://github.com/radicle-dev/radicle-link/blob/fixes/librad/src/net/peer.rs#L526) -- again the same -- and the `urn` is built with that `origin [here](https://github.com/radicle-dev/radicle-link/blob/fixes/librad/src/net/peer.rs#L537-L540).

So is what you're trying to point out, outside of this section of code? [This change](https://github.com/radicle-dev/radicle-link/pull/483/files#diff-a6e6c0235179147a6171f883cc230707458f6f101dad98fa4444c38cd7b816d0L98) perhaps?

Sorry if I'm missing the point, I'm trying to get it straight in my head :)

kim (approved):

xla (approved):
🍀 ⏬ 🍷 🌜

PR #507, opened 2021-01-31
radicle-copy

FintanH:
The aim of this is to introduce an API for creating projects alongside their working copies. There's a binary included here for the moment to make it easier for playing around with, but I plan to move that `radicle-bins` before the PR gets merged -- I'll leave it in draft until then.

This is part of moving the functionality out of upstream over to link. I took the notion to do this since there were some Radicle fans that wanted to automate creating projects (so part of the plan is to have a batch add :))

An addition to `librad` are these `relations` modules. They introduce the concept of how peers in your tracking graph are related to the projects. The reason this was added was because it's useful to add peers to the include only if you've ended up replicating their data. It's also used in upstream to see if you have replicated a peer yet when you request to track them.


copy/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "radicle-copy"
+version = "0.1.0"
+authors = ["The Radicle Team <dev@radicle.xyz>"]
+edition = "2018"
+license = "GPL-3.0-or-later"
+
+[dependencies]
+# FIXME: Move to radicle-bins
+anyhow = "1.0"
```
kim, line 10: We may want to loosen the constraint to "1" -- we depend on `anyhow` elsewhere with this bound, and it's (too) easy to make `cargo` pull in more than one version of the same crate.

copy/src/bin/main.rs
```diff
@@ -0,0 +1,198 @@
...
+    signer::{BoxedSigner, SomeSigner},
+};
+use radicle_keystore::{
+    crypto::{self, Pwhash},
+    pinentry::Prompt,
+    FileStorage,
+    Keystore,
+};
+
+const SECRET_KEY_FILE: &str = "librad.key";
```
kim, line 34: Awww, this guy again -- do we have this hardcoded in so many places that we'll need to stick to this filename forever?

copy/src/bin/main.rs
```diff
@@ -0,0 +1,198 @@
...
+pub enum Command {
+    Existing(Existing),
+    Fork(Fork),
+    New(New),
+    Update(Update),
+}
+
+/// 🆙 Update the remotes that exist in the include file for the given project
+#[derive(Debug, FromArgs)]
+#[argh(subcommand, name = "include-update")]
```
FintanH, line 57: Oh one annoying thing about `argh` is that we can't do something like `include update` as sub-sub-commands. Or at least not that I can see, so I created a ticket https://github.com/google/argh/issues/79

copy/src/fork.rs
```diff
@@ -0,0 +1,238 @@
...
+
+    #[error(transparent)]
+    Ref(#[from] git_ext::name::Error),
+
+    /// An error occurred in the local transport.
+    #[error(transparent)]
+    Transport(#[from] librad::git::local::transport::Error),
+}
+
+/// Based off of the `Fork`, clone the project using the provided inputs.
```
kim, line 52: Is the terminology "fork" used in `upstream`? That'd be unfortunate, because the defining property of a fork is a change of ownership, which is not what is happening here. Can we agonise over this?

FintanH: It's a mixture between "checkout" and "fork" :grimacing: If you're creating a copy from your own project it's a checkout and if you're transitioning from a project you're only tracking then it's a "fork".

copy/src/bin/main.rs
FintanH: Hmmmm, we _could_ just look under the `keys` directory and pick the one that's there. If there's multiple then allow the user to specify which via a CLI param. So there are 3 scenarios:
1. No keys exist -- throw an error saying they need to initialise
2. There's a single file under `keys`, e.g. `keys/ringo.key`, and we use that
3. There are many files under `keys`, e.g. `keys/john.key`, `keys/paul.key`, `keys/george.key`, `keys/ringo.key`, and we throw an error saying PICK ONE

copy/src/bin/main.rs
kim: I think you need to insert an extra indirection through a `struct`, instead of ~~doing the obvious thing~~ nesting `enum`s: https://github.com/google/argh/blob/master/tests/lib.rs#L612-L630

FintanH: arrrggghhhhhhhh, thanks for finding that. I'll try it out

copy/src/fork.rs
kim: I'd propose "chork", then

copy/src/bin/main.rs
FintanH: Always the 0, 1, or many problem. I added my proposal above for now anyway. Maybe we should just fix it to a single key. Move fast and delete things :racing_car:

FintanH:
I'm happy with the functionality and organisation. I think it would be a good idea to have tests as well, and maybe a quick look over the docs again too (I cribbed a lot of them from upstream and haven't checked if all of them make sense still).

kim (approved):
🌸 🌱 🪴 🍄

FintanH:
Closed via https://github.com/radicle-dev/radicle-link/commit/4aa18fbe3ba0a79e1c5645096387a885bac7cfc7

PR #508, opened 2021-02-02
librad: reorganise integration tests

kim:
The slightly off module structure is due to cargo compling all top-level
modules into executables.

There are certainly more opportunities to extract common subroutines,
but this is left to a future brooming pass.


xla (approved):
🏟 🤕 🐒 😘

FintanH (approved):
:metal:

PR #510, opened 2021-02-07
copy: batch create

FintanH:
Adding a batch creation option. It allows the person to specify a path
to a manifest file, in JSON format, and iterates over the projects
creating them.

### DIFF
--- a/copy/src/cli/main.rs
+++ b/copy/src/cli/main.rs
@@ -4,7 +4,6 @@
 // Linking Exception. For full terms see the included LICENSE file.
 
 use std::{
-    convert::TryFrom as _,
     fs,
     path::{Path, PathBuf},
 };
@@ -15,9 +14,7 @@ use librad::{
     git::{
         identities::{self, local},
         storage::Storage,
-        Urn,
     },
-    git_ext::{OneLevel, RefLike},
     keys::{PublicKey, SecretKey},
     profile::Profile,
     signer::{BoxedSigner, SomeSigner},
@@ -30,10 +27,7 @@ use radicle_keystore::{
 };
 
 use super::args::{community, garden, Args, Command, Community, Garden};
-use crate::{
-    garden::{graft, plant, repot},
-    include,
-};
+use crate::{garden::graft, include};
 
 pub fn main() -> anyhow::Result<()> {
     let args: Args = argh::from_env();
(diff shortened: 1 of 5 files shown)


copy/src/garden/repot.rs
```diff
@@ -20,6 +20,9 @@ use crate::{git, sealed::Sealed};
 
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
+    #[error("attempted to canonicalize the path `{path}` but we encountered an error: {err}")]
```
xla, line 23: Good UX.

xla (changes requested):
> Adding a batch creation option. It allows the person to specify a path
> to a manifest file, in JSON format, and iterates over the projects
> creating them.

Could you provide a use-case and/or the motivation of this change? As it stands now it's hard to understand why this addition is meaningful.

FintanH:
> Could you provide a use-case and/or the motivation of this change? As it stands now it's hard to understand why this addition is meaningful.

Ya good point, I should have prefaced this PR with that :) The issue with the CLI at the moment is that if you wanted to compose calls to create projects you would have to enter your password/passphrase every time. This version allows you to create a manifest file to do the creation in one-fell-swoop after entering the passphrase once.

I was chatting with dxos alongside @abbey-titcomb, like a week ago, and they were looking for a way to do this. Basically, it'll make it easier to bootstrap a bunch of projects in your monorepo easily. The next piece of the puzzle will be having a tool to pull projects down from your GitHub, GitLab, or whatever you're using, and generating this manifest file. Then this can produce your Radicle projects :tada:

copy/src/cli/main.rs
```diff
@@ -45,36 +39,35 @@ pub fn main() -> anyhow::Result<()> {
...
-
-                let default_branch = OneLevel::from(RefLike::try_from(default_branch.as_str())?);
-                let raw = Plant::new(description, default_branch, name, path);
-                let valid = Plant::validate(raw)?;
-                let path = valid.path();
-                let project = plant(paths.clone(), signer, &storage, whoami, valid)?;
-
-                project_success(&project.urn(), path);
+            garden::Options::Bed(bed) => {
+                for pot in bed.load()? {
```
FintanH, line 43: Note to self: This would be better if we validate all the payloads first

kim:
> The issue with the CLI at the moment is that if you wanted to compose calls to create projects you would have to enter your password/passphrase every time.

This is somewhat orthogonal, and perhaps it's time to devise a default way to handle this. It's a lot less complicated when there's no electron between keyboard and executable -- the remote helper is able to use git credential helpers today, but it would also not exactly be rocket science to allow GPG/SSH agent signers (modulo figuring out how that would interact with `upstream`).

kim:
> This version allows you to create a manifest file to do the creation in one-fell-swoop

Do you think JSON is a good format, though? It's virtually impossible to edit by hand.

FintanH:
> Do you think JSON is a good format, though? It's virtually impossible to edit by hand.

I've no strong opinion there. I just went with JSON because of ubiquity. What format would make more sense? Dhall?! :upside_down_face:

kim:
Ya, idk, YAML perhaps.

/me ducks

The thing is that this currently requires to store a JSON array in a file (`[ { .. }, { .. } ]`), while for practical reasons you'll probably want to walk a directory containing one-manifest-per-file. YAML would _optionally_ allow to concatenate into a single file.

xla:
> Ya good point, I should have prefaced this PR with that :) The issue with the CLI at the moment is that if you wanted to compose calls to create projects you would have to enter your password/passphrase every time. This version allows you to create a manifest file to do the creation in one-fell-swoop after entering the passphrase once.

As time hinted at above @kim it seems orthogonal. It appears more a band-aid to the symptom than addressing the underlying issue. It shows that introducing an intermediate like the manifest opens up a lot of questions and adds additional machinery that needs maintenance.

kim:
I'm not against this. Creating fresh projects (of either kind) from the CLI is something I would like to see. 

What is a bit odd indeed with the current code is that the metadata payload is restricted, where it should really be the identity document. I also think that the composability goal is only met if the inputs can be piped to the program's stdin. Turns out [serde has some streaming support](https://docs.serde.rs/serde_json/struct.StreamDeserializer.html), so perhaps JSON objects back-to-back wouldn't be so bad.

FintanH:
> As time hinted at above @kim it seems orthogonal. It appears more a band-aid to the symptom than addressing the underlying issue. It shows that introducing an intermediate like the manifest opens up a lot of questions and adds additional machinery that needs maintenance.

I'm not sure I follow. A manifest is just some way of having a batched input into the system that we can decode and execute actions on. What underlying issue are you referring to?

> What is a bit odd indeed with the current code is that the metadata payload is restricted, where it should really be the identity document.

This is because the part of the input is the directory for pointing to or creating the working copy -- for existing or new working copies respectively. I suppose this could be changed to be a pairing of a payload and a directory, but then we have to answer a couple of questions:
1. In the case of an existing working copy, does it matter that the project name can be different from the directory name? i.e. the project's name is `foo` and the directory is `/home/haptop/Developer/bar`. I suppose this is fine really, but might be a bit confusing. 
2. To initialise the working copy we require a default branch, but the payload doesn't require one. Should the CLI have a `const` definition for new projects and infer it from the working copy if it's existing? Or are there other options we can consider to solve this?

>  I also think that the composability goal is only met if the inputs can be piped to the program's stdin. Turns out serde has some streaming support, so perhaps JSON objects back-to-back wouldn't be so bad.

Ah thanks for the link! I wanted a more `Iterator` style creation but went for the simplest option, I should've looked harderer

FintanH:
> while for practical reasons you'll probably want to walk a directory containing one-manifest-per-file

Oh could you expand on this a bit more, I'm not sure I follow :)

xla (approved):
💴 〰️ 🈁 ⤵️

FintanH:
Closing this with #682 and upcoming CLI RFC in mind

PR #511, opened 2021-02-07
tracing-attributes version bound

FintanH:
0.1.12 breaks our builds at the moment so we should specify an upper
bound of 0.1.11.

See: https://github.com/tokio-rs/tracing/issues/1227

Feel free to fix/merge while I'm out next week :smile:


xla (approved):
⛩ 🐱 🗞 ✍

PR #512, opened 2021-02-16
Use fetch instead of download

FintanH:
We have to use fetch instead of download due to a bug that is resulting
in a bad negotiation between two clients where an extra `want` is
issued after a `have`. Some more details are here
https://github.com/libgit2/libgit2/issues/5799.

This should only be temporary because this means we fetch the refs
advertisement multiple times, redundantly.

### DIFF
--- a/librad/src/git/fetch.rs
+++ b/librad/src/git/fetch.rs
@@ -577,6 +577,7 @@ impl<'a> DefaultFetcher<'a> {
         &mut self,
         fetchspecs: Fetchspecs<PeerId, git::Revision>,
     ) -> Result<FetchResult, git2::Error> {
+        let mut updated_tips = BTreeMap::new();
         {
             let limit = fetchspecs.fetch_limit();
             let refspecs = fetchspecs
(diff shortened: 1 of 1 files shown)


kim:
Does https://github.com/libgit2/libgit2/issues/3917 affect us as well? Because we'd still be setting remote callbacks _after_ connecting, I think. Or maybe not, because `fetch` will reconnect?

FintanH:
> Does libgit2/libgit2#3917 affect us as well? Because we'd still be setting remote callbacks after connecting, I think. Or maybe not, because fetch will reconnect?

Finding it hard to tell from the C code, but I can try to play with it and see in a minimal example.

Interestingly, `git2` hides being able to set the callbacks on `connect` and has a TODO https://docs.rs/git2/0.13.17/src/git2/remote.rs.html#146-158.

kim:
I'm seeing log output from the `fetch`, but not from `update_tips`. That means it's working, but we need to move the `update_tips` callback to the callbacks for `fetch` to make the return value be correct.

FintanH:
Ah I'm a dunce... I need to add back the `update_tips` code.

FintanH:
Closed via https://github.com/radicle-dev/radicle-link/commit/a94fc4a27e1080743a05814c7a3685dbc63f428f

PR #514, opened 2021-02-18
Update Cargo.toml: tracing-attributes

FintanH:
Since there was a release of tracing-attributes
(https://crates.io/crates/tracing-attributes/0.1.13) we can place a
better range on the versions that is lower than 0.1.12 and anything
upwards of 0.1.13, inclusive.

### DIFF
--- a/librad/Cargo.toml
+++ b/librad/Cargo.toml
@@ -47,6 +47,7 @@ tempfile = "3.1"
 thiserror = "1.0"
 time = "0.2"
 tracing = "0.1"
+tracing-attributes = "<0.12.0, ^0.1.13"
 tracing-futures = "0.2"
 unicode-normalization = "0.1"
 uuid = { version = "0.8", features = ["v4"] }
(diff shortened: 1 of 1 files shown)


librad/Cargo.toml
```diff
@@ -47,6 +47,7 @@ tempfile = "3.1"
 thiserror = "1.0"
 time = "0.2"
 tracing = "0.1"
+tracing-attributes = "<0.12.0, ^0.1.13"
```
xla, line 50: Why the bound to be smaller than `0.12.0`?

xla (approved):
🍥 🙋 🐁 🛰

librad/Cargo.toml
FintanH: It allows dependents to pick a `tracing-attributes` lower than `0.12.0` so that they're not pinned to the `0.1.13` or above :)

xla: cargoooooooooooooooooo

FintanH: :car::dash:

kim (approved):

PR #516, opened 2021-02-18
instrumentation: stream error span + URN display

kim:
Fixes a few instances of URNs going through Debug instead of Display for
logging. This is still not ideal, because composite values are often
better logged through Debug, but then the URN field is somewhat less
useful again. Need to investigate with upstream why the Value trait is
sealed.

Also improves the span reporting when a stream error occurs.


FintanH (approved):
A bit annoying that we have to skip and then add fields.. I'll have to check out your proposal on switching logging framework

PR #518, opened 2021-02-18
Bump minicbor

kim:
Since 0.6, minicbor has gained support for unambiguously mapping byte
slices to CBOR arrays. We can remove some code thusly


kim (commented):
Note that this patch is stacked on top of #516, and the PR base set accordingly
to get 'Hüb to show a minimal diff. There is no dependency, though, so the patch
should apply against master.

You will see this more often from me.

xla (approved):
👮 🎌 🐻 🎷

FintanH (approved):
Nice :+1:

PR #519, opened 2021-02-18
git, net: Add a "capability" trait to mark pooled storage

kim:
It turns out we will need access to a `git::Storage` from some stream
handlers, specifically in pooled form due to `Send`/`Sync` requirements.
As we need to pull that out from a larger structure (respectively
wrappers), a trait is in order. Since we need to thread the constraint
through the various IO function, we defined a composite trait to make
constraint signatures shorter.

### DIFF
--- a/librad/src/git/storage/pool.rs
+++ b/librad/src/git/storage/pool.rs
@@ -3,30 +3,51 @@
 // This file is part of radicle-link, distributed under the GPLv3 with Radicle
 // Linking Exception. For full terms see the included LICENSE file.
 
-use std::sync::{Arc, Mutex};
+use std::{
+    ops::Deref,
+    sync::{Arc, Mutex},
+};
 
-use deadpool::managed::{Manager, RecycleResult};
+use deadpool::managed::{self, Manager, Object, RecycleResult};
 
 use super::{Error, Storage};
 use crate::{paths::Paths, signer::Signer};
 
 pub type Pool = deadpool::managed::Pool<Storage, Error>;
-pub type Pooled = deadpool::managed::Object<Storage, Error>;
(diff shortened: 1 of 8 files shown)


librad/src/git/storage/pool.rs
```diff
@@ -3,30 +3,53 @@
...
+
+#[async_trait]
+impl Pooled for Pool {
+    async fn get(&self) -> Result<PooledRef, PoolError> {
+        self.get().await.map(PooledRef::from)
+    }
+}
+
+/// A reference to a pooled [`Storage`].
 // TODO: may go away once https://github.com/bikeshedder/deadpool/pull/69
```
FintanH, line 32: fyi: it looks like this is in `0.7.0` https://github.com/bikeshedder/deadpool/blob/deadpool-v0.7.0/src/managed/mod.rs#L152-L162

FintanH (approved):
Ya, makes sense :+1:

kim:
fyi: it looks like this is in 0.7.0
>

Ah right, thanks. We can pull this in now that we’re on EINS. Will amend.

>

kim:
Well, nevermind. Because of trait introduced here, there is not much value in generalising to `AsRef`, because then we will need to specify the concrete type for the trait anyways. I'll keep the wrapper, as that at least doesn't leak the dependency type.

FintanH:
> Well, nevermind. Because of trait introduced here, there is not much value in generalising to AsRef, because then we will need to specify the concrete type for the trait anyways. I'll keep the wrapper, as that at least doesn't leak the dependency type.

Ah, I see. Sounds good :+1: Should we keep the TODO then, or remove it?

PR #521, opened 2021-02-19
git-ext, storage: Add a way to quickly create a commit

kim:
Mainly in integration tests, we often want to create some git history.
Currently, the only way to do that is to go through a working copy,
which exercises that flow, but is often inconvenient (and slow).

Since we want to avoid handing out a raw git handle to the underlying
"monorepo" storage, we instead expose a public means to create a commit
in a reasonably safe and almost ergonomic way.

### DIFF
--- a/librad/src/git/util.rs
+++ b/librad/src/git/util.rs
@@ -1,0 +1,91 @@
+// Copyright © 2019-2020 The Radicle Foundation <hello@radicle.foundation>
+//
+// This file is part of radicle-link, distributed under the GPLv3 with Radicle
+// Linking Exception. For full terms see the included LICENSE file.
+
+use git_ext::{is_not_found_err, reference};
+use std_ext::result::ResultExt as _;
+
+use super::{
+    refs::{self, Refs},
+    types::Namespace,
+};
+
+pub use super::{Storage, Urn};
+pub use git_ext::Tree;
+
+pub mod error {
+    use super::*;
+    use thiserror::Error;
+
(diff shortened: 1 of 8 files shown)


librad/src/git/util.rs
```diff
@@ -0,0 +1,91 @@
...
+    message: &str,
+) -> Result<git2::Oid, error::QuickCommit> {
+    let repo = storage.as_raw();
+
+    let author = repo.signature()?;
+    let branch = {
+        let path = reference::Qualified::from(urn.path.clone().unwrap_or(reflike!("master")));
+        reflike!("refs/namespaces")
+            .join(Namespace::from(urn))
+            .join(reference::Qualified::from(reference::OneLevel::from(path)))
```
FintanH, line 66: Why do we qualify to go to one level and go back to qualified again? Seems a bit convoluted, or at least I'm confused by it.

FintanH (dismissed):
Makes sense, should be useful :)

kim:
Why do we qualify to go to one level and go back to qualified again?
>
>
The construction via `from` ensures that it’s `refs/heads`, and not some
other category. Could be that the first conversion to qualified could be
omitted in favour of onelevel. Could also change to use `into_qualified` if
that makes it clearer.

>

FintanH (approved):
Slick :+1:

xla (approved):
🖊 📦 📙 💈

kim:
Closed via 18a4bd0

Issue #464, opened 2020-12-02
Move away from serde for Canonical JSON


kim:
`serde` requires canonicalisation to happen when domain types are already
erased. This creates several footguns:

* String types may not compare equal after a roundtrip
* Set types do not have a counterpart in the `serde` datamodel: they come out
  the same as arrays (i.e. insertion-ordered lists). What we want is to always
  have ordered-set semantics.
* Floating point numbers (which are illegal in Canonical JSON) cause a runtime
  error

To remedy this, we should precisely constrain what types can be used to compose
a structure subject to canonicalisation.

The proposal is to always go through an intermediate representation (akin to
`enum Value`), which is already guaranteed to hold the canonical form by
construction. Type-directed conversion for `std` types may be provided. Literals
may be supported via proc-macros.

Reasoning:

* It is no less memory-efficient than what we have now, as we need to buffer
  map-shaped objects before canonicalisation
* It is less surprising (e.g. plain `String` or `&str` would just not be
  representable)
* The actual encoding / decoding could be made much more efficient in terms of
  code size, compilation cost, dependencies, and possibly even runtime
  performance.

FintanH:
Would this essentially look like:
```rust
use crate::Cstring;
use std::collections::{BTreeMap, BTreeSet};

pub enum Value {
    Object(BTreeMap<Cstring, Value>),
    Array(BTreeSet<Value>),
    String(Cstring),
    Number(Number),
    Bool(bool),
    Null,
}

pub enum Number {
    U64(u64),
    I64(i64),
}

impl Canonical for Value {
/* left as an exercise to the implementor ^_^ */
}
```

Or is it more nuanced than that?

kim:
We’d probably want to provide derive macros for user-defined extension payloads.

PR #525, opened 2021-02-23
RFC: Identity Proofs

kim:
[Rendered](https://github.com/radicle-dev/radicle-link/blob/6213f30a9c520432662952e3498deb1ed42b21ef/docs/rfc/identity_proofs.md)


docs/rfc/identity_proofs.md
```diff
@@ -0,0 +1,214 @@
...
+Technically, this revision needs to be approved by all key delegations for
+verification to pass later on, but since we assume that eligible keys are held
+by the same person, it may be acceptable to publish the proof right away for
+user experience reasons.
+
+The actual proof consists of the following tuple:
+
+    (root, revision, public key, signature)
+
+> TODO: include a protocol bit (denoting "git"), for forwards compat?
```
NunoAlexandre, line 117: :ok_hand:

docs/rfc/identity_proofs.md
```diff
@@ -0,0 +1,214 @@
...
+      at the given `revision` is retrieved, its `root` compared against the
+      supplied one, and it is verified that the given `public key` is a member
+      of the `delegations` set. If expiration is specified in the claim, it is
+      determined if `expiration` seconds already elapsed since `created`
+      according to the clock, in which case verification fails.
+  2.2 Otherwise, the identity verification procedure is run first, after which
+      the history is walked backwards until the `revision` is found. Then, step
+      2.1 is performed.
+
+3. Lastly, the identity history needs to be examined for revocations.
```
NunoAlexandre, line 156: Could you explain why this is? As in, say that we validated the tip of the history where a new claim is added. Why is it necessary to examine the earlier history for revocations?

docs/rfc/identity_proofs.md
```diff
@@ -0,0 +1,214 @@
...
+
+It should be noted, however, that refreshing the proof from time to time in
+order to ensure freshness of the data retrieved through `radicle-link` is not
+always practicable.
+
+In order to prove that the(ir own) server is not lying by omission, Keybase
+[anchors a merkle root][keybase-stellar] on a blockchain, which includes all
+sigchains registered in the Keybase directory. Because `radicle-link` does not
+have such a central directory, this approach could only be applied to a partial
+view of the network.
```
NunoAlexandre, line 174: Something for us to consider?

NunoAlexandre (approved):
:mount_fuji:

docs/rfc/identity_proofs.md
kim: Ya I guess this can be collapsed into:

* find tip of refs/namespaces/<root>/refs/remotes/<public key>/rad/id 
* verify
* check claim is included
* if yes, walk backwards to verify revision is an ancestor, and signing key was eligible 

Originally I thought to avoid explicit revocation, but then, it is hard to prevent _accidental_ removal when the identity document is updated, which would result in an undefined state.

docs/rfc/identity_proofs.md
kim: Ceramic does something like this, where people can batch their transactions into a larger one (and share tx fees? Idk). Not sure where you get the merkle proof from.

docs/rfc/identity_proofs.md
FintanH: So if the revision is older than the latest tip, you'd have to look at its children for revocations, right?

docs/rfc/identity_proofs.md
```diff
@@ -0,0 +1,214 @@
...
+
+A claim can be revoked by creating a new identity revision which simply does not
+contain the claim payload. Likewise, a later claim describing the same `(service
+name, account)` mapping invalidates an earlier one.
+
+> NOTE: claiming multiple accounts per service is permissible in principle, but
+> might be undesirable for certain services. In this case, service-specific
+> post-validation is required.
+
+## Verification
```
FintanH, line 138: Should this section also have the step of verifying the claim's `proof`, if present?

docs/rfc/identity_proofs.md
kim: No actually not, that's where the confusion stems from: if we define that revocation is equivalent to creating a new revision which simply does not contain the "claim" payload, then we only need to look at the tip. If, however, the revision from the proof 4-tuple is not in the ancestry path, we either have a fork or not the latest tip.

FintanH (commented):
This looks good to me. I'd just like to clarify some of my understanding -- which will likely lead to more questions -- before approving :)

I'd also like to point out, to anybody who's looking to develop on top of this on `radicle-upstream`, that we would need to have the identity document update and signing flows implemented. This is because we'll have the initial `Person` document with just the `name` and then the document would be updated with the claims, which in turn will need signing for others in the network to at least trust _that_.

docs/rfc/identity_proofs.md
FintanH: > if we define that revocation is equivalent to creating a new revision which simply does not contain the "claim" payload

But isn't that how we are [defining revocation](https://github.com/radicle-dev/radicle-link/pull/525/files#diff-5389480c1bf9cf91c20b588fb42a26972437745bc31fa234d6be995afbaa001bR130-R131)? :smile: Or am I getting further confused? :grimacing:

kim: Yes, exactly. If doc at revision `A` verifies, but does not contain the claim, the proof can be rejected right away. To _report_ that it is a revocation (ie. it was valid at a previous point in time), revision `B` (from the proof tuple) must be an ancestor of `A` **and** the claim must be present at `A`. Arguably, you can compute this distinction lazily, because the difference between "was never valid" and "was previously valid" is only meaningful to humans.

FintanH: Ok, this is clear. Thanks!

docs/rfc/identity_proofs.md
```diff
@@ -0,0 +1,214 @@
...
+    * `created`
+
+      Creation timestamp of the claim, in seconds since 1970-01-01T00:00:00Z.
+
+    * `expires`
+
+      Seconds relative to `created`, after which the claim should no longer be
+      considered.
+
+* `proof` (optional)
```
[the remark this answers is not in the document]
kim: You know I like to pronounce the types in my head, so the name of an expression does not need to include it :) "proof URL"

docs/rfc/identity_proofs.md
kim: The proof URL is just a hint where to get the payload from. Say we'd support Twitter, then we can infer https://twitter.com/iamdevloper/status from what we have, but _every_ client would need to scrape the timeline until they find the proof in https://twitter.com/iamdevloper/status/1361327114260271109. Same thing for blockchain proofs, _although_ you could substitute the web3-provider.

FintanH: But what I'm not understanding is where this comes into play as part of the attestation. Is it up to the malkovich to check this or up to the client program? Is that why there's no recommendation on what to do with it?

docs/rfc/identity_proofs.md
```diff
@@ -0,0 +1,214 @@
...
+the similarities), and a desirable feature of applications built on top of it.
+
+The following sections describe how such claims shall be stored in the identity
+payload of a `radicle-link` identity, how to obtain a publishable proof, and how
+to verify it.
+
+## Claims
+
+Claims can only be made by identities of kind `Person`, and claim a single
+external account identifier. They are introduced by defining a new payload type,
```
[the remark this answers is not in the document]
kim: This is actually a good point. I don't think there is a way this would be beneficial, though, unless we would _also_ include the `HASH(claim)` in the tuple, or would be able to compute `HASH(claim)` (which would preclude time-based expiration). In this case, we could avoid having to read any data but the commit and immediate tree, and just compare hashes. There's no way around traversing the history backwards, though, because we need to check that `public-key` was in the delegations :thinking:

docs/rfc/identity_proofs.md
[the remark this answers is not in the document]
FintanH: I'm not sure I follow :sweat_smile: But I suppose you've answered it [here](https://github.com/radicle-dev/radicle-link/pull/525/files#r582778043) :)

docs/rfc/identity_proofs.md
kim: There isn't really anything to check. Say all you have is the identity doc. Now you invoke `rad sozial überprüfung`, and that reads the payload, sees aha deep-link into twitter, let's go and get the 4-tuple from there. If you already have the 4-tuple, you can omit this step. If you _don't_ have a 4-tuple, _nor_ a deeplink, that nifty program needs to traverse the twitter timeline until it finds a tweet which _looks_ like it could contain a 4-tuple.

The drawback is of course that adding this URL creates a new revision.

kim:
It is actually somewhat hilarious how inefficient this is in theory. Either I'm missing something, or life was a little easier on MySQL.

CodeSandwich (commented):
Great document, kudos for putting huge effort into codifying all these details!

docs/rfc/identity_proofs.md
```diff
@@ -0,0 +1,232 @@
...
+
+The actual proof consists of the following tuple:
+
+    (root, revision, public-key, signature)
+
+Note that the "git" protocol specifier of `radicle-link` URNs is implied, that
+is, future version MUST treat the absence of a disambiguating value as denoting
+"git".
+
+The values `root`, `revision`, and `public-key` are specified in
```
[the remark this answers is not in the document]
kim: A `revision` is the hash of the `tree` object, where the document `blob` contains the parent `tree` hash. The `commit` hash is not relevant for verification, yet the commit _message_ contains the signatures.

This means:
* two entirely disjoint _commit_ histories can describe the same _document_ history
* it is not possible to produce the same `revision` in a disjoint _document_ history, unless a hash collision is found, such that a quorum of the key delegations will be controlled by an attacker. Of course, if one owns a quorum of the keys, a fork can be produced.
* without the `root`, we wouldn't be able to _address_ any particular commit history (a `radicle-link` URN is `rad:git:<root>`). If all content was already available locally, we could find the `revision` tree, inspect the document to find the parent, retrieve that, etc, until we reach the `root` -- and then start over and find a commit which includes `revision`, in order to verify the signature chain. That's not very efficient, but doable. The problem is when we _don't_ have the content -- for example when we see this proof being posted on `<social network of choice>`, and now want to know what it is referring to.

docs/rfc/identity_proofs.md
```diff
@@ -0,0 +1,232 @@
...
+inspecting the identity storage.
+
+It is beyond the scope of this document to devise the exact external format to
+serialise the tuple into, as this is expected to vary from service to service.
+
+## Revocation
+
+A claim can be revoked by creating a new identity revision which simply does not
+contain the claim payload. Likewise, a later claim describing the same `SERVICE`
+invalidates an earlier one.
```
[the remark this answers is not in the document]
kim: I think temporary lack of validity is always the case -- you need to fulfill the signing obligation.

Whether or not we want to allow multiple accounts per service is an open question I'd say -- we can allow the payload type to be an array. I feel like the semantics of this is a bit wonky, though: if, for whatever reason, I choose to be one person on Radicle, but two persons on Twitter, then probably I'd like to include some indication of what's different (like "official", "shitposting" or something). But then, we'd also need to include this bit in the proof, don't we? Otherwise we'd need to try all matching.

docs/rfc/identity_proofs.md
```diff
@@ -0,0 +1,232 @@
...
+
+```json
+{
+    "SERVICE": {
+        "account": "STRING"
+        "expiration": {
+            "created": INTEGER,
+            "expires": INTEGER
+        },
+        "proof": "URL"
```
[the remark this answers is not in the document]
kim: Yeah, that'd probably be a good idea. Although we can define the `URL` type as an enum in the host language, so as to make it extensible.

docs/rfc/identity_proofs.md
```diff
@@ -0,0 +1,232 @@
...
+    https://radicle.xyz/link/claim/v1
+
+The shape of the JSON structure is:
+
+```json
+{
+    "SERVICE": {
+        "account": "STRING"
+        "expiration": {
+            "created": INTEGER,
```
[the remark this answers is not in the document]
kim: I'm not sure... a timestamp is always just a local timestamp, so unless some additional semantics are ascribed to it, there isn't so much one can do with it. Besides, there is also a redundancy: since we know the `PeerId` (== `public-key`), we can reify the commit which introduced the `revision`, which has a timestamp.

docs/rfc/identity_proofs.md
```diff
@@ -0,0 +1,232 @@
...
+* treat "web3"- like any other web-service
+
+  This may entail finding ways to reduce costs, for example by storing the proof
+  payload in transaction logs (which can be verified separately).
+
+---
+
+[ids]: ../spec/identities.md
+[wot]: https://en.wikipedia.org/wiki/Web_of_trust
+[keybase]: https://keybase.io
```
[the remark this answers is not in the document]
kim: Do you know of a place where keybase describes their thing in a more spec-like manner? I feel like most of their docs are lacking a bit on the details.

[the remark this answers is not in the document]
CodeSandwich: I don't know, this was the first time ever that I've entered their docs :D

docs/rfc/identity_proofs.md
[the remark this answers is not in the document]
kim: We may also opt to just omit this -- it is probably a bit weird to manage, because you can only add the URL after the proof was submitted, creating a new revision. Tools may also want to check the _plausibility_ of the combo `SERVICE` + `URL`. So perhaps just leaving it to the tool is also an option.

kim:
It has been concluded that for the use-case of radicle-dev/radicle-upstream#965, a simple one-way attestation in the other direction suffices, and can be expressed with a much more compact proof:

Consider a contract which allows to claim (and unclaim) some Radicle ID (URN). No two claims for the same ID at the same time are permitted. After a successful claim, the _transaction ID_ is included in the claimed URN's identity history.

If either the Ethereum address _or_ the `radicle-link` identity is trusted, only the inclusion of the transaction in a block on the main chain needs to be verified. Navigation from address to URN entails inspecting the transaction payload. Navigating from `radicle-link` identity to address entails inspecting the transaction sender.

Note that no elevation of trust results from this, nor assurance of an association with a "real" person. Neither does it improve the fork resiliency of the `radicle-link` identity, nor hiding of updates. I therefore think that this RFC has a justification in its own right.

Since the identity document payload can be freely extended by library users, and the verification semantics are entirely defined by the external system, I don't think a subsequent RFC is required. I also don't think any changes to `radicle-link` are required, except perhaps for convenience or code organisation reasons. #464 would not hurt, though.

docs/rfc/identity_proofs.md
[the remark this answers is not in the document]
CodeSandwich: > I think temporary lack of validity is always the case -- you need to fulfill the signing obligation.

I think that we can solve this problem easily if we allow the opposite claims to overlap in validity time ranges only partially.

Ethereum could be a good example. Let's say that today is day 80 (of whatever). I find a claim on Ethereum ranging from day 50 to 200, looks good. I go to link and find a counterpart claim from day 50 to 100, still good, now I can consider this attestation valid until day 100. A few days pass and I get an update from link: it claims the same attestation from days 50 to 200. This is great, now I know that the attestation is valid until day 200.

This schema allows seamless updates of attestations despite lack of synchronization and guarantees that on both sides we only need to know the single latest claim to have the full knowledge.

The only downside is that the validity ranges now must be solid: they have only start, end and no gaps inside. This is probably fine, I can't think of any use case where I would need to express something like "this Twitter account will be mine until April 1st, then it'll be somebody else's for a week and after that again mine for a year".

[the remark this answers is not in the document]
CodeSandwich: > I choose to be one person on Radicle, but two persons on Twitter, then probably I'd like to include some indication of what's different

If my previous comment is correct, we are guaranteed to only need 1 entry in the identity JSON per account. We could bypass the multiple accounts problem by duplicating the services, e.g. "twitter", "twitter_2" and "twitter_3". This will make the number of accounts strictly limited, but it'll be trivial to build a UI to display them and keep track of which account is which. Another advantage is that for now we can stick with having only "twitter" and add support for "twitter_2" when we fell like doing so. The JSON will stay unambiguous to interpret and easy to navigate.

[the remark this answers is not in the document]
kim: I am not sure I'm following. You **must** create a new revision always, but it may remain invalid until it was signed by the delegate quorum. There is also no guarantee that the tip will be seen by others -- that's why the revision is important: if you come across a proof, but are not able to retrieve the revision it is talking about, there is something funky going on.

It is impossible to talk about time without a reference. Arguably, when you want to have expiration, wall-clock time is _probably_ good enough -- after a week, _most_ people's computers will agree that `created + expires` is in the past. You can't safely base something like handover on this, though: if I publish such a proof on my homepage, but later sell the domain, the buyer can keep the proof up, but it would be compromised. If your time unit is block height, that problem goes away.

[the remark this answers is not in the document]
CodeSandwich: > You must create a new revision always, but it may remain invalid until it was signed by the delegate quorum.

I really don't think that the quorum signing process matters here. Let's assume that all revisions not quorum-signed are invalid and thus will be ignored by everyone except the quorum members.

> the revision is important

It seems that you envision a system where a link identity revision contains a link to proof on an external service (transaction hash) and that proof (transaction content) contains a reference to said identity revision leading to a chicken and egg problem.

We've discussed that yesterday and I think that we came to a conclusion, that there are means of deciding, which link revision is the tip (from best effort perspective, there may always be a better one floating on the network). This would break the cycle and allow looser and simpler relation between claims.

> if you come across a proof, but are not able to retrieve the revision it is talking about, there is something funky going on

Does any revision except the tip matter? Even if a 3rd party claim contains a proper revision, it doesn't matter if the tip doesn't recognize the said 3rd party.

> wall-clock time is probably good enough

Absolutely. You can't verify any security without a decent clock synchronization. If it's off by a few days and the user uses it to decide if they trust something on the internet, it's their own fault.

> if I publish such a proof on my homepage, but later sell the domain, the buyer can keep the proof up, but it would be compromised

I think that there is a missing part in this example: after selling a domain I buy a new one and hand over the attestation to it, right? In that case there's nothing we can do about it, if after handing over there are still people who are looking at a claim on a compromised 3rd party and hold an old link state, they will always trust the new owner. It doesn't matter how we measure time.

> If your time unit is block height, that problem goes away.

How? BTW every block has a timestamp with precision of a few seconds, but it probably doesn't matter if the claim on Ethereum contains the validity time range submitted as a parameter.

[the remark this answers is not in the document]
kim: I don’t understand what you are debating, so I think we may just stop here.

What we’ve discussed works pretty much exactly like the ACME protocol (RFC 8555), with comparable guarantees. This is enough for the purpose, but not enough for other security aspects of a `radicle-link` identity. Therefore, this RFC stands on its own.

You are welcome to suggest improvements to this proposal, or ask to clarify aspects of it, but I would like to ask you to do so in a more systematic way — just stating that this or that doesn’t matter, or that something clearly works without explaining your axioms makes it really hard to figure out how to respond.

So, if you forget about blockchains — I have removed everything except for merkle root anchoring — do the questions you raised in your last comment persist?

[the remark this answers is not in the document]
CodeSandwich: I'm trying to figure out how is the handover supposed to work? I haven't found it in the RFC, so I've proposed a solution, but you seem to have a different solution already in your mind.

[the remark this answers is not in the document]
kim: Aha, ok cool. I brought up this example trying to guess whether you are trying to find some flaw in this RFC, or considering if there needs to be anything else for the planned contract.

Maybe let’s look at the contract first: we said it’s just `register(URN)`, `unregister(URN)`, and we store the corresponding tx hashes in link. Let’s say the link revision which contains the unregister- (or simply removal of the register-, I’m not sure) tx gets lost. When I follow the register tx, I will find that it’s valid, but without an extra step I don’t know if later it was unregistered (and maybe someone else registered). However, we know that it is virtually impossible to _hide_ the chain prefix, so the window in which someone else who registered id `X` can pretend to own it is a couple of confirmations.

So I would say this is fairly safe without auto-expiration, assuming that it’s not hard to check if the _current_ address holding the link ID is the same as the one in the tx we know.


Let’s take away the contract: I post a proof somewhere where there is a risk of takeover (or just handover), like a DNS name. Let’s say I’m using LetsEncrypt. The is no way to perform the online check we just introduced, because there is no information about who owns the domain, or if that changed (that is, we cannot ask “who owned the domain at time X”). Because we could also be missing the revocation revision from link, it would be wise to specify an expiration time in the claim (preferably set to some time _before_ the domain expires).

Finally, we want to defend against account compromise (whether that’s social media or Ethereum). The only option we have is to revoke on the link side, but we cannot be sure that this is received. That’s where anchoring (or: checkpointing) the latest revision somewhere else comes in: if I see a revision there which I don’t have, I should reject any proofs until I have it. Without this, an expiration time will also do, but with larger uncertainty.

Does this make sense?

docs/rfc/identity_proofs.md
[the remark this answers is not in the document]
CodeSandwich: During our last chat you've explained, that the expiration date of a claim on link side is used for detection of abandoned identities and hidden updates. Do we need a separate expiration date for each service then? Could it be moved to the top level of the JSON and updated whenever a new version of identity is being prepared?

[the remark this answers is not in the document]
kim: Indeed.

If we treat social proofs different from radicleth attestations, we’ll still end up with two timestamps, though. Or we have the same namespace, but different variants (with very different semantics). Hm. JSON is hard.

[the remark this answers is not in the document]
CodeSandwich: Yeah, we'd end up with two timestamps per service:
- one for the whole document, mandatory and revoking everything if expired, even claims with no expiration date
- one for the service, optional, may revoke only the said service if expired

docs/rfc/0002-identity_proofs.md
```diff
@@ -0,0 +1,212 @@
...
+    * `expires`
+
+      Seconds relative to `created`, after which the claim should no longer be
+      considered.
+
+* `proof` (optional)
+
+  A URL to assist verification tooling in retrieving the proof from the external
+  system. This is mainly a convenience, and obviously requires creation of a new
+  revision after the fact.
```
cloudhead, line 100: Why after the fact?

docs/rfc/0002-identity_proofs.md
```diff
@@ -0,0 +1,212 @@
...
+
+Note that the "git" protocol specifier of `radicle-link` URNs is implied, that
+is, future version MUST treat the absence of a disambiguating value as denoting
+"git".
+
+The values `root`, `revision`, and `public-key` are specified in
+[identities][ids], and it is RECOMMENDED to follow the serialisation formats
+devised there. `signature` is the Ed25519 signature over `revision`, in much the
+same way as the actual revision is signed. All values can thus be obtained by
+inspecting the identity storage.
```
cloudhead, line 122: Perhaps the answer to the above question is missing here: the exact revision to be signed is the one which contains the claim, correct? Perhaps worth mentioning if that is the case.

docs/rfc/0002-identity_proofs.md
cloudhead: This almost seems like a non-feature, since it requires a new identity revision, I'm not sure if clients will deem it worthwhile to support :shrug:

docs/rfc/0002-identity_proofs.md
```diff
@@ -0,0 +1,212 @@
...
+
+It is beyond the scope of this document to devise the exact external format to
+serialise the tuple into, as this is expected to vary from service to service.
```
cloudhead, line 125: This might be a good candidate for a separate RFC with some proposed plain-text formats :thought_balloon:

docs/rfc/0002-identity_proofs.md
```diff
@@ -0,0 +1,212 @@
...
+   latest valid tip (recall that this is not necessarily the same as what the
+   ref points to). The proof is rejected if one of the following is true:
+
+   4.1 `root` does not match
+
+   4.2 the document does not contain a claim for `(SERVICE, account)`
+
+   4.3 the document contains a claim for `(SERVICE, account)`, has an
+       `expiration`, and `expiration.expires` is smaller than `time() -
+       expiration.created`
```
cloudhead, line 164: Is this missing a verification clause for revoked claims? Shouldn't we traverse the history chronologically?

docs/rfc/0002-identity_proofs.md
kim: Since there’s no central database, this would mean that you can only navigate from the external system to link, but not the other way round — which would be a bit lame I think.

Perhaps the proof signature could be over the previous revision + a random nonce 🤔 If space is constrained, maybe the preimage could be omitted entirely (except the root hash)?

docs/rfc/0002-identity_proofs.md
kim: Wondering if we should do something fun instead of the usual ones. Matrix? lobste.rs?

docs/rfc/0002-identity_proofs.md
kim: The `tree` at the latest tip either contains the claim, or it doesn’t (4.2). If it does, and is not expired (4.3), we need to find the revision which introduced it to verify that it is indeed an ancestor, and that the signing key was eligible at that point.

cloudhead: Ah I see, the instructions are a bit higher level than I understood.

docs/rfc/0002-identity_proofs.md
cloudhead: hahaha yes. mastodon, ssb?

docs/rfc/0002-identity_proofs.md
cloudhead: Yeah, true.

Signing the previous revision + account + nonce is not a bad idea :thinking: but seems like it may complicate implementations slightly due to the verification requiring two revisions as input instead of one?

I'm wondering in what scenarios the proof URL will be used? Because if there are legit use-cases for it, then it probably shouldn't be optional, and we should go with your idea.

So the typical use-case is that of sybil.org for example, where you can verify the statement: "I am @cloudhead and address 0xabcdef123 belongs to me". They must store the tweet id somewhere, since that's how they verify the signature. So yeah it seems kind of important :thinking:

kim: > the verification requiring two revisions as input instead of one?

Myeah, so:

## With Nonce

1. Post signature over `R || nonce`, store `URL`
2. Create `R+1` containing `nonce`, `SERVICE: {..., proof: Some(URL) }`
3. Sign `R+1` using `delegations` to "finalise" it

### Verify

1. Check signature, abort if invalid
2. Find `R+1`, check that
    1. is valid
    2. `R` is parent
    3. `R` has key in delegations
    4. `R+1` has claim
    5. `R+1` claim is not expired
3. Check `HEAD` still has the claim

vs.

## Without Nonce

1. Create `R` containing `SERVICE: { ..., proof: None }`
2. Post signature over `R`, store `URL`
3. Create `R+1` with `SERVICE: { ..., proof: Some(URL) }`
4. Sign `R` **and** `R+1` using `delegations` to finalise

### Verify

1. Check signature, abort if invalid
2. Find `R`, check that
    1. is valid
    2. has key in delegations
    3. has claim
    4. claim is not expired
3. Check `HEAD` still has the claim
  
So the issue with the nonce is that you need to publish the nonce, _or_ that you can only verify the signature if you know `R+1`'s _payload_. The issue without the nonce is that you need to sign two successive revisions for finalisation.

> I'm wondering in what scenarios the proof URL will be used?

It's when you start from `link` -- you'll want an O(1) method of getting at the proof payload.

> then it probably shouldn't be optional

Without a nonce, it's a chicken-and-egg problem :)


> sybil.org for example

sybil.org is simpler, because:

1. there is only **one** key (the address)
2. they basically store _all_ claims in a central database ([verified.json](https://github.com/Uniswap/sybil-list/blob/master/verified.json))
3. there is nothing to verify about the database itself (because 1.) 

I mean, the proposed proof URL doesn't technically need to be signed, either. But if it's not in the identity document, then where else could we put it? Annotate `R` with a [git-notes](https://git-scm.com/docs/git-notes) note?

PR #530, opened 2021-02-26
ci: Add Hüb workflow

kim:
Adds a GitHüb workflow running a minimal set of lints + tests on
proprietary platforms. Additionally, this "fast set" is run for pull
requests from other repositories, so that we can disable running the
(slow) builds for externals on our hardware.

### DIFF
--- a/.github/workflows/fast.yaml
+++ b/.github/workflows/fast.yaml
@@ -1,0 +1,95 @@
+name: fast
+on: [pull_request]
+jobs:
+  fmt:
+    if: github.repository_owner != 'radicle-dev'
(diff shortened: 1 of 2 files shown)


FintanH (dismissed):
Slick :ok_hand: Hopefully these toolchain values don't get out of sync :upside_down_face:

FintanH (approved):

xla:
Did they end up running for this PR?

kim:
Would anyone have expected that expression languages based on YAML are a pain in the buttocks?

FintanH (dismissed):

Issue #524, opened 2021-02-21
seed: inaccessible .gitconfig crashes seed


viraptor:
If `~/.gitconfig` cannot be read, the seed node fails to start with:

```
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: RequestFailed', /home/viraptor/radicle-bins/seed/src/frontend.rs:295:48
...
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Api(Pool(Backend(Git(Error { code: -1, klass: 7, message: "failed to stat \'/home/viraptor/.gitconfig\'" }))))'
```

Since a missing file is just ignored, it would be great if an inaccessible file is also skipped. (maybe just warned about?)

This can come up in a system with stricter security policies / sandboxed seed service.

kim:
I'm not exactly sure under which confinements this can be provoked, but possibly this piece of DWIM is resposible for it: https://github.com/libgit2/libgit2/blob/81c98af777329817827609a90462d8b2fd4a845b/src/repository.c#L1046

Since we are effectively only interested in the repo-scoped config, we could open the `Config` object by path.

PR #535, opened 2021-03-02
git: Use repo-local config only

kim:
By default, `git2::Repository::config` will search for the whole set of
layered config files git may want to respect. This isn't what we want
for `Storage`: only the repo-local config (we write ourselves) should be
considered.

Also, trying to resolve user/system configs seems to create issues in
some setups (see #524).

Fixes: #524


FintanH (approved):

xla (approved):
💩 👌 🌨 🕟

PR #538, opened 2021-03-02
WIP: prevent concurrent fetch of the same URN

kim:


librad/src/internal/klock.rs
```diff
@@ -0,0 +1,103 @@
...
+where
+    T: Hash,
+    H: BuildHasher,
+{
+    /// Acquires a lock for the given `key`, blocking the current thread until
+    /// it is able to do so.
+    ///
+    /// This method is reentrant: the current thread may acquire more than one
+    /// lock on the same or different keys.
+    pub fn lock(&self, key: &T) -> KlockGuard {
```
FintanH, line 78: So there are two cases where this will wait:
1. Trying to acquire a lock on the same URN on a different thread
2. All the slots are full

Is that correct?

kim:
> So there are two cases where this will wait:
> 1. Trying to acquire a lock on the same URN on a different thread
> 2. All the slots are full
>
> Is that correct?

Yes. I think. So far, I've only read the _documentation_ of `ReentrantMutex` :D

FintanH:
> Yes. I think. So far, I've only read the _documentation_ of `ReentrantMutex` :D

Hahaha that's how I came to this conclusion too :laughing:

PR #540, opened 2021-03-03
git: Prevent concurrent fetch of the same URN

kim:
It is in principle safe to fetch multiple packfiles concurrently, even
if their contents end up being identical. However, updating the refs
after a fetch is not an atomic operation, and even though `libgit2`
tries to detect concurrent modifications, operations which rely on
read-after-write semantics will not function properly.

To fix this, we effectively serialise `replicate` of the same logical
repository (ie. URN). There are several tradeoffs, as documented in the
code.

As for continuous brooming, the `git::fetch` module is broken up into
more submodules for reduced strain on scrolling devices.


librad/src/git/fetch/fetcher.rs
```diff
@@ -0,0 +1,179 @@
...
+        urn: git::Urn,
+        remote_peer: PeerId,
+        addr_hints: Addrs,
+    ) -> Result<Self, git2::Error>
+    where
+        Addrs: IntoIterator<Item = SocketAddr>,
+    {
+        // TODO: Revisit once we have busy grafting.
+        //
+        // Tradeoffs:
```
kim, line 60: There is also the forth possibility of `try_lock`ing here, that is fail-fast a `replicate` if one is already in progress. That's also not ideal, because the later `replicate` could actually be from a different remote peer (which may offer different data).

FintanH: Wouldn't a `try_lock` give some indication to wait and try again?

FintanH (dismissed):

librad/src/git/fetch/fetcher.rs
kim: Yes, but that means the problem is somewhere else + needs scrutinising the error variants. 

I think I’ll leave this patch queued until I have basic grafting up to see if this is practical.

FintanH: I'm thinking it's harder to spot a lock not being available and so the remote goes away compared to, we couldn't lock, so we wait and try again. But happy to wait and see :)

xla (approved):
utACK

🚶 🎉 🚷 🚌

librad/src/git/fetch/fetcher.rs
kim: You were right -- we should almost always abort

FintanH (approved):

kim:
I have amended this to use an elaborate tracking + waiting strategy. As per usual, the finicky part is to juggle blocking and async components, which _may_ turn out to yield undesirable fairness properties in practice (see commentary on `git::storage::fetcher::retrying`).

Note that, as a consequence, callers of `replication::replicate` are now required to BYO `Fetcher` (ie. decide themselves if an how they want to retry). I have punted on devising how to retry when using `Peer::using_storage`, and as a consequence the seed doesn't.

librad/src/git/storage/pool.rs
```diff
@@ -39,12 +39,24 @@ impl Deref for PooledRef {
     }
 }
 
+impl DerefMut for PooledRef {
```
kim, line 42: ach, unrelated.

I was trying to prevent same-thread access to the `DashMap` by borrowing `self` mutably, but that didn't work out.

librad/src/net/peer.rs
```diff
@@ -33,29 +33,54 @@ pub use storage::Storage as PeerStorage;
...
     pub protocol: protocol::Config,
-    pub storage_pools: PoolSizes,
+    pub storage: config::Storage,
 }
 
-#[derive(Clone, Copy)]
-pub struct PoolSizes {
-    /// Number of [`git::storage::Storage`] instances to pool for [`Peer`]
-    /// consumers.
+pub mod config {
```
FintanH, line 39: One (unrelated) general question I have is: when do we want to inline `mod`s into files vs creating a new file? The diffs when eventually moving to a file becomes kind of annoying to read so I think I fall into the preference of just going with the file structure. One exception I don't mind is the `pub mod error` pattern. What do yis think?

FintanH (approved):
This looks good to me :+1: Only had an unrelated question

librad/src/net/peer.rs
kim: I'd prefer to employ a discipline of doing shuffles in a separate commit (I can split this one if you want). It's just the nature of ML-style languages that one needs to cut the papyrus every so often, imho. Otherwise where do you stop? With enough code accumulated, you'll converge to Java-style one-type-per-file eventually, which is no longer navigable without IDE support.

FintanH: > I can split this one if you want

Nah, alles klar. Finding the right module split has always haunted me :sweat_smile: I can see pros on both sides of the argument :)

xla (dismissed):
🛐 📒 📈 🐌

PR #542, opened 2021-03-03
net: Relax membership check before discarding gossip as unsolicited

kim:
Gossip messages must only be accepted from peers in the partial view, as
otherwise it would be too easy to flood the network with random
messages, amplified by faithful peers. However, requiring active set
membership is too strict, as promotion / demotion is not synchronised.

Instead, we check for partial view inclusion regardless of status.

### DIFF
--- a/librad/src/net/protocol.rs
+++ b/librad/src/net/protocol.rs
@@ -495,6 +495,6 @@ where
     }
 
     fn is_member(&self, peer: &PeerId) -> bool {
-        self.is_active(peer)
+        self.is_known(peer)
     }
 }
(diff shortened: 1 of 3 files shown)


FintanH:
This makes sense but what I'd like to understand better is how these members could move between the active and passive sets. From what I could tell, the demotion happens on connection closes and also random selection. You mentioned IRL that the paper relies on TCP whereas our UDP sockets can seem like they've gone away. 

This would line up with Upstream saying that its connection to the seed is active but the seed saying that the connection is passive. Would I haven't looked into is how a connection might go the other way. Can the seed probe a "passive" connection and find that it's actually active?

kim:
@FintanH does #543 answer your wonderings?

FintanH:
> @FintanH does #543 answer your wonderings?

Answered :white_check_mark: 

Thanks for fixing both :blush:

FintanH (dismissed):

PR #543, opened 2021-03-03
net: Fix `Disconnect` not being sent on demotion

kim:
When a peer is demoted from the active view, the other end needs to be
notified of that. This was not the case, because past self outsmarted
himself.

Note that this does not invalidate #542, as gossip could be in-flight
before the `Disconnect` arrives.


FintanH (approved):
Ahaha the ol' wildcard making one miss a match arm :upside_down_face:

PR #544, opened 2021-03-03
Correct partition logic

FintanH:
We were adding the elements into the wrong sets (booleans gouged out my
eyes). This patch fixes the logic and adds some proptests to ensure I
know I what I'm doing (hint: I don't).

### DIFF
--- a/librad/src/git/replication.rs
+++ b/librad/src/git/replication.rs
@@ -245,7 +245,7 @@ where
                 },
             };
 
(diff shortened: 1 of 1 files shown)


xla (approved):
Thy radiant bools blind me.

🚗 📑 💃 🆔

kim (approved):
AWS took all my math, so :man_shrugging:

PR #545, opened 2021-03-03
protocol: expose memebership lists via tincan

xla:
To get an understanding where in the partial view a certain peer is the
protocol can be phoned now to return the list of active, known and
passive peers.

### DIFF
--- a/librad/src/net/protocol.rs
+++ b/librad/src/net/protocol.rs
@@ -174,6 +174,34 @@ impl TinCans {
         rx.await.unwrap_or_default()
     }
 
+    pub async fn membership(&self) -> event::downstream::MembershipInfo {
+        use event::{
+            downstream::{Info::*, MembershipInfo},
+            Downstream,
+        };
+
+        let (tx, rx) = tokio::sync::oneshot::channel();
+        let tx = Arc::new(Mutex::new(Some(tx)));
+
+        if let Err(tincan::error::SendError(e)) =
+            self.downstream.send(Downstream::Info(Membership(tx)))
+        {
+            match e {
+                Downstream::Info(Membership(reply)) => {
+                    reply
+                        .lock()
+                        .take()
+                        .expect("if chan send failed, there can't be another contender")
+                        .send(MembershipInfo::default())
+                        .ok();
+                },
+                _ => unreachable!(),
+            }
+        }
+
+        rx.await.unwrap_or_default()
+    }
+
     pub async fn stats(&self) -> event::downstream::Stats {
         use event::{
             downstream::{Info::*, Stats},
--- a/librad/src/net/protocol/event.rs
+++ b/librad/src/net/protocol/event.rs
@@ -42,7 +42,14 @@ pub mod downstream {
     #[derive(Clone)]
     pub enum Info {
         ConnectedPeers(Reply<Vec<PeerId>>),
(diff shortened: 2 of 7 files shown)


librad/src/net/protocol/event.rs
```diff
@@ -42,9 +42,17 @@ pub mod downstream {
...
     pub enum Info {
         ConnectedPeers(Reply<Vec<PeerId>>),
+        Membership(Reply<MembershipInfo>),
         Stats(Reply<Stats>),
     }
 
+    #[derive(Clone, Debug, Default)]
+    pub struct MembershipInfo {
+        pub active: Vec<PeerId>,
+        pub known: Vec<PeerId>,
```
kim, line 52: Why this? `known` == `active + passive`

librad/src/net/protocol.rs
```diff
@@ -174,6 +174,34 @@ impl TinCans {
         rx.await.unwrap_or_default()
     }
 
+    pub async fn membership(&self) -> event::downstream::MembershipInfo {
```
FintanH, line 177: My only critique is that this is pretty much the same as the above `connected_peers` function. It could be useful to just abstract it out and then it's:
* Build channel
* Send message
* Expect response + use default otherwise

Might make it easier to add any future cases. Lemme know what yis think

seed/src/handle.rs
```diff
@@ -28,6 +28,16 @@ impl NodeHandle {
         Self { channel }
     }
 
+    pub async fn get_membership(&mut self) -> Result<MembershipInfo, NodeError> {
```
FintanH, line 31: Just noticing these `get_*` functions are similar too

librad/src/net/protocol.rs
kim: Actually... `connected_peers` is subtly different: it considers _all_ connections. It is permissible to talk to a node but not participate in gossip (eg. only `git`, or some to-come info API).

Did someone say write more docs?
image

xla: While with this addition the rule of 3 applies, I'm not sure how much we gain from the DRYness here.

xla: Can we leave this for a future exercise?

FintanH: Does that only differ in the receiver of `ConnectedPeers`? When I diff these two function blocks:

```
5c4
<     self.downstream.send(Downstream::Info(Membership(tx)))
---
>     self.downstream.send(Downstream::Info(ConnectedPeers(tx)))
8c7
<         Downstream::Info(Membership(reply)) => {
---
>         Downstream::Info(ConnectedPeers(reply)) => {
13c12
<                 .send(MembershipInfo::default())
---
>                 .send(vec![])
19d17
< 
```

seed/src/handle.rs
xla: Same as above.

librad/src/net/protocol.rs
kim: Oh code structure. I hear the preferred way to do this in Rust is to make a macro ... ahm wait

:P

kim (approved):

librad/src/net/protocol.rs
FintanH: I'm fine with how it is

PR #549, opened 2021-03-04
git-helpers: Add informative git-remote-rad error message

### LINKED ISSUE #471: Improve git remote helper error message
Currently the git remote helper panics if it is run stand-alone from the CLI.

```
$ git-remote-rad
thread 'main' panicked at 'index out of bounds: the len is 0 but the index is 0', /Users/rudolfs/.cargo/git/checkouts/radicle-link-009c17032bc5eda9/308d2dd/git-helpers/src/remote_helper.rs:46:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

We should show something more informative like:

```
This remote helper is transparently used by Git when you use commands
such as "git fetch <URL>", "git clone <URL>", , "git push <URL>" or
"git remote add <nick> <URL>", where <URL> begins with "rad://".
```

[Link](https://git-scm.com/docs/git-remote-ext).

Refs: https://github.com/radicle-dev/radicle-upstream/issues/1466.

polyedre:
This pull-request apply the suggestion of #471 and adds a more informative message when `git-remote-rad` is ran from the command line directly.

The error message is now

```
Error: This remote helper is transparently used by Git when you use commands
such as "git fetch <URL>", "git clone <URL>", "git push <URL>" or
"git remote add <nick> <URL>", where <URL> begins with "rad://".
See https://git-scm.com/docs/git-remote-ext for more detail.
``` 

instead of

```
thread 'main' panicked at 'index out of bounds: the len is 0 but the index is 0', /cache/cargo/git/checkouts/radicle-link-009c17032bc5eda9/308d2dd/git-helpers/src/remote_helper.rs:46:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Closes #471

### DIFF
--- a/git-helpers/src/remote_helper.rs
+++ b/git-helpers/src/remote_helper.rs
@@ -31,6 +31,14 @@ const SECRET_KEY_FILE: &str = "librad.key";
 pub fn run() -> anyhow::Result<()> {
     let url = {
         let args = env::args().skip(1).take(2).collect::<Vec<_>>();
+        if args.is_empty() {
+            return Err(anyhow::anyhow!(
+                r#"This remote helper is transparently used by Git when you use commands
+such as "git fetch <URL>", "git clone <URL>", "git push <URL>" or
+"git remote add <nick> <URL>", where <URL> begins with "rad://".
+See https://git-scm.com/docs/git-remote-ext for more detail."#
+            ));
+        }
         args[0]
             .parse()
             .or_else(|_| args[1].parse())


git-helpers/src/remote_helper.rs
```diff
@@ -31,6 +31,14 @@ const SECRET_KEY_FILE: &str = "librad.key";
 pub fn run() -> anyhow::Result<()> {
     let url = {
         let args = env::args().skip(1).take(2).collect::<Vec<_>>();
+        if args.is_empty() {
+            return Err(anyhow::anyhow!(
```
xla, line 35: You can avoid the escapes in the string by using the [Raw string literals](https://doc.rust-lang.org/reference/tokens.html#raw-string-literals)

xla (changes requested):
Thanks for improving the developer experience. See my inline comment for how to make the string a bit easier to read and maintain.

git-helpers/src/remote_helper.rs
polyedre: Thanks for the review ! I did not know which one to choose. I made the change !

xla (approved):
🌮 🦂 🔖 🚑

kim (approved):

kim:
Thanks!

PR #550, opened 2021-03-05
git::identities: add root functions

FintanH:
We add `root` functions for the `Person` and `Project` documents. This
allows us to have a top-level function that will give us the expected
`Urn` for the given payload and delegates.

### DIFF
--- a/librad/src/git/identities/person.rs
+++ b/librad/src/git/identities/person.rs
@@ -79,6 +79,16 @@ pub fn verify(storage: &Storage, urn: &Urn) -> Result<Option<VerifiedPerson>, Er
     }
 }
 
+/// Get the root [`Urn`] for the given `payload` and set of `delegations`.
+#[tracing::instrument(level = "debug", skip(storage), err)]
+pub fn urn<P>(storage: &Storage, payload: P, delegations: delegation::Direct) -> Result<Urn, Error>
+where
+    P: Into<PersonPayload> + Debug,
+{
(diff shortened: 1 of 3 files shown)


kim:
This is great.

I think the terminology is confusing, though: there is no such thing as a "root URN" -- the URN stays the same always, `root` is basically pronounced "root hash". So I think public functions which compute the URN from the `Doc` should be called `compute_urn` (or `urn` for short).

I apologise also for using the `Revision` alias to annotate root hashes -- it's not actually a revision, because it's only a blob hash, not a tree hash (happens to be the same type in git).

FintanH:
> I think the terminology is confusing, though: there is no such thing as a "root URN" -- the URN stays the same always, root is basically pronounced "root hash". So I think public functions which compute the URN from the Doc should be called compute_urn (or urn for short).

Makes sense :+1: For the lower-level function, maybe we could call it `base` since it's what kicks off the `Doc` and `Revision`? Or is that fine as `root`?

FintanH:
> For the lower-level function, maybe we could call it base since it's what kicks off the Doc and Revision? Or is that fine as root?

i.e. these ones https://github.com/radicle-dev/radicle-link/pull/550/files#diff-dd0e454aece5943bf4baa7adc7cb5b63ce0f4700f583961b17fc624dc9e9fd7bR503-R518

xla (dismissed):
🏋 🚵 💯 🕵

kim (approved):
:+1:

PR #552, opened 2021-03-05
identities: Split list into `list` and `list_urns`

kim:
Sometimes it is useful to only list the top-level URNs without accessing
any objects. `list` can then be implemented in terms of `list_urn` +
storage access.


xla (approved):
👺 🔍 🐲 🛡

FintanH (dismissed):
Sleek

PR #554, opened 2021-03-05
git: Thread through the updated tips when `replicate`-ing

kim:
This will help reasoning about what actually happened: we expect certain
refs to be updated during `replicate`, but we couldn't tell before if
they were. Also, no information about any updated "user branches" was
propagated.

### DIFF
--- a/librad/src/git/replication.rs
+++ b/librad/src/git/replication.rs
@@ -82,26 +82,26 @@ pub struct Config {
     pub fetch_limit: fetch::Limit,
 }
 
-pub enum Replication {
-    Clone {
-        urn: Urn,
-        identity: SomeIdentity,
-        fetched_peers: BTreeSet<PeerId>,
-    },
-    Fetch {
-        urn: Urn,
-        identity: SomeIdentity,
-        existing: BTreeSet<PeerId>,
-    },
+/// The success outcome of [`self::replicate`].
+pub struct ReplicateResult {
+    /// The set of refs which were updated during the sync, along with the
+    /// [`ext::Oid`] they are now pointing to.
+    ///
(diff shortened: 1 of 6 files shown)


librad/src/git/replication.rs
```diff
@@ -82,26 +82,26 @@ pub struct Config {
...
-        identity: SomeIdentity,
-        fetched_peers: BTreeSet<PeerId>,
-    },
-    Fetch {
-        urn: Urn,
-        identity: SomeIdentity,
-        existing: BTreeSet<PeerId>,
-    },
+/// The success outcome of [`self::replicate`].
+pub struct ReplicateResult {
```
FintanH, line 86: It would be useful if we distinguished between `Cloned` and `Fetched`. For example, this would allow us to distinguish that a project was cloned fresh and we expect to see all the `rad/*` refs.

kim: How would you go about this? Just add another field `what_happened: ClonedOrFetched`?

FintanH: Ya, that would probably be the easiest. Maybe `enum ReplicationStatus` to mirror the `IdStatus` naming

FintanH: Did GH mess up your flow? :sweat_smile:

kim: No I don't think -- this is what ended up in master: fd7cbb2e9f2ed1011f1d044f745b343f0def96c3

FintanH: Ach so, then we're not going for this extra field? And are we following up on the `Even`/`Uneven` result for `person` in another patch?

kim: Yeah I’ll add that later

PR #557, opened 2021-03-05
RFC: Ethereum attestation

CodeSandwich:
This is an RFC and a documentation for Ethereum attestation strategy. It's built on top of https://github.com/radicle-dev/radicle-link/pull/525 and supports https://github.com/radicle-dev/radicle-upstream/issues/965. The implementation of the smart contract is here: https://github.com/radicle-dev/radicle-contracts/pull/129.

### DIFF
--- a/docs/rfc/ethereum_attestation.md
+++ b/docs/rfc/ethereum_attestation.md
@@ -9,8 +9,8 @@
 
 The attestation between Link and Ethereum is a valuable building block for a user identity.
 It brings the Link reputation coming from projects and contributions to the
-Ethereum world of DAOs and donations, where it's important to know, who's behind an address.
-On the other hand it lends Ethereum account reputation with its assets and undeniable history
+Ethereum world of DAOs and donations, where it's important to know who's behind an address.
+On the other hand, it lends Ethereum account reputation with its assets and undeniable history
 to Link to build user's trust in the identity.
 
 ## Overview
@@ -26,7 +26,7 @@ Under this key there is stored an ethereum address claim following this conventi
 
 - `account` - the claimed ethereum address, encoded according to [EIP-55][eip-55],
 e.g. using [ethers.js][ethers-addr]
-- `proof` - always `null`
+- `proof` - not set (according to [Identity Proofs RFC][rfc] it's an optional field)
 
 Example:
 ```json
@@ -36,15 +36,14 @@ Example:
         "expiration": {
             "created": 1614768373,
             "expires": 31536000
-        },
-        "proof": null
+        }
     }
 }
 ```
 
 ## Ethereum smart contract
 
-A new smart contract is deployed to the network, which is used to publish link claims:
+A new Ethereum smart contract is deployed to the network, which lets users claim their Radicle IDs:
 
 ```solidity
 contract Claims {
@@ -53,7 +52,7 @@ contract Claims {
 }
 ```
 
-In order to claim an ID, call `claim` using your Ethereum account.
+To claim an ID, call `claim` using your Ethereum account.
 It will emit an event `Claimed`, which later can be queried to discover your attestation.
 The claims have no expiration date and don't need to be renewed.
 
@@ -67,23 +66,24 @@ Currently supported `version` values:
 `000000000000000000000000fb3102b74d7254eed7f18a31a3ba1ea946bb1a99`
 - `2` - an `id` is a SHA-256 root hash
 
-Everybody must use the same smart contract address.
-Claims sent to an address not known by the users won't be noticed by anyone and will be useless.
+We need to deploy an official instance of the `Claims` smart contract and
+it must be used by all the users.
+If anybody makes a claim using a different instance, it won't be recognized by others.
 
 ## Creation of an attestation
 
 You need to perform 2 actions in any order:
 - Add or update an `ethereum` entry in your identity JSON.
 The entry's `account` must be your Ethereum address.
-It's higly recommended to set a reasonable expiration date as Ethereum claims don't expire.
+It's highly recommended to set a reasonable expiration date as Ethereum claims don't expire.
 - Call `claim` in the `Claims` smart contract. The `id` must point to your link identity.
 
 ## Discovery from an Ethereum address
 
 When you have an Ethereum address, you can find the claimed link ID using an Ethereum client.
 The example calls are based on the standard [client JSON RPC API][rpc] and should be exposed
-by your favorite Ethereum client library.
-It is important that the client must be trusted not to hide the events.
+by your favourite Ethereum client library.
+It's important that the client must be trusted not to hide the events.
 
 - Use [getLogs][rpc-logs] to get the newest `Claimed` event filtered for the given ethereum address
 - Get the event's `transactionHash` field and use it to fetch the transaction which emitted it with
@@ -113,7 +113,7 @@ To revoke a claim on Link, update and publish the identity JSON.
 You can change the claimed Ethereum address or remove the `ethereum` section altogether.
 
 To revoke a claim on Ethereum, call the `claim` function in `Claims` contract.
-You can claim a different link ID or an ID `0` to stop claiming anything.
+You can claim a different link ID or an ID `0` to revoke any claim you may have.
 
 ---
 
--- a/docs/rfc/ethereum_attestation.md
+++ b/docs/rfc/ethereum_attestation.md
@@ -15,29 +15,30 @@ to Link to build user's trust in the identity.
 
 ## Overview
 
-This RFC is built on top of [Identity Proofs RFC][rfc].
+This RFC is built on top of [identities spec][identities].
 It introduces support for Ethereum address claims on Link
 and a smart contract on Ethereum to make Link identity claims on Ethereum.
 
 ## Link identity JSON extension
 
-The identity JSON supports a new key: `ethereum`.
+The identity JSON `payload` structure supports a new key: `https://radicle.xyz/ethereum/claim/v1`.
 Under this key there is stored an ethereum address claim following this convention:
 
-- `account` - the claimed ethereum address, encoded according to [EIP-55][eip-55],
+- `address` - the claimed ethereum address, encoded according to [EIP-55][eip-55],
 e.g. using [ethers.js][ethers-addr]
-- `proof` - not set (according to [Identity Proofs RFC][rfc] it's an optional field)
+- `expiration` - the claim expiration timestamp, encoded as a [JavaScript Date][date]
 
 Example:
 ```json
 {
-    "ethereum": {
-        "account": "0xAb5801a7D398351b8bE11C439e05C5B3259aeC9B",
-        "expiration": {
-            "created": 1614768373,
-            "expires": 31536000
+    "payload": {
+        "https://radicle.xyz/ethereum/claim/v1": {
+            "account": "0xAb5801a7D398351b8bE11C439e05C5B3259aeC9B",
+            "expiration": "2021-03-19T23:15:30.000Z"
         }
+        ...
     }
+    ...
 }
 ```
 
@@ -73,9 +74,9 @@ If anybody makes a claim using a different instance, it won't be recognized by o
 ## Creation of an attestation
 
 You need to perform 2 actions in any order:
-- Add or update an `ethereum` entry in your identity JSON.
+- Add or update an `https://radicle.xyz/ethereum/claim/v1` entry in your identity JSON.
 The entry's `account` must be your Ethereum address.
-It's highly recommended to set a reasonable expiration date as Ethereum claims don't expire.
+It's highly recommended to set a short expiration date as Ethereum claims don't expire.
 - Call `claim` in the `Claims` smart contract. The `id` must point to your link identity.
 
 ## Discovery from an Ethereum address
@@ -98,7 +99,7 @@ see [Discovery from a Link ID](#discovery-from-a-link-id)
 
 When you have a link ID, you can find the claimed Ethereum address.
 Obtain the tip of its identity chain and read the ethereum address from the identity JSON
-`account` field in section `ethereum`, unless it's expired.
+`account` field in section `https://radicle.xyz/ethereum/claim/v1`, unless it's expired.
 You need to verify that the given Ethereum address claims back the link ID,
 see [Discovery from an Ethereum address](#discovery-from-an-ethereum-address).
 
@@ -110,15 +111,17 @@ Only one claim needs to be revoked to break the attestation,
 but to improve security you should revoke both sides if you can.
 
 To revoke a claim on Link, update and publish the identity JSON.
-You can change the claimed Ethereum address or remove the `ethereum` section altogether.
+You can change the claimed Ethereum address or remove
+the `https://radicle.xyz/ethereum/claim/v1` section altogether.
 
 To revoke a claim on Ethereum, call the `claim` function in `Claims` contract.
 You can claim a different link ID or an ID `0` to revoke any claim you may have.
 
 ---
 
-[rfc]: ./identity_proofs.md
+[identities]: ../spec/sections/002-identities/index.md
 [eip-55]: https://eips.ethereum.org/EIPS/eip-55
+[date]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toJSON
 [ethers-addr]: https://docs.ethers.io/v5/api/utils/address/
 [rpc]: https://eth.wiki/json-rpc/API
 [rpc-logs]: https://eth.wiki/json-rpc/API#eth_getlogs
--- a/docs/rfc/ethereum_attestation.md
+++ b/docs/rfc/ethereum_attestation.md
@@ -33,7 +33,7 @@ Example:
 {
     "payload": {
         "https://radicle.xyz/ethereum/claim/v1": {
-            "account": "0xAb5801a7D398351b8bE11C439e05C5B3259aeC9B",
+            "address": "0xAb5801a7D398351b8bE11C439e05C5B3259aeC9B",
             "expiration": "2021-03-19T23:15:30.000Z"
         }
         ...
@@ -75,7 +75,7 @@ If anybody makes a claim using a different instance, it won't be recognized by o
 
 You need to perform 2 actions in any order:
 - Add or update an `https://radicle.xyz/ethereum/claim/v1` entry in your identity JSON.
-The entry's `account` must be your Ethereum address.
+The entry's `address` must be your Ethereum address.
 It's highly recommended to set a short expiration date as Ethereum claims don't expire.
 - Call `claim` in the `Claims` smart contract. The `id` must point to your link identity.
 
@@ -99,7 +99,7 @@ see [Discovery from a Link ID](#discovery-from-a-link-id)
 
 When you have a link ID, you can find the claimed Ethereum address.
 Obtain the tip of its identity chain and read the ethereum address from the identity JSON
-`account` field in section `https://radicle.xyz/ethereum/claim/v1`, unless it's expired.
+`address` field in section `https://radicle.xyz/ethereum/claim/v1`, unless it's expired.
 You need to verify that the given Ethereum address claims back the link ID,
 see [Discovery from an Ethereum address](#discovery-from-an-ethereum-address).
 
--- a/docs/rfc/ethereum_attestation.md
+++ b/docs/rfc/ethereum_attestation.md
@@ -15,7 +15,7 @@ to Link to build user's trust in the identity.
 
 ## Overview
 
-This RFC is built on top of [identities spec][identities].
+This RFC is built on top of the [identities spec][identities].
 It introduces support for Ethereum address claims on Link
 and a smart contract on Ethereum to make Link identity claims on Ethereum.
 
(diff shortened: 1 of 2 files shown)


kim:
Is it feasible in practice to get a signature using the transaction key over some arbitrary data?

With this, we are back to a system which is only usable if:

* `upstream` is installed, configured, and **online**
* the web3 provider is available
* the link identity is available and consistent (modulo the expiration window)

In other words, it’s an online verification. I understand how limited Ethereum is, but this just isn’t very useful.

CodeSandwich:
@kim 

> get a signature using the transaction key over some arbitrary data

I'm sorry, but I'm not sure what do you mean by that, could you explain that?

> In other words, it’s an online verification.

Yes, it is and it can't be different to have trust in it. Unless you have a fresh view of both networks, you don't know if the other side still hasn't revoked its claim.

NunoAlexandre:
> > In other words, it’s an online verification.
> 
> Yes, it is and it can't be different to have trust in it. Unless you have a fresh view of both networks, you don't know if the other side still hasn't revoked its claim.

Ya, this was something we discussed with @cloudhead and agreed on the tradeoff.  

> but this just isn’t very useful.

How do you mean 'useful', @kim?

docs/rfc/ethereum_attestation.md
```diff
@@ -0,0 +1,126 @@
...
+* Author: @CodeSandwich
+* Date: 2021-03-05
+* Status: draft
+* Community discussion: n/a
+
+## Motivation
+
+The attestation between Link and Ethereum is a valuable building block for a user identity.
+It brings the Link reputation coming from projects and contributions to the
+Ethereum world of DAOs and donations, where it's important to know, who's behind an address.
```
NunoAlexandre, line 12:
```suggestion
Ethereum world of DAOs and donations, where it's important to know who's behind an address.
```

docs/rfc/ethereum_attestation.md
```diff
@@ -0,0 +1,126 @@
...
+On the other hand it lends Ethereum account reputation with its assets and undeniable history
```
NunoAlexandre, line 13:
```suggestion
On the other hand, it lends Ethereum account reputation with its assets and undeniable history
```

docs/rfc/ethereum_attestation.md
```diff
@@ -0,0 +1,126 @@
...
+            "expires": 31536000
+        },
+        "proof": null
+    }
+}
+```
+
+## Ethereum smart contract
+
+A new smart contract is deployed to the network, which is used to publish link claims:
```
NunoAlexandre, line 47: This sounds a little to vague to me. I'd recommend be more specific and say something like:

```suggestion
A new Ethereum smart contract is deployed to the network, which allows users to claim their Radicle ID.
```

docs/rfc/ethereum_attestation.md
```diff
@@ -0,0 +1,126 @@
...
+which is guaranteed to not match any existing identity.
+
+Currently supported `version` values:
+- `1` - an `id` is a SHA-1 root hash. The excess high bytes are zeros, e.g. for hash
+`fb3102b74d7254eed7f18a31a3ba1ea946bb1a99` the passed `id` is
+`000000000000000000000000fb3102b74d7254eed7f18a31a3ba1ea946bb1a99`
+- `2` - an `id` is a SHA-256 root hash
+
+Everybody must use the same smart contract address.
+Claims sent to an address not known by the users won't be noticed by anyone and will be useless.
```
NunoAlexandre, line 71: > an address not known by the users won't be noticed by anyone and will be useless.

Could rephrase to in a way so that it transmits the idea of there being an official contract address that is the only one taken into account?

docs/rfc/ethereum_attestation.md
```diff
@@ -0,0 +1,126 @@
...
+When your attestation for whatever reason is no longer valid,
+you should revoke it as soon as possible.
+Only one claim needs to be revoked to break the attestation,
+but to improve security you should revoke both sides if you can.
+
+To revoke a claim on Link, update and publish the identity JSON.
+You can change the claimed Ethereum address or remove the `ethereum` section altogether.
+
+To revoke a claim on Ethereum, call the `claim` function in `Claims` contract.
+You can claim a different link ID or an ID `0` to stop claiming anything.
```
NunoAlexandre, line 116: > to stop claiming anything.

could we rephrase so something like:

"to revoke any claim you may have."

NunoAlexandre (commented):
Looks fair to the plan to me :popcorn: 

Before approving, I want to understand what Kim means by an online-only verification setup not being useful.

kim:
I'm thinking that we didn't think through our threat model properly. Consider this:

Say we have established the link `ethereum <-> radicle-link` somehow. Now, an attacker gets access to the ethereum private key. What is she going to do about the attestation? 

**Nothing of course!** 

Funds are flowing in to the address because of the reputation of the off-chain identity, so there's no reason to change it to a different one. Ideally, the attacker finds a way to transfer funds away without anyone noticing, so the money keeps coming.

When this happens, and we still have access to the compromised key, then the effect of revoking the registration should be that transfers which get triggered by some other contract **are cancelled**, and it is no longer possible to **set up new funding** (unless one uses the address directly, of course). Revoking the off-chain identity is a weak defense, because **it can't be checked on-chain**. I'm not sure how things are set up, would this be possible?

Ok, what about the other way round: the off-chain identity is compromised? In this case, the attacker would want to change the ethereum address to one she has control over. Because she can't change the root hash (without losing the identity) **and** the mapping is 1:1, that's impossible. We can recover from this by pointing the ethereum account to a fresh off-chain identity.

One problem with this is that we **shall under no circumstances lose the ethereum private key**. 

That's probably just generally a weakness of how ethereum accounts work, but to drive the point home: if we'd introduce expiration for the registration contract, the attacker who already controls the off-chain identity just needs to sit out the expiration window and register with a different account.

---

If this all makes sense, then it looks like we can spare ourselves to look up the registration from the chain tip _just for verification_. We only need to do this when we actually want to set up a transfer / funding contract. Which means that we could just store the _transaction receipt_ (instead of only the hash) of the registration off-chain, and verify signatures offline.

Thoughts?

CodeSandwich:
The smart contracts aren't aware of attestations and even if they were, they wouldn't be able to verify them. That means that the off-chain software bears the full responsibility of checking the attestations and telling the smart contracts to do something in case of an emergency. Does it address your doubts?

kim:
> That means that the off-chain software bears the full responsibility of checking the attestations and telling the smart contracts to do something in case of an emergency. Does it address your doubts?

No. If that responsibility is off-chain, then you either didn't understand what I wrote, or this feature is completely pointless and I'm vetoing it.

docs/rfc/ethereum_attestation.md
```diff
@@ -0,0 +1,126 @@
...
+and a smart contract on Ethereum to make Link identity claims on Ethereum.
+
+## Link identity JSON extension
+
+The identity JSON supports a new key: `ethereum`.
+Under this key there is stored an ethereum address claim following this convention:
+
+- `account` - the claimed ethereum address, encoded according to [EIP-55][eip-55],
+e.g. using [ethers.js][ethers-addr]
+- `proof` - always `null`
```
cloudhead, line 29: This is optional, so if we aren't using it, it shouldn't be set?

docs/rfc/ethereum_attestation.md
```diff
@@ -0,0 +1,126 @@
...
+```solidity
+contract Claims {
+    event Claimed(address indexed addr);
+    function claim(uint8 version, uint256 id) public;
+}
+```
+
+In order to claim an ID, call `claim` using your Ethereum account.
+It will emit an event `Claimed`, which later can be queried to discover your attestation.
+The claims have no expiration date and don't need to be renewed.
```
cloudhead, line 58: This means that the link can be revoked with only the ethereum account, right? Is that desirable? Basically, since the Link identity doesn't point to a specific attestation, it can be revoked from the ethereum side, without having the Link key.

CodeSandwich: That's right, if you loose your Link key, you need to be able to revoke the attestation using only your Ethereum key.

cloudhead:
The transaction payload could be stored in Link, it would then be possible to verify the signature offline. Of course, you'd still eventually have to check that this transaction is included in a block, but at least you'd know that this ethereum key signed this radicle id?

CodeSandwich:
@cloudhead
Yes, we've discussed the offline verification. The problem with it is that unless you query the Ethereum client, you don't know if the claim has been published or if it was revoked. We came to a conclusion that a proof this weak isn't reliable enough to be useful and isn't worth the added complexity.

kim:
On Mon Mar 8, 2021 at 11:11 AM CET, Alexis Sellier wrote:
> The transaction payload could be stored in Link, it would then be
> possible to verify the signature offline. Of course, you'd still
> eventually have to check that this transaction is included in a block,
> but at least you'd know that this ethereum key signed this radicle id?

The point is that this only matters when you perform some action on-chain, at
that precise moment. Even if you verify (with a known link identity) online,
that doesn't provide you with any certainty that you can from now on safely
transact. At no point in time is there a guarantee that you will **ever**
receive the link identity, so it is unsound to base the on-chain security on
a bunch of off-chain signatures. Instead, on-chain transactions which go through
our contracts must check the claim status, and fail if there is none.

I suggest to drop this feature for now, and have it designed by a real
cryptographer.

CodeSandwich:
> Instead, on-chain transactions which go through our contracts must check the claim status, and fail if there is none.

How could that be done? We could store on-chain a signature made with the Link key, but unless the Ethereum runtime can query Link identity chain, it wouldn't prove anything. The only actor, who can reasonably well tell, what's the state of the world is off-chain software and IMO that's where we should decide if an attestation is fine.

CodeSandwich:
After considering it, we could have a scheme, where we would keep Link signatures on-chain. We could let anybody revoke any of these signatures as long as the transaction holds a payload signed by the revoked Link key. In that case on-chain contracts would be able to tell if the attestation is complete as long as Link posts these revocations to both Ethereum and identity chain :thinking:

kim:
Let me phrase it differently one more time: an ethereum address is just a key pair, right? So, if you can produce a signature, you can prove that you own it. A `radicle-link` root hash, however, is **not** a key pair -- this is by design, because otherwise you will have to keep the secret key in "secure offline storage", which people are notoriously bad at. Instead, we have a "survivable key compromise" scheme, where you can secure your identity with a k-of-n multisig.

The problem is that there is now no way I could think of which would allow a concise proof of ownership to be stored on-chain -- even if you put a multisig there, the set of keys that produced it may have been revoked in the meantime. This means that the only guarantee we have is that there is only one ethereum address which can claim a URN at any point in time. Because the reverse direction can't be verified on-chain, it is basically pointless: if you don't know the latest off-chain revision, you may still send money to the wrong address.

We said that anchoring with the goal of stopping progress if there is proof that a later revision of a link ID exists is a separate concern, and that it is not obvious how to solve it without a bulk-anchoring service. However, without solving it first, I do not see how the idea of attestations makes any sense.

cloudhead:
> I suggest to drop this feature for now, and have it designed by a real
cryptographer.

How about the subset of this feature which is not a proof of ownership, but just a "profile link"?

Eg these claims:

* "As the provable owner of this Radicle ID, this is my ethereum address"
* "As the provable owner of this Ethereum account, this is my Radicle ID"

And basically we DON'T do the two-way attestation. It's just a one way link, which serves the product purpose?

kim:
Sure. This doesn't require any addition to the `radicle-link` protocol, though. The claim doesn't even need to be stored inside the identity doc, it only needs to be signed using one or more of the delegate keys. This has the added benefit of it being distributable in any way one would like.

CodeSandwich:
Where would you see the claim stored on Link side?

kim:
> Where would you see the claim stored on Link side?

`refs/heads/cloudhead-ethereum`

or

`refs/heads/main` -> `.ethereum/cloudhead`


for example

CodeSandwich:
Ok, what are advantages and disadvantages of storing the claim there instead of the identity JSON?

kim:
You can publish it anywhere. If it wasn’t for the additional overhead of coming up with formal specifications of radicle “methods”, I’d even suggest to use a [DID](https://www.w3.org/TR/did-core) — they are exactly that: a way to claim different identities, where it is up to the consumer to verify the claims against their respective “methods”.

kim:
Ah scrap that, you’d want to retain a causal happens-before relationship to be able to revoke it. It’s still just `previous_payload.with_ext(eth_addr)`.

CodeSandwich:
@kim It looks very flexible and powerful, I think that it's worth implementing at some point! I also like the solution, which lets you undoubtedly decide on-chain, if the attestation is right, it's the ultimate solution of all solutions.

For now maybe we should stay pragmatic and come up with something that is simple, easy to implement and fills just our current use case. For example right now we don't need to store the claim in multiple places, we only need to look it up with Link. I think that if we stick with extending the identity JSON, we can reuse many existing tools, file formats and behaviors:
- we don't increase complexity with additional files
- we can reuse most of the infrastructure for lookup in the identity file, also it's a logical place to store an Ethereum claim
- we don't need to manage expirations separately
- we don't need to design, who can update the Ethereum claim and when

It's by no means a perfect solution, but for now it seems good enough. We can deprecate it and gradually migrate when we need to design the right solution.

docs/rfc/ethereum_attestation.md
CodeSandwich: Fixed

docs/rfc/ethereum_attestation.md
CodeSandwich: Fixed

docs/rfc/ethereum_attestation.md
CodeSandwich: Fixed

docs/rfc/ethereum_attestation.md
CodeSandwich: Fixed

docs/rfc/ethereum_attestation.md
CodeSandwich: Fixed

CodeSandwich:
That looks quite simple already, which is great. What are the advantages of moving the payload from the person identity to a separate payload? I've got a few issues with this approach:

- IIUC due to lack of support for Ethereum claims in `librad` in order to use it in 3rd party (e.g. in Upstream) we need to either import yet another library (which we need to write first) or reimplement the logic there
- we probably can't reuse the generic expiration logic built for identities
- separate claims in identity have separate local expiration dates anyway, how is a separate payload advantageous?

kim:
1. There is no relation between #525 and this one, they have very different semantics
2. There is no support for expirations as of now, so there isn't anything to _re_-implement
3. If and when #525 gets implemented, it will be a separate library / tool either way
4. This library is rather unlikely to ever include a web3 client (for "social proofs", a generic HTTP/S client suffices)
5. I have expressed my concerns about this feature _ad nauseam_, but I'm told this is a "product need". Since "product" == Upstream (and in fact, the feature is only usable with Upstream as it stands), it probably just belongs there.
6. Should there be a convincing solution to identity binding in the future, it will most likely work very differently to both #525 and this one, _but_ the old version will need to continue to work, possibly indefinitely. Instead of trying to guess from the shape of the payload which version applies, the code can just look for an unambiguous namespace.
7. The whole point of user-defined top-level namespaces is to make things extensible, and be able to come up with breaking semantics in the future.

Is that enough? ;)

CodeSandwich:
Thank you, that makes sense, I'll update the proposal :)

CodeSandwich:
@kim @NunoAlexandre Updated

docs/rfc/ethereum_attestation.md
```diff
@@ -15,29 +15,30 @@ to Link to build user's trust in the identity.
...
 ```json
 {
-    "ethereum": {
-        "account": "0xAb5801a7D398351b8bE11C439e05C5B3259aeC9B",
-        "expiration": {
-            "created": 1614768373,
-            "expires": 31536000
+    "payload": {
+        "https://radicle.xyz/ethereum/claim/v1": {
+            "account": "0xAb5801a7D398351b8bE11C439e05C5B3259aeC9B",
```
NunoAlexandre, line 36: s/account/address

docs/rfc/ethereum_attestation.md
```diff
@@ -98,7 +99,7 @@ see [Discovery from a Link ID](#discovery-from-a-link-id)
 
 When you have a link ID, you can find the claimed Ethereum address.
 Obtain the tip of its identity chain and read the ethereum address from the identity JSON
-`account` field in section `ethereum`, unless it's expired.
+`account` field in section `https://radicle.xyz/ethereum/claim/v1`, unless it's expired.
```
NunoAlexandre, line 102: s/account/address

NunoAlexandre (approved):
LGTM, just a nit to refer to the `address` field consistently.

docs/rfc/ethereum_attestation.md
CodeSandwich: Fixed

docs/rfc/ethereum_attestation.md
CodeSandwich: Fixed

NunoAlexandre (approved):

docs/rfc/ethereum_attestation.md
```diff
@@ -0,0 +1,129 @@
...
+
+The attestation between Link and Ethereum is a valuable building block for a user identity.
+It brings the Link reputation coming from projects and contributions to the
+Ethereum world of DAOs and donations, where it's important to know who's behind an address.
+On the other hand, it lends Ethereum account reputation with its assets and undeniable history
+to Link to build user's trust in the identity.
+
+## Overview
+
+This RFC is built on top of [identities spec][identities].
```
FintanH, line 18:
```suggestion
This RFC is built on top of the [identities spec][identities].
```

docs/rfc/ethereum_attestation.md
FintanH: The payload above has an expiration date. So this "claim" seems to be false :smirk:

docs/rfc/ethereum_attestation.md
```diff
@@ -0,0 +1,129 @@
...
+
+To claim an ID, call `claim` using your Ethereum account.
+It will emit an event `Claimed`, which later can be queried to discover your attestation.
+The claims have no expiration date and don't need to be renewed.
+
+Every new claim invalidates previous ones made with the same account.
+To only revoke a claim without creating a new one, use ID `0`,
+which is guaranteed to not match any existing identity.
+
+Currently supported `version` values:
```
FintanH, line 64: Is `version` supposed to be in the payload above? Or where did it come from?

FintanH: Or would you use `claim/v1` and `claim/v2` URLs?

docs/rfc/ethereum_attestation.md
```diff
@@ -0,0 +1,129 @@
...
+- `1` - an `id` is a SHA-1 root hash. The excess high bytes are zeros, e.g. for hash
+`fb3102b74d7254eed7f18a31a3ba1ea946bb1a99` the passed `id` is
+`000000000000000000000000fb3102b74d7254eed7f18a31a3ba1ea946bb1a99`
+- `2` - an `id` is a SHA-256 root hash
+
+We need to deploy an official instance of the `Claims` smart contract and
```
FintanH, line 70: How do we attest this? weh weh weh

docs/rfc/ethereum_attestation.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Discovery from a Link ID
+
+When you have a link ID, you can find the claimed Ethereum address.
+Obtain the tip of its identity chain and read the ethereum address from the identity JSON
+`address` field in section `https://radicle.xyz/ethereum/claim/v1`, unless it's expired.
+You need to verify that the given Ethereum address claims back the link ID,
+see [Discovery from an Ethereum address](#discovery-from-an-ethereum-address).
+
+## Revocation of an attestation
```
FintanH, line 106: It might be good to call out here that a revocation will only be as good as the propagation of the document in the network. What I mean is that if I revoke my address, I still have to wait for others to fetch those changes -- so _could_ still end up transferring MONIEZ to the old address.

FintanH (commented):
A couple of things aren't clear in this RFC:
1. Is this only regarding to `Person` documents?
2. "Link ID" is used a lot, but I'm unsure as to what that's actually referring to. Is it a URN? Is it the Conent ID? Is it something else?

CodeSandwich:
> The payload above has an expiration date. So this "claim" seems to be false smirk

The payload goes to Link identity JSON, the claim you're referring to is an Ethereum counterpart. The first has an expiration date, the latter not.

docs/rfc/ethereum_attestation.md
CodeSandwich: The payload is the Link-stored data, the `version` is used only in Ethereum. It has no corelation with the Link payload URL versioning.

docs/rfc/ethereum_attestation.md
CodeSandwich: We don't, there can be multiple instances deployed and it won't affect the overall security. If you start using a different instance, you probably won't notice other people's claims and they won't notice yours

docs/rfc/ethereum_attestation.md
CodeSandwich: Fixed

docs/rfc/ethereum_attestation.md
CodeSandwich: Added

docs/rfc/ethereum_attestation.md
FintanH: Ah, I missed it in the contract function :)

CodeSandwich:
@FintanH 
> "Link ID" is used a lot, but I'm unsure as to what that's actually referring to. Is it a URN? Is it the Conent ID? Is it something else?

Good point, I've updated the document to be more precise in context of the identities spec and its vocabulary. Now it should be clear, when we're using an identity root and when we're talking about an identity as an abstract concept of telling users apart.

CodeSandwich:
This RFC is no longer about Link, I'm closing this PR and moving it to Upstream

PR #558, opened 2021-03-08
replication: fetch rad/ids for remotes

FintanH:
We can't build a fetchspec that says `remote/*/rad/ids/*`. So once we've
fetched all remotes we make a second fetch call using `Peek` to get all
the `rad/ids/*` for all those remotes.

We confirm this works by extending the menage test to check for the
peer's `rad/ids` reference exists.


kim (dismissed):

FintanH:
Could I get another sanity check after the rebase? :pray:

kim (approved):
lgtm

PR #559, opened 2021-03-08
membership: ensure parity with active and passive members

FintanH:
When we add active we never checked to remove them from the passive
list. This means we end up double counting a peer.

### DIFF
--- a/librad/src/net/protocol/membership/partial_view.rs
+++ b/librad/src/net/protocol/membership/partial_view.rs
@@ -127,8 +127,14 @@ where
             return vec![];
         }
 
-        let demoted = if self.num_active() >= self.max_active {
+        let demoted = if self.is_active_full() {
             self.demote_random()
+        } else {
+            vec![]
+        };
+
+        let evicted = if self.is_passive(&info.peer_id) {
+            self.evict(&info.peer_id)
         } else {
             vec![]
         };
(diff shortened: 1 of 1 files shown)


FintanH:
One thing I'm thinking is that a `promote` function might make more sense, but without being too familiar I'd have to scope around the usages a bit to understand it better. Putting this up now for early feedback

xla (dismissed):
（╹◡╹）

kim:
> One thing I'm thinking is that a `promote` function might make more sense

There used to be one I think, but the only difference was that promotion implies that it was passive before, while `add_active` doesn't. So this is just splitting hairs on words, which I'm known to react to in grumpy ways.

librad/src/net/protocol/membership/partial_view.rs
```diff
@@ -127,12 +127,16 @@ where
...
 
-        let demoted = if self.num_active() >= self.max_active {
+        let demoted = if self.is_active_full() {
             self.demote_random()
         } else {
             vec![]
         };
 
+        if self.is_passive(&info.peer_id) {
+            self.passive.remove(&info.peer_id);
```
kim, line 137: This needs to add an `Evicted` transition to the return value, for example by calling `evict`.

kim (changes requested):

kim (dismissed):
:ok_hand:

PR #560, opened 2021-03-09
identities: Rectify divergence computations

kim:
Previously, both fork detection and selection of the most recent tip
among delegates was based on the commit history, which is incorrect --
delegates can maintain their own commit history, as long as the document
histories (ie. commit trees) don't diverge. It also turned out that we
wouldn't propagate whether the local `rad/id` branch is even with the
delegate view(s) for `Person` identities.

Additionally, return an indicator of what "mode" `replicate` was operating on
(clone / fetch).


FintanH (dismissed):
Looks good to me :+1:

librad/src/git/replication.rs
```diff
@@ -374,11 +403,20 @@ fn unsafe_into_urn(reference: Reference<git_ext::RefLike>) -> Urn {
     reference.try_into().expect("namespace is set")
 }
 
-#[allow(clippy::unit_arg)]
+/// Set the `rad/id` ref of `urn` to the given [`ext::Oid`].
+///
+/// No-op if the ref already exists. Returns the [`ext::Oid`] the ref points to
```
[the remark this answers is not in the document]
FintanH: Cool!

librad/src/git/replication.rs
```diff
@@ -539,22 +543,45 @@ mod person {
...
+            for pers in delegates.values().cloned() {
+                match prev {
+                    None => prev = Some(pers),
+                    Some(p) => {
+                        let newer = identities::person::newer(storage, p, pers)?;
+                        prev = Some(newer);
+                    },
+                }
+            }
+            prev.expect("empty delegations")
```
kim, line 572: Hrm... I was convinced that we couldn't get here with empty delegates, but maybe we can? There's a `NoTrustee` error variant, but it doesn't seem to get constructed ever (and `rustc` doesn't warn **/me shakes fist**)

FintanH: Ya, I just scanned through `identities::git` and I don't see any check on the delegations :grimacing:

kim: It cannot happen(tm), because we run `verify` before getting here. Just to be sure, I added a `cmp::max(1, ..)` to `quorum_threshold()`. I think I'll cook up a simple `NonEmptySet` next.

FintanH: :man_cook:

PR #565, opened 2021-03-11
seed: announce after track

FintanH:
Propagates that seed replicated a new peer for an existing `Urn`. We can likely consider removing this post-grafting/NAT'ing


xla (approved):
🆔 ↩️ 💇 😑

kim (approved):

PR #566, opened 2021-03-11
seed: expose config parameters

FintanH:
Stacked on #565 

Allows us to tune the seed's configuration parameters.

### DIFF
--- a/seed/src/lib.rs
+++ b/seed/src/lib.rs
@@ -12,7 +12,7 @@ pub mod handle;
 pub mod project;
 pub mod signer;
 
-use std::{collections::HashSet, net::SocketAddr, path::PathBuf, time::Duration};
+use std::{collections::HashSet, net::SocketAddr, time::Duration};
 
 use futures::{channel::mpsc as chan, select, sink::SinkExt as _, stream::StreamExt as _};
 use thiserror::Error;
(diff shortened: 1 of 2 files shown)


xla (dismissed):
🐪 🕔 🛃 🕧

seed/src/lib.rs
```diff
@@ -152,11 +162,11 @@ impl Node {
             protocol: protocol::Config {
                 paths,
                 listen_addr: config.listen_addr,
-                membership: Default::default(),
+                membership: config.membership.clone(),
```
kim, line 165: It's a bit odd to first consume the values, but then keep them around until `run`, which consumes `self` and obviously doesn't need those protocol parameters anymore.

Perhaps `peer_config` should just be built in `run`? Or `NodeConfig` should be consumed and `Node` should keep some other type with only the things it still needs?

FintanH: Ya, it feels weird indeed. Sorry I was cutting corners.

I think I'd like to make life easier for us here by moving the `radicle-bins/seed` over here and just have the exe there -- like you suggested. This dance is tiring. It would help with recognising these things, I think.

xla (approved):
🍵 📨 🈯️ 🤗

kim (approved):

Issue #531, opened 2021-02-26
spec: Stylez


kim:
For the so inclined, the CSS for the HTML-rendered spec could use some love. Also the `docs/index.html` page (which is deployed to GH pages). Additionally, it would be cool to have the `index.html` somehow be autogenerated, and include an index of the RFCs (rendered as HTML).

PR #567, opened 2021-03-14
Deploy documentation to github pages using github actions

polyedre:
This PR adds a Github action `documentation.yml` that build the html part of the documentation and publish it to Github pages.

Before merging this pull request, it is required to create the `gh-pages` branch and delete all the files in it:

```
git checkout --orphan gh-pages
git rm -rf .
```

The documentation has been deployed as an example at https://polyedre.github.io/radicle-link/ by [this job](https://github.com/polyedre/radicle-link/actions).

The content of the `gh-pages` branch is [here](https://github.com/polyedre/radicle-link/tree/gh-pages).

Note: The compilation of the pdf file is not included in the PR. I still need to figure what dependencies needs to be installed.

Reference #531


FintanH (approved):
This looks great :smile: Thanks for taking the time to do this

kim:
We already publish the `docs` directory as GH Pages. 

The reason it's set up this way is that `pandoc` rarely fails (exits != 0) -- but it may create output that was not intended exactly like that by the author (yay markdown!). This means that changes to the documentation need to be reviewed in text **as well** as rendered form. I don't know of a way to automate this in a way that would actually fail a build if the output doesn't look good.

There used to be a job which would render the markdown and fail the build if the output is not equal to what's checked in to the repository. We may consider re-instantiating that, as GH actions can actually be constrained to matching paths (so we don't have to run this if nothing changed below `docs`).

As mentioned in #531, it would also be great to generate an index page + the RFCs, and that could be decoupled from the spec output. The trouble is: I don't have a good idea how to _retain_ spec output in the main branch, but then have an additional deploy step which generates the site but reuses the spec output.

polyedre:
Ok I understand. I don't know if it is possible to setup something like Gitlab's review apps for this documentation. Maybe it's not  worth it.

I don't feel confident enough to contribute to the other points of the #531 issue. The issue may need to be more detailed.

Feel free to close this PR !

kim:
I played around with your work, and rerolled as #572

PR #568, opened 2021-03-16
type-driven replication development

FintanH:
Stacked on top of https://github.com/radicle-dev/radicle-link/pull/560 but is also up-to-date with `master` because I did some rebasing, hence the other commits.

Splitting the replication logic into two file modules:
* project
* person

The aim is to witness the computations in a type-driven fashion.  I'll try to give a breakdown of what's here so far for some early feedback.

The first thing was to split the two scenarios of:
a. the entity doesn't exist in our monorepo
b. the entity does exist

In `a.` we are relying on the provider we are fetching from to give us their representation of the identity. We `Peek` at their refs, determine the identity type, and `clone`. 
For the case of the project, using what we got from the provider we can calculate the set of delegate `PeerId`s we want to fetch from the provider and build up a view of all the delegates. We can proceed to do the dance of verifying, tracking, fetching signed refs, adopting the latest tip.
For the case of the `person`, it's a lot simpler because we don't need to dance with `rad/ids` when verifying.

In `b.` we can rely on our own representation of the identity. This means we can fetch everything based on our tracking graph and do the verification dance.

There are still some things I need to clean up:
- [x] There's no pruning done, just a `Pruned` type that isn't used
- [x] It should be the case that when we build up our delegate views that we can account for each delegate, i.e. for every anonymous peer we should have one remote entry, and for every indirect person, we should have _at least_ one remote from their set of keys.
- [x] Trace the tips, tracked peers, etc.
- [x] Probably some more wrangling of types

### DIFF
--- a/librad/src/git/identities/common.rs
+++ b/librad/src/git/identities/common.rs
@@ -22,6 +22,10 @@ impl<'a> From<&'a Urn> for IdRef<'a> {
 }
 
 impl<'a> IdRef<'a> {
+    pub fn oid(&self, storage: &Storage) -> Result<git2::Oid, git2::Error> {
+        Reference::rad_id(Namespace::from(self.0)).oid(storage.as_raw())
+    }
+
     pub fn create(
         &self,
         storage: &Storage,
--- a/librad/src/git/identities/error.rs
+++ b/librad/src/git/identities/error.rs
@@ -55,5 +55,11 @@ pub enum Error {
     Store(#[from] identities::git::error::Store),
 
     #[error(transparent)]
+    PersHist(#[from] identities::git::error::History<identities::git::PersonDoc>),
+
+    #[error(transparent)]
+    ProjHist(#[from] identities::git::error::History<identities::git::ProjectDoc>),
+
+    #[error(transparent)]
     Git(#[from] git2::Error),
 }
--- a/librad/src/git/identities/person.rs
+++ b/librad/src/git/identities/person.rs
@@ -5,7 +5,6 @@
 
 use std::{convert::TryFrom, fmt::Debug};
 
-use nonempty::NonEmpty;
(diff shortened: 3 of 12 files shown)


kim:
I think this is an improvement, strucuturally, but I also feel like it's not becoming _simpler_ to see what is going on. This is probably because the abstraction is moved more into the "identities domain", while the effects are still ... basically unconstrained.

In my head, we should only need to reach of the identity types when we a. do verification, and b. compute a subset of the refs which are expected to exist. What I'm looking for is an abstraction over the repository-level operations:

Conceptually, we have _repositories_ (ie namespaces ie URNs). Inside _repositories_, we have _trees_, each owned by a separate peer. To add to the confusion, there are several kinds of trees, but we are really only interested in the _ref tree_. As an example:

```
refs/namespaces/deepbee/refs/remotes/fa1afe1/*
```

is the `deepbee` _repository_'s _ref tree_ **owned** by `fa1afe1`. Below that _ref tree_, there are _subtrees_ which are distinguishable by their _root_ (aka "category", eg `heads`, `rad`, `rad/ids`), and partially by a specific shape (eg. we expect `rad` to have an `id` leaf and a `rad/ids` subtree, which in turn has n leaves of kind root hash).

What the fetch procedure is about is essentially:

* Fetch the smallest amount of subtrees necessary to assert validity
* Map the _ownership_ of ref trees
* Assert presence of certain subtrees
* Compute _additional_ ref trees which are either _required_ to be present, or _should_ be
* Verify a bunch of signatures

So I think the abstraction needs to be _parametrised over_ the identity type.

FintanH:
> So I think the abstraction needs to be parametrised over the identity type.

One clarification, are you referring to `librad::identities::Identity` here?

FintanH:
Closing since a different approach will be taken, afaiu.

Issue #208, opened 2020-06-24
PeerApi: Prove a method to "sync" with the currently-connected peers


kim:
This would require protocol support for efficiently computing the intersection
of URNs two peers have locally. As a first step, however, we can just implement
it as a loop which calls `Storage::fetch_repo` for each local URN and connected
peer, ignoring errors.

FintanH:
So interestingly enough @xla pretty much implemented this in proxy. This is likely one of the components that we should migrate back to link.

kim:
Yes would be good, so we can replace it with a more efficient implementation.

xla:
> Yes would be good, so we can replace it with a more efficient implementation.

:100:

PR #569, opened 2021-03-16
net: Mutual fetch (grafting pt. 1)

kim:
As the first step towards the grafting protocol (aka gossip-independent
bulk sync), a peer now may initiate a reverse replicate (rere) whenever
a git fetch comes in. Whether a rere is initiated depends on an
intersection of locally tracked and remotely advertised `PeerId`s.

To prevent this procedure from looping indefinitely, the initiator now
includes a random nonce in the fetch request header. Only if the nonce
is present and determined to be reasonably fresh, a rere is attempted
(which, in turn, does **not** set any nonce). This nonce may come in
handy in other ways, too.

Note that the nonce is generated by default, but it is possible to
explicitly avoid a rere by disabling it.


librad/src/net/protocol/io/graft.rs
```diff
@@ -0,0 +1,193 @@
...
+    .await?
+}
+
+fn is_interesting<P>(remote_peer: &P, remote_heads: &RemoteHeads, remotes: &Remotes<P>) -> bool
+where
+    P: Ord + FromStr,
+{
+    let tracked_local = remotes.flatten().collect::<BTreeSet<_>>();
+    //
+    // Ok, computer science NERDS. Can we make this actually efficient?
```
FintanH, line 116: Since we're just looking for _not being disjoint_, there is the opportunity to short-circuit with something like `any`. That is, we iterate through the `refs`, parse, and check if it's in `tracked_local`. That should be linear in `refs` and has the best-case of exiting as early as possible.

However, is this not an opportunity for us to _know_ which `PeerId`s two sides share and that means we can focus replication on those only?

librad/src/net/protocol/io/graft.rs
```diff
@@ -0,0 +1,193 @@
...
+/// set of [`PeerId`]s acc. to the local [`Remotes`] for the given [`Urn`], and
+/// the advertised remote tracking branches of the remote peer (plus the remote
+/// peer itself).
+///
+/// This function is invoked when the `remote_peer` initiates a fetch (hence
+/// "rere" -- replicate replicate), as per [`super::recv::git`].
+///
+/// A fetch initiated by this function will **not** generate a
+/// [`crate::git::p2p::header::Header::nonce`], so as to not trigger reres
+/// recursively. Using this function thus requires to inspect the git header for
```
FintanH, line 68: Can you expand on this "inspect the git header" part? Is it something the caller does? Do we have an example of it?

kim: I linked to [`super::recv::git`], is that not enough?

FintanH: Right, I missed that :+1:

librad/src/net/protocol/io/graft.rs
kim: > to short-circuit with something like any

Ah yeah... not sure why I thought there could be duplicates in the input.

> However, is this not an opportunity for us to know which PeerIds two sides share and that means we can focus replication on those only?

But we can't make use of this information -- `replicate` figures out which remotes it wants, so knowing that it will find at least one is enough.

FintanH: > But we can't make use of this information -- replicate figures out which remotes it wants, so knowing that it will find at least one is enough.

We can't at the moment, but I'm thinking we could allow `replicate` to be intialised with a set of peers that's interesting for the fetch. Maybe that's not an improvement though?

kim: Yes, we would save re-reading the tracked peers when computing the `Peek` fetchspecs. That's not nothing, but also not super-critical. I thought we might want to apply more algebra to the replication procedure first :nerd_face:

FintanH: Sounds good :smile: I better `solve for x` so

librad/src/net/protocol/io/graft.rs
kim: The question I wanted you to ask was: why is it that the nonce is set by default, and unset when `rere`ing, vs. unset by default and set by `rere`.

The answer is that I'm not sure either.

FintanH: Ok, now you've hooked me. I think I need a bit more prose on the `nonce`. I _thought_ I understood its purpose but now I've unconvinced myself.

kim: Maybe you can say what you thought?

What's documented here is that, since we trigger the `rere` by just a regular fetch and not some special-fetch, we need to be able to distinguish which one it is regardless, otherwise we'd just `rere` until eternity. It is possible that a fetch doesn't trigger a `rere` at all (because not interesting), or that the `rere` comes in after the original fetch has completed (very likely, actually).

So, we need to correlate `re` and `rere` pairs over some time window, right?

Things to keep in mind:

* The feature may get rolled out to a network which doesn't know `rere`, but is otherwise compatible
* If one side doesn't adhere to the protocol, can it do harm to the other side?
* Are there reasons to opt-out of the feature, and if so, which end should be able to decide?
* What other uses of a unique "request identifier" can you think of?

FintanH: > Maybe you can say what you thought?

I understood that it was marking when to mutually replicate but didn't quite piece it together, so I took another look and I'll try to explain what I think is happening.

The `Header` is the advertisement of what git action to do. We're embedding the `nonce` into the `Header` to advertise to the remote that we're in a mutual fetch mode. So `p1` will initiate a `replicate` from `p2` and will use a `nonce`. `p2` will see this `nonce` and decide if any of the remotes that `p1` has are interesting to it [`p2`]. If `p2` sees something interesting then it will also initiate a `replicate` but _will not_ send a `nonce` otherwise we would just keep going around in circles.

Admittedly, your explanation in the thread helped too.

> * The feature may get rolled out to a network which doesn't know rere, but is otherwise compatible
> * If one side doesn't adhere to the protocol, can it do harm to the other side?

These make sense, and from what I understand the `nonce` being optional satisfies these requirements. If the `nonce` is missing the other mutual fetch never happens but the `remote_peer` still at least gives the peer replicating the data it requested.

> * Are there reasons to opt-out of the feature, and if so, which end should be able to decide?

Again, a `nonce` could be ignored by intercepting the header and setting it `None`, right?

> * What other uses of a unique "request identifier" can you think of?

Heh, good question... None that spring to mind but that doesn't necessarily mean there won't be any. So maybe we shouldn't paint ourselves into a corner with the `nonce` thinking it'll be the only `nonce`. What I mean is that `n=` is now representing mutual fetches. But if we extend our cases and we need another `nonce` how do we choose the next parameter names and differentiate them?

One other question, [GitService::run](https://github.com/radicle-dev/radicle-link/pull/569/files#diff-310c08d56ff78cfbcec33b5d1c52979878d05c7f337eeb53b130ce606d655c26R101) ignores the `nonce`. Why is that? :thinking:

kim: > Again, a nonce could be ignored by intercepting the header and setting it None, right?

Uhm, you need to BYO fetcher now, which you can tell to not set the nonce. Since `replicate` is part of the public API, library users can say "fetch me this, but tell the other end to not dial back".

> Again, a nonce could be ignored by intercepting the header and setting it None, right?

We can stick more parameters at the end of the header line (`\0` separated). But since this is just a random value, I'm not sure what you have in mind by "another nonce".

A common pattern around nonces is that you expect the other end to "echo" it back to you, which I thought could come in handy.

> One other question, GitService::run ignores the nonce. Why is that?

Maybe you missed that I split the invocation of the "git server" in two steps now? You first call `service` which parses the header and returns you a type from which you can inspect it. Then, you actually run the thing -- which doesn't have any business with the `n` parameter, but needs to know the `service` and `repo` from the header.

FintanH: > not sure what you have in mind by "another nonce".

So if we had something like:
```rust
GitUrl {
   /// mutual fetch nonce
   nonce: Option<u32>,
   /// that new feature that we introduced that also required a nonce
   new_feature_nonce: Option<u32>,
}
```

> Maybe you missed that I split the invocation of the "git server" in two steps now? You first call service which parses the header and returns you a type from which you can inspect it. Then, you actually run the thing -- which doesn't have any business with the n parameter, but needs to know the service and repo from the header.

Ah yes, I see that now :+1: Something that could aid with connecting the concepts is some documentation on the `nonce` fields because their construction and their use are in different parts of the code base.

This has all been super helpful for understanding :blush:

kim: > /// that new feature that we introduced that also required a nonce

Yes sure, but why would we introduce another _random_ value? I mean, we could figure out that we want a monotonically increasing value instead, in which case it might be better to encode `n` now such that it doesn’t look like a number, so we could distinguish it. Or we could figure out that we want `proofofwork=256bits` instead, in which case we’d probably deprecate `n`, because that’s obviously superior entropy.

kim: > "fetch me this, but tell the other end to not dial back"

Yeah, so, Mr. Duck, you convinced me that we actually need to keep track of both sent and seen nonces (which I was trying to avoid). Otherwise, if you present a fresh nonce every time, you can keep the other side busy. If it is only `rere` which sends the nonce, it can keep the other side busy by just not sending a nonce. Instead, `rere` needs to echo the nonce it saw. Unnonced requests should get throttled. Thanks

kim: Or maybe I should not think about such things later in the day.

FintanH:
Just looking through the backlog, would this solve https://github.com/radicle-dev/radicle-link/issues/208, or is it only one part of the story?

kim:
It's pt. 1, as indicated in the subject

librad/src/net/protocol/io/graft.rs
FintanH: > Yes sure, but why would we introduce another random value?

I can't predict the future, unfortunately :smile: This was just a postulation of another case happening to pop up.

> we actually need to keep track of both sent and seen nonces

So this would say, "hey I've seen a nonce from this peer recently, and I answered so you can piss off I'm not gonna `rere`, come back tomorrow"?

PR #571, opened 2021-03-17
git: Fix Tracked iterator to ignore URN path

kim:
Some unlucky invocations would falsely write empty remotes to the sigrefs.


xla (approved):
🤐 🙎 🌇 5️⃣

FintanH (approved):

PR #572, opened 2021-03-18
CI: Add 'Hüb action to render spec

kim:
Updates to a more recent pandoc version and fixes the latex template to
be able to use `--citeproc`. The workflow job renders the spec and
asserts that it is identical to the checked-in output.

Co-authored-by: Lucas Henry <lucas_henry@live.fr>


kim:
For some reason, the pdf rendering is non-deterministic at 'Hüb. I can produce byte-for-byte identical pdfs locally. :thinking:

kim:
Ok, not ideal -- had to downgrade to a previous version of pandoc (or maybe it's the texlive bundled with the docker image?). Revisit when pandoc 2.12 lands in arch.

xla (approved):
💄 5️⃣ 📬 🚤

FintanH (approved):

PR #573, opened 2021-03-22
storage: check for tags in has_commit

FintanH:
The use of `find_commit` meant that we were ignoring tag objects. This
patch uses `find_object` instead to check if we have `Commit` or `Tag`
and then confirms if the SHA is in the given reference.

We extend the gossip test to see if the gossip of the tag gets propagated
back to `peer1`.

### DIFF
--- a/librad/src/git/storage.rs
+++ b/librad/src/git/storage.rs
@@ -136,7 +136,21 @@ impl Storage {
             .or_matches(is_not_found_err, || Ok(false))
     }
 
-    #[tracing::instrument(level = "debug", skip(self), err)]
+    /// Check the existence of `oid`.
+    ///
+    /// The `oid` may only point to an **annotated tag** or a **commit**.
+    ///
+    /// The result will be `false` if:
+    ///
+    /// 1. No object could be found for `oid`
+    /// 2. The object kind was not an **annotated tag** or **commit**
+    /// 3. The reference path for the `urn` could not be found (it defaults to
+    /// `rad/id` if not provided)
+    /// 4. The tip SHA was not:
+    ///    a. the same as the tag's SHA
+    ///    b. in the history of the commit
+    /// 5. The `oid` was the zero [`git2::Oid`]
+    #[tracing::instrument(level = "debug", skip(self, urn), fields(urn = %urn), err)]
     pub fn has_commit<Oid>(&self, urn: &Urn, oid: Oid) -> Result<bool, Error>
     where
         Oid: AsRef<git2::Oid> + Debug,
@@ -149,8 +163,8 @@ impl Storage {
         }
 
         self.backend
-            .find_commit(*oid)
-            .and_then(|commit| {
+            .find_object(*oid, None)
+            .and_then(|object| {
                 let namespace = Namespace::from(urn);
                 let branch = {
                     let path = match &urn.path {
@@ -161,13 +175,40 @@ impl Storage {
                 };
                 self.backend
                     .refname_to_id(&format!("refs/namespaces/{}/refs/{}", namespace, branch))
-                    .and_then(|tip| {
-                        Ok(tip == commit.id()
-                            || self.backend.graph_descendant_of(tip, commit.id())?)
+                    .and_then(|tip| self.has_tag_or_commit(tip, &object))
+                    .or_matches(is_not_found_err, || {
+                        tracing::debug!("ref not found");
+                        Ok(false)
                     })
-                    .or_matches(is_not_found_err, || Ok(false))
             })
-            .or_matches(is_not_found_err, || Ok(false))
+            .or_matches(is_not_found_err, || {
+                tracing::debug!("object not found");
+                Ok(false)
+            })
+    }
+
+    fn has_tag_or_commit(
+        &self,
+        tip: git2::Oid,
+        object: &git2::Object,
+    ) -> Result<bool, git2::Error> {
+        let as_commit = |obj: &git2::Object| {
+            let commit = obj.as_commit()?;
+            Some(if tip == commit.id() {
+                Ok(true)
+            } else {
+                self.backend.graph_descendant_of(tip, commit.id())
+            })
+        };
+
+        let as_tag = |obj: &git2::Object| {
+            let tag = obj.as_tag()?;
+            Some(Ok(tip == tag.id()))
+        };
+
+        as_commit(object)
+            .or_else(|| as_tag(object))
+            .unwrap_or(Ok(false))
     }
 
     #[tracing::instrument(level = "debug", skip(self), err)]
--- a/librad/src/git/storage.rs
+++ b/librad/src/git/storage.rs
@@ -150,18 +150,20 @@ impl Storage {
     where
         Oid: AsRef<git2::Oid> + Debug,
     {
-        let oid = match self.find_object(oid)? {
+        let (oid, kind) = match self.find_object(oid)? {
             None => return Ok(false),
             Some(object) => match object.kind() {
-                Some(git2::ObjectType::Commit) => object.id(),
+                Some(git2::ObjectType::Commit) => (object.id(), git2::ObjectType::Commit),
                 _ => return Ok(false),
             },
         };
 
-        let tip = self.tip(urn)?;
+        let tip = self.tip(urn, kind)?;
         Ok(tip
             .map(|tip| {
-                Ok::<_, git2::Error>(*tip == oid || self.backend.graph_descendant_of(*tip, oid)?)
+                Ok::<_, git2::Error>(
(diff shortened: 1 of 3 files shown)


FintanH:
One thing I punted on was the naming. I'm not sure if `has_commit` still fits :shrug:

xla (approved):
🐦 ⬛️ 👦 🚦

geigerzaehler (approved):

geigerzaehler (approved):

xla (dismissed):
🍦 ⛅️ ⚡️ 🎃

kim (changes requested):
This does the wrong thing, semantically.

The reason we look for the requested commit to be an ancestor of the branch name supplied is that we have a monorepo -- if we just looked for the commit sha, it could be in a completely different namespace.

Tags are a different beast altogether:

a. We MUST not allow tags to be rewritten
b. We MUST discard "lightweight" tags

Thus, if the intention is to make tags be propagated, we need to check that a. the sha peels to a tag object, and b. the tag object's sha matches exactly what was given.

FintanH:
>  b. We MUST discard "lightweight" tags

I can't spot anything in `git2` (nor remember) to tell if a tag is lightweight or not. Is the `git2::Tag` struct always the annotated version, do you know?

kim:
> Is the git2::Tag struct always the annotated version, do you know?

Yes it is. I think the most DWIM-safe way might be to `find_object` and then match the `ObjectType`.

librad/src/git/storage.rs
```diff
@@ -161,13 +175,40 @@ impl Storage {
...
-                    .or_matches(is_not_found_err, || Ok(false))
             })
-            .or_matches(is_not_found_err, || Ok(false))
+            .or_matches(is_not_found_err, || {
+                tracing::debug!("object not found");
+                Ok(false)
+            })
+    }
+
+    fn has_tag_or_commit(
```
FintanH, line 190: @kim I didn't exactly use the `ObjectType` but instead decided to use `as_object`. I believe this covers the requirements you pointed out

kim:
Ok. I would still prefer to not go down the DWIM route, and have two separate, well-defined methods `has_commit` and `has_tag` instead. `has (commit V tag)` is relevant in exactly one place, which is `impl broadcast::LocalStorage`. If you want to precompose there, that'd be fine.

xla (dismissed):
🔙 🍝 👃 😭

librad/src/git/storage.rs
```diff
@@ -136,51 +136,106 @@ impl Storage {
...
     where
         Oid: AsRef<git2::Oid> + Debug,
     {
-        let oid = oid.as_ref();
-        if oid.is_zero() {
-            // XXX: should this be a panic or error?
-            tracing::warn!("zero oid");
-            return Ok(false);
-        }
+        let oid = match self.has_object(oid)? {
```
kim, line 153: This accesses the odb twice

librad/src/git/storage.rs
```diff
@@ -136,51 +136,106 @@ impl Storage {
...
+                };
+                Ok(tip == tag.id())
+            })
+            .or_matches(is_not_found_err, || {
+                tracing::debug!("tag not found");
+                Ok(false)
+            })
+    }
+
+    fn get_urn_oid(&self, urn: &Urn) -> Result<Option<git2::Oid>, git2::Error> {
```
kim, line 207: Not sure... if you want to remove redundant code, I think `tip` does the same thing.

librad/src/git/storage.rs
```diff
@@ -136,51 +136,106 @@ impl Storage {
...
+        };
+        self.backend
+            .refname_to_id(&format!("refs/namespaces/{}/refs/{}", namespace, branch))
+            .map(Some)
+            .or_matches(is_not_found_err, || Ok(None))
     }
 
     #[tracing::instrument(level = "debug", skip(self), err)]
-    pub fn has_object<Oid>(&self, oid: Oid) -> Result<bool, Error>
+    pub fn has_object<Oid>(&self, oid: Oid) -> Result<Option<git2::Oid>, Error>
```
kim, line 223: `has` or `is` should return `bool`, no?

librad/src/git/storage.rs
```diff
@@ -136,51 +136,106 @@ impl Storage {
...
+                    Some(tip) => tip,
+                    None => return Ok(false),
+                };
+                Ok(tip == tag.id())
+            })
+            .or_matches(is_not_found_err, || {
+                tracing::debug!("tag not found");
+                Ok(false)
+            })
+    }
```
kim, line 205: You're accessing the backend three times here -- why? I think you want to just find the tip of the ref which corresponds to the URN's path, check that the object is of the right type, and then compare the Oids.

Please also note that the tag name (as recorded in the tag object) is not the same as the ref name.

FintanH: > Please also note that the tag name (as recorded in the tag object) is not the same as the ref name.

I'm not following what you're alluding to here :thinking: The URN should resolve to a tip and we get the tag via that, so where is the tag name involved?

kim: Something for @geigerzaehler to consider: normally, tags are in the global namespace of a repository. This means that the tag name (the `tag` header in the object) is equal to the ref (after `tags/`). For example, the tag `v1.0` is usually at `refs/tags/v1.0`. If the tag is GPG-signed, that signature implies that this is the `v1.0` tag, and it shall be at `refs/tags/v1.0`, and there shall be no other `v1.0` tags beside me.

Since we are scoping tags to the peer-id, this is no longer true -- there can be more than one tag object with `tag v1.0` in the repository. `git verify-tag` does _not_ verify that the tag and ref names match, if you care about "tag teleport" attacks, you'd have to incantate `git verify-tag --verbose`, and look at the output. This is obviously not becoming easier to understand when there are N `v1.0` tags in the repo.

**tl;dr** it would be good compare the ref name (acc. to our conventions) and the tag name higher up in the stack, before showing a "merge" button for example.

FintanH: Ach, I see :) @geigerzaehler might be a good idea to add this to any docs you have on this feature so we don't lose track of the info

geigerzaehler: Thanks for the note. I’ll make sure that we check that the `tag` field matches the ref when resolving the merge requests.

librad/src/git/storage.rs
```diff
@@ -136,8 +136,65 @@ impl Storage {
...
+    {
+        let oid = match self.find_object(oid)? {
+            None => return Ok(false),
+            Some(object) => match object.kind() {
+                Some(git2::ObjectType::Tag) => object.id(),
+                _ => return Ok(false),
+            },
+        };
+
+        let tip = self.tip(urn)?;
```
kim, line 192: Uh-oh wait -- `tip` peels to a commit, which is _probably_ not the `oid` passed here. Just using `refname_to_id` here is cheaper and what you want.

I'm not sure if we should remove the implicit type conversion of `tip`. I think this would break the assumptions of the identities stuff, so perhaps it should return `Object` instead. This puts the burden of deciding what to do if the tip is _not_ a commit on the caller, which may require some thought.

FintanH: Ach nooooo. I completely missed that. Should have got better laser eye surgery.

> I think this would break the assumptions of the identities stuff, so perhaps it should return Object instead. This puts the burden of deciding what to do if the tip is not a commit on the caller, which may require some thought.

I tried grepping for its use and it seems to only be used here. The identity stuff seems to use `peel_to_commit` directly. So the only affected spots would be `has_commit` and `has_tag` here.

librad/src/git/storage.rs
```diff
@@ -193,7 +236,14 @@ impl Storage {
 
         match reference {
             None => Ok(None),
-            Some(r) => Ok(Some(r.peel_to_commit()?.id().into())),
+            Some(r) => r
+                .peel(kind)
+                .map(Some)
+                .or_matches::<Error, _, _>(is_invalid_spec_err, || {
```
kim, line 242: Sorry for the persistent nitpicking, but I think this is perhaps not the best idea. We use `Result<Option<T>, E>` where `Ok(None)` means "doesn't exist". A type mismatch is clearly something I'd expect to be an error. Let's just say `tip` gives me the `Object` the ref points to, or `None` if the reference doesn't exist.

FintanH: Ya fair, my thinking was that `None` meant that the `Object` doesn't exist for the type supplied. But you're right, the user messed up so the error should be the loud and proud.

kim (dismissed):

xla (dismissed):
utACK

😪 😛 💧 🗃

PR #576, opened 2021-03-30
daemon: migrating the radicle-upstream/coco code

FintanH:
We're moving the code that was living in radicle-upstream/coco to start
as the foundation of a daemon.

The one caveat is that coco/source.rs will not be present as its
migration path is towards radicle-surf.

Everything else is a verbatim move of the coco files where the only
changes were for imports. These were mostly using `librad::git_ext` as
opposed to `radicle_git_ext` and the removal of `radicle_surf::git2`.


daemon/Cargo.toml
```diff
@@ -0,0 +1,52 @@
+[package]
+name = "daemon"
+version = "0.1.0"
+license = "GPL-3.0-only"
+authors = ["radicle team <dev@radicle.xyz>"]
+edition = "2018"
+description = "Abstractions and utilities to run and interact with radicle-link"
+readme = "https://github.com/radicle-dev/radicle-upstream/blob/master/proxy/README.md"
```
kim, line 8: :monocle_face:

kim (commented):
> These were mostly using `librad::git_ext` as opposed to `radicle_git_ext`

Did this stop working? Why?

daemon/Cargo.toml
xla: morelike

FintanH:
> Did this stop working? Why?

I wanted to be consistent by using `librad::git_ext` since it re-exported it. However, now that I came to realise that I need to add `radicle-git-ext` anyway because macro expansion I'm thinking of swinging the other way.

daemon/src/config.rs
```diff
@@ -0,0 +1,116 @@
...
+        seeds
+            .iter()
+            .map(|seed| (seed.peer_id, seed.addrs.as_slice())),
+    )
+    .expect("unable to resolve seeds")
+}
+
+/// Stream based discovery based on a watch.
+#[derive(Clone)]
+pub struct StreamDiscovery {
```
xla, line 86: Do we want to promote this into librad::discovery?

daemon/src/control.rs
```diff
@@ -0,0 +1,316 @@
+//! Utility for fixture data in the monorepo.
```
xla, line 1: Not sure this module should make it over as the fixtures and git-platinum handling seems to be upstream specific, unless I'm missing something.

daemon/src/git_helper.rs
```diff
@@ -0,0 +1,69 @@
+//! git-remote-rad git helper related functionality.
```
xla, line 1: Also doesn't seem to belong in the context of the daemon, either general utility pulled in from this repo or back in Upstream.

daemon/src/lib.rs
```diff
@@ -0,0 +1,53 @@
...
+pub use librad::git_ext;
+
+pub use radicle_git_helpers::remote_helper;
+
+pub mod config;
+pub mod control;
+pub mod convert;
+pub mod git_helper;
+mod identifier;
+pub use identifier::Identifier;
```
xla, line 45: Was this moved to upstream::api?

daemon/src/peer/run_state/config.rs
```diff
@@ -0,0 +1,90 @@
+//! Configuration types for configuring the `RunState`.
+
+use std::time::Duration;
+
+/// Default time to wait between announcement subroutine runs.
+pub(super) const DEFAULT_ANNOUNCE_INTERVAL: Duration = std::time::Duration::from_secs(1);
```
xla, line 6: Are these used outside of this file?

daemon/src/project.rs
```diff
@@ -0,0 +1,69 @@
...
+/// Set the upstream of the given branch to the given remote.
+///
+/// This writes to the `config` directly. The entry will look like the following:
+///
+/// ```text
+/// [branch "main"]
+///     remote = rad
+///     merge = refs/heads/main
+/// ```
+fn set_upstream<Url>(
```
xla, line 31: Seems to be similar to other utilities better placed somewhere else.

xla (changes requested):
Big move! Already commented inline a bunch, but it would be nice to separate already utility concerned with working copies, etc. and keep the `daemon` crate focused on a running peer managing the subroutines.

daemon/src/lib.rs
FintanH: This is what prompted me. I noticed it as I was going through the files.

FintanH:
Making this a draft because there's some bits to get right first :) Thanks for the feedback, keep it coming :blush:

FintanH:
@xla I'd like to keep moving things around to a minimum, so while I think you're right on some of the points I think it'd be good to capture those as follow up tasks and only consider removal of files for now. How do you feel about that?

FintanH:
> I get the intent, just feels like we do the work twice as there obvious targets to shuffle. But in the end

Understandable, but my reasoning here is that by breaking into these steps we should ensure that we maintain parity with the current state of Upstream. Once that's ensured we can't start fiddling bits. Slow progress yes, but I think it's easier to progress that way.

kim:
> I need to add radicle-git-ext anyway because macro expansion

That's the part I don't understand

daemon/src/config.rs
xla: Is this captured for a follow-up?

daemon/src/identifier.rs
```diff
@@ -0,0 +1,113 @@
+// Copyright © 2019-2020 The Radicle Foundation <hello@radicle.foundation>
```
FintanH, line 1: Removing this

xla: Was this left behind to live in API?

daemon/src/keystore.rs
```diff
@@ -0,0 +1,171 @@
+// Copyright © 2019-2020 The Radicle Foundation <hello@radicle.foundation>
```
xla, line 1: Would we wanna peel this away not maintain keystore util in this crate? Not for this change but maybe something to capture.

daemon/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+    clippy::similar_names,
+    clippy::too_many_lines
+)]
+#![feature(duration_zero, hash_set_entry, never_type)]
+
+#[cfg(test)]
+#[macro_use]
+extern crate assert_matches;
+
+pub use librad::{
```
xla, line 36: Is this still necessary given that downstreams have to depend on the repo anyway?

daemon/src/peer/run_state/config.rs
xla: bump

xla (commented):
Looks good to me, just a bunch of points about capturing follow-ups

daemon/src/identifier.rs
FintanH: Just removed it :)

FintanH: https://github.com/radicle-dev/radicle-link/pull/576/commits/11f2aa8a81508879e1b20390d91e3a8281a8ae18

daemon/src/lib.rs
FintanH: They don't need to depend on `librad`, they have to depend on `radicle-git-ext`. I think it would make sense just to re-export `librad` though and not specify the other re-exports. Whatya think?

daemon/src/peer/run_state/config.rs
FintanH: I don't think so. I'll privatise the sector

daemon/src/lib.rs
xla: Originally the reexport of `librad` was suppose to hide away the link traces from the api crate which dependent on coco and coco on librad. As we now know that was a bad idea to begin with and downstream has to get a repo dependency anyway, if it need librad types it should depend on it directly imho. It also appears odd that here a re-export of another crate from the same workspace takes place.

FintanH: > As we now know that was a bad idea to begin with and downstream has to get a repo dependency anyway

But it doesn't need a `librad` repo dependency is what I'm saying, the sum total of deps it needs to declare could be `radicle-daemon` and `radicle-git-ext`. The latter being an unfortunate circumstance of `radicle-macros` functions expanding to `::radicle_git_ext`.

> It also appears odd that here a re-export of another crate from the same workspace takes place.

This also happens in `librad` with `macros`, `git-ext`, and `std-ext`.

So the issue with having multiple upstream declarations is you end up in this situation:
```
[dependency.foo]
git = "<dev/foo>"
rev = "abc"

[dependency.bar]
git = "<dev/bar>"
rev = "abc"
```

and now every time they update `"abc"` they have to do it everywhere. If they don't then they get weird-ass errors where the types are the same but come from different versions of the project -- we saw this with `radicle-git-ext` when fixing up some upstream code.

Re-exporting `librad` from `radicle-daemon` will mean that the types used are always on the same versioning, which is very useful IMO.

The alternatives to re-exporting are:
1. Require downstream users to add deps with the SHAs that they have to manage manually
2. Require downstream users to point to a particular branch, i.e. `master`, for all related deps and run `cargo update` to get the latest changes 

So ya, we could let them add deps in these manners :) I suppose we just have to weigh up the options (or maybe there's options I'm missing?). @geigerzaehler might be interested in this as well :eyes:

geigerzaehler: > Re-exporting `librad` from `radicle-daemon` will mean that the types used are always on the same versioning, which is very useful IMO.

I agree and would prefer to have `librad` re-exported from `radicle-deamon`.

> I think it would make sense just to re-export `librad` though and not specify the other re-exports. Whatya think?

That would be a good idea. The API surface is already quite large and complicated so anything we can do to streamline this would be appreciated.

kim: Please keep in mind that a daemon is a process, not a library. As such, the surface API of this crate should be equivalent to its HTTP API.

Some kind of "prelude" crate which exports things to build your own daemon (or whatever application) is a separate concern. I am a strong proponent of designing libraries in a way such that even "internal" / unstable calls are accessible to consumers, which is not something Rust crates are very good at, and one of the reasons I'm in favour of a crate-zoo which can be recomposed into whatever preferences users might have.

FintanH: Thanks for the input @geigerzaehler and @kim :bow: 

> Please keep in mind that a daemon is a process, not a library. As such, the surface API of this crate should be equivalent to its HTTP API.

So because I wanted to stage these moves there is no HTTP API in here yet. This means that we need to have some parity for re-exporting until we move that portion of the code over. 

With that in mind, I would say we keep this as is to keep parity -- which is currently green https://github.com/radicle-dev/radicle-upstream/pull/1753. Then when we move on to the next stage of moving things over we can remove the re-exports here and the HTTP API can rely on `path = ../librad`. 

Does that sound reasonable to yis?

daemon/src/config.rs
FintanH: https://github.com/radicle-dev/radicle-link/issues/588

daemon/src/project.rs
FintanH: Should be addressed in https://github.com/radicle-dev/radicle-link/issues/587

alexjg (dismissed):

xla (approved):
🏐 🙋 🎖 🐵

daemon/src/tests/gossip.rs
```diff
@@ -0,0 +1,221 @@
+// Copyright © 2019-2020 The Radicle Foundation <hello@radicle.foundation>
```
geigerzaehler, line 1: This file is currently located at `daemon/src/tests/gossip.rs`. I believe it should be located at `daemon/tests/gossip.rs`. Otherwise these tests are never run.

FintanH: Ahahaha thanks for the catch :smile:

Issue #575, opened 2021-03-24
Timeouts don't seem to cause retries


viraptor:
I've connected between a seed running net-next (ce7dc5e69c89845124edc7b01b6c9e6e92c76273) and matching upstream (82f0960e539d72631e72e17f357142cd461992d9). I got some projects successfully trying to replicate: (seed logs)
```
Mar 24 00:35:24.976  INFO incoming: librad::net::protocol::io::streams: new ingress bidi stream remote_id=hynmraey7hm8q4g7ipnnejc6oah9fccfnppa49xntoyxo6wgr6y1uh remote_addr=220.235....
Mar 24 00:35:24.976  INFO invoke_service: librad::git::p2p::server: upload pack ls
Mar 24 00:35:24.980  INFO invoke_service: librad::git::p2p::server: upload pack ls done
Mar 24 00:35:25.035  INFO incoming: librad::net::protocol::io::streams: new ingress bidi stream remote_id=hynmraey7hm8q4g7ipnnejc6oah9fccfnppa49xntoyxo6wgr6y1uh remote_addr=220.235....
Mar 24 00:35:25.036  INFO invoke_service: librad::git::p2p::server: upload pack ls
Mar 24 00:35:25.039  INFO invoke_service: librad::git::p2p::server: upload pack ls done
```
However just after that I got:
```
Mar 24 00:36:44.929  WARN incoming: librad::net::protocol::io::streams: ingress uni error err=Connection(ApplicationClosed(ApplicationClose { error_code: 8, reason: b"timeout" }))
```
Which is ok, network hiccups happen.
However it looks like this doesn't schedule any retries. My upstream node was disconnected permanently, doesn't exist on the seed's peer list anymore. The upstream client still thinks it's connected (based on the node count), but the seed id does not appear anywhere in the logs.

xla:
What's the output of the `/membership` endpoint on the seed frontend when you run into this?

viraptor:
Ok, some more info - we can ignore the git parts from the original report. something more than just timeouts seems to be happening here.
Using seed from the current master (5554f44):

One side (connecting):
```
Mar 31 00:10:21.806 DEBUG drive{id=0}: quinn_proto::connection: ECN not acknowledged by peer
[...]

PR #577, opened 2021-03-30
net: advance membership on connection errors

kim:
When switching from long-lived bidi- to ad-hoc uni-streams for gossip,
it was not properly considered that the partial view will not repair
itself in mostly-idle networks -- when there's nothing to send, and
nothing received, we need a cue from the connection driver and dispatch
that to the membership state machine. Note that this relies on PTO
expiration respectively delivery of CONNECTION_CLOSE frames from the
remote, and thus doesn't address any weird states due to "unclean"
reconnects within the PTO period.

Related-to: #575

### DIFF
--- a/librad/src/net/protocol.rs
+++ b/librad/src/net/protocol.rs
@@ -16,7 +16,7 @@ use std::{
 use futures::{
     channel::mpsc,
     future::{BoxFuture, FutureExt as _, TryFutureExt as _},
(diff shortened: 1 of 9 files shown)


xla (approved):
🐊 🚨 ✖️ 🈸

FintanH (approved):
Looks good :+1:

viraptor:
Tested and it seems to solve the original issue.

PR #580, opened 2021-04-01
net: Interrogation API (grafting pt. 2)

kim:
In the second installment towards the grafting protocol, we introduce an
API and associated stream type to interrogate a known remote peer about
certain data. Of relevance for grafting is strictly speaking only the
RPC to ask for the complete list of URNs.

We use the Xor filter probabilistic datastructure for that, in order to
gain compact and bounded responses (see commentary inline). The work to
compute the set intersection will be on the requester, while the
responder recomputes the filter only when necessary (morally; we can't
currently intercept refs updates in the relevant places, so recompute
after an expiry timeout).

### DIFF
--- a/librad/src/net/protocol/interrogation/xor.rs
+++ b/librad/src/net/protocol/interrogation/xor.rs
@@ -1,0 +1,233 @@
+// Copyright © 2019-2020 The Radicle Foundation <hello@radicle.foundation>
+//
+// This file is part of radicle-link, distributed under the GPLv3 with Radicle
+// Linking Exception. For full terms see the included LICENSE file.
+
+use sized_vec::Vec as SVec;
+use thiserror::Error;
+use typenum::{IsLessOrEqual, Unsigned, U10000};
+use xorf::{Filter as _, Xor16};
+
+use crate::identities::{SomeUrn, Urn};
+
+/// Maximum number of elements permitted in a single [`Xor`] filter.
+///
+/// Currently: 10,000
+pub type MaxElements = U10000;
+// approx. `MaxElements * 1.23`, but not exactly for all choices of
+// `MaxElements`
+const MAX_FINGERPRINTS: u64 = 12_330;
+
+#[derive(Debug, Error)]
+#[non_exhaustive]
+pub enum BuildError<E: std::error::Error + Send + Sync + 'static> {
+    #[error("too many elements")]
+    TooManyElements,
+
+    #[error(transparent)]
+    Inner(#[from] E),
+}
+
+/// Compact representation of a potentially large number of local [`Urn`]s, with
+/// approximate membership tests.
+///
+/// We use Lemire et.al.'s [Xor filter][xor] with 16-bit fingerprints, which
+/// gives a false positive rate of < 0.02. The number of elements in the filter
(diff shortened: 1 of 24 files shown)


librad/src/net/protocol/interrogation/xor.rs
```diff
@@ -0,0 +1,233 @@
...
+                } else {
+                    Ok(Decode::decode(d)?)
+                }
+            },
+        }
+    }
+}
+
+fn xor_hash(urn: &SomeUrn) -> u64 {
+    let SomeUrn::Git(Urn { id, path: _ }) = urn;
```
FintanH, line 177: I'd prefer we got a compiler error from a missing case in `match` if we added a new variant :)

librad/src/net/protocol/interrogation/xor.rs
```diff
@@ -0,0 +1,233 @@
...
+            let digest = self.hasher.finalize_reset();
+            let oid = git2::Oid::from_bytes(&digest).unwrap();
+            SomeUrn::Git(Urn::new(oid.into()))
+        }
+    }
+
+    #[test]
+    fn false_negatives() {
+        let mut bob = BuildUrn::new();
+        let urns = SVec::<MaxElements, _>::fill(|i| bob.build(&i.to_be_bytes()));
```
FintanH, line 211: "Can he build it?! Yes he can!"

librad/src/net/protocol/io/send/request_response.rs
```diff
@@ -0,0 +1,50 @@
...
+    protocol::{error, interrogation, quic, upgrade},
+};
+
+pub trait Pairing {
+    type Response;
+    type Upgrade: Into<upgrade::UpgradeRequest>;
+    const UPGRADE: Self::Upgrade;
+}
+
+impl Pairing for interrogation::Request {
```
FintanH, line 23: Are there other `Pairing`s? I was expecting to possibly see `Git`, `Gossip`, and `Membership` ones based on the different upgrades we've defined

FintanH (commented):
Slick :ok_hand:

librad/src/net/protocol/interrogation/xor.rs
kim: But you will - Rust won’t accept refutable patterns in a let binding.

librad/src/net/protocol/io/send/request_response.rs
kim: Yeah but they’re not allowed in a request/response pattern. I was briefly tempted to generalise the `send_rpc` function in a similar way, but then decided to leave it to a later patch, because it would add quite some diff noise to this one.

librad/src/net/protocol/interrogation/xor.rs
FintanH: Right, I mixed this up with `if let` :facepalm:

FintanH (approved):

librad/src/net/protocol/interrogation/xor.rs
kim: No no, admit it: you were playing with this Huskall thing again!

PR #582, opened 2021-04-02
identities: add remove_ext to Payload

NunoAlexandre:
Motivated by the need to remove an Ethereum claim ext in the Upstream,
which is done by removing claim.

### DIFF
--- a/librad/src/identities/payload.rs
+++ b/librad/src/identities/payload.rs
@@ -264,6 +264,16 @@ where
     {
         self.ext.range(range)
     }
+
+    pub fn remove_ext<U>(&mut self) -> Result<Option<U>, serde_json::Error>
+    where
+        U: HasNamespace + serde::de::DeserializeOwned,
+    {
+        self.ext
+            .remove(U::namespace())
+            .map(serde_json::from_value)
+            .transpose()
+    }
 }
 
 impl<T> serde::Serialize for Payload<T>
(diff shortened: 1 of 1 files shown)


librad/src/identities/payload.rs
```diff
@@ -725,6 +735,28 @@ mod tests {
...
+
+        assert_eq!(
+            payload.get_ext::<UpstreamUser>().unwrap().unwrap(),
+            UpstreamUser {
+                registered_as: "cloudhead".into(),
+            }
+        );
+
+        let result = payload.remove_ext::<UpstreamUser>();
+        assert!(result.is_ok());
```
FintanH, line 756:
```suggestion
        assert_eq!(result, Ok(UpstreamUser { registered_as: "cloudhead".into() }));
```

small suggestion to make sure get out what we put in. If it's a case of `Debug` impls missing we can also use `assert_matches!`

FintanH (dismissed):
Makes sense to me :+1:

librad/src/identities/payload.rs
NunoAlexandre: Good one! I changed it to this, since it returns `Result<Option<U>, _>`

``` rust
        assert_eq!(
            result.unwrap(),
            Some(UpstreamUser {
                registered_as: "cloudhead".into()
            })
        );
```

FintanH: I have a feeling you might have force pushed the wrong commit :thinking: I don't see the code as updated :sweat_smile:

NunoAlexandre: Oops! Here it is: https://github.com/radicle-dev/radicle-link/pull/582/files#diff-db230ec3b5c04fb040752d237b3eaa60cd2725238e5a0c4d679055dbf4db93b5R755-R761

xla (approved):
🌇 🎐 👦 ⬛️

kim (approved):
I think it could’ve been solved equivalently by `set_ext(None)`, but I’m also ok with this.

NunoAlexandre:
> I think it could’ve been solved equivalently by `set_ext(None)`, but I’m also ok with this.

Good one. I will leave it as is since 1. I find "remove" more idiomatic than "set to None" and 2. because if "remove" was implemented by calling `set_ext(None)`, we wouldn't have the removed ext being returned, which _could_ be useful eventually.

PR #585, opened 2021-04-06
fetch: limit testing and increase

FintanH:
Adding a smoke test to show that a peer with 4 remotes will fail to replicate from another peer due to `PeekAll` exceeding the fetch limit. This pushes the limit up to 10kB but definitely isn't the fix. 

https://github.com/radicle-dev/radicle-link/pull/568 can partly address this since it doesn't use `PeekAll` anymore and works out the remotes it needs to peek at based on the provider, so it can scale the `Peek` in that case.

This increase will help with the replication issues that were noticed when testing Upstream.

### DIFF
--- a/librad/src/git/fetch.rs
+++ b/librad/src/git/fetch.rs
@@ -41,7 +41,7 @@ use crate::{
 pub const ONE_KB: usize = 1024;
 /// 5KiB for use in [`Limit`], specifically for the `peek` field, when we would
 /// like to fetch `rad/id` , `rad/self`, `rad/ids/*` references.
-pub const FIVE_KB: usize = ONE_KB * 5;
+pub const TEN_KB: usize = ONE_KB * 10;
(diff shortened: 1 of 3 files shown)


librad/src/git/fetch.rs
```diff
@@ -41,15 +41,15 @@ use crate::{
 pub const ONE_KB: usize = 1024;
 /// 5KiB for use in [`Limit`], specifically for the `peek` field, when we would
 /// like to fetch `rad/id` , `rad/self`, `rad/ids/*` references.
-pub const FIVE_KB: usize = ONE_KB * 5;
+pub const TEN_KB: usize = ONE_KB * 10;
```
xla, line 44: Very honest.

xla (approved):
> Nobody has more than 3 friends anyway.

🐹 🌊 📤 👢

PR #591, opened 2021-04-08
WIP: Emit events on state changes in the waiting room

alexjg:
This PR modifies the `WaitingRoom` to emit `Transition` objects whenever it's state changes. These events are then routed to the `Subroutine::subscriber` so that they can be consumed by out of process listeners who are interested in network state.


daemon/src/request/waiting_room.rs
```diff
@@ -189,30 +234,50 @@ impl<T, D> WaitingRoom<T, D> {
...
+                        })
+                    },
+                    Either::Left(mismatch) => Err(Error::StateMismatch((&mismatch).into())),
+                }
             },
         }
     }
 
+    /// Create a transition where the before and after state are the same
+    pub fn tick(&self, timestamp: T) -> Transition<T>
```
FintanH, line 269: Hmmmm I'm curious about this, as per one of my previous comments in `event.rs` :thinking: Why do you think it's needed?

daemon/src/request/waiting_room.rs
```diff
@@ -623,21 +770,26 @@ mod test {
 
         for remote_peer in &peers[0..NUM_CLONES] {
             waiting_room.cloning(&urn, *remote_peer, ())?;
-            waiting_room.cloning_failed(&urn, *remote_peer, ())?;
+            waiting_room.cloning_failed(&urn, *remote_peer, (), "no reason".to_string())?;
```
FintanH, line 773: lol :smile:

daemon/src/peer/run_state.rs
```diff
@@ -26,12 +26,17 @@ use librad::{
...
     peer::PeerId,
 };
+use waiting_room::TransitionWithResult;
 
 use crate::{
     convert::MaybeFrom,
     peer::{announcement, control},
-    request::waiting_room::{self, WaitingRoom},
+    request::{
+        waiting_room::{self, Transition as WaitingRoomTransition, WaitingRoom},
```
FintanH, line 35: I'm more fond of using crate prefixes instead of renaming, so `waiting_room::Transition` rather than `WaitingRoomTransition`. I find it more clear when tracking through types.

What do you think?

daemon/src/peer/run_state.rs
```diff
@@ -26,12 +26,17 @@ use librad::{
     },
     peer::PeerId,
 };
+use waiting_room::TransitionWithResult;
```
FintanH, line 29: nit: this can probably be moved down to the import list below :)

daemon/src/peer/run_state.rs
```diff
@@ -318,10 +361,14 @@ impl RunState {
                         provider: PeerInfo { peer_id, .. },
                         result,
                     } => {
-                        if let Err(waiting_room::Error::TimeOut { .. }) =
-                            self.waiting_room.found(&urn, peer_id, SystemTime::now())
-                        {
-                            cmds.push(Command::Request(command::Request::TimedOut(urn.clone())));
+                        let transition = self.waiting_room.found(&urn, peer_id, SystemTime::now());
+                        if let Ok(transition) = transition.map(TransitionWithResult::transition) {
```
FintanH, line 365: And if there was an error? Do we not want to log that?

FintanH (commented):
Looking good so far :) Have a few points I'd like to flesh out with you

daemon/src/request/event.rs
```diff
@@ -0,0 +1,74 @@
...
+    },
+    /// A request was timed out
+    TimedOut {
+        /// The urn that timed out
+        urn: Urn,
+        /// The attempts that were made before the timeout
+        attempts: Attempts,
+    },
+    /// The waiting room was ticked - this is effectively a NOOP
+    Tick,
```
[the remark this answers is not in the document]
alexjg: Yeah it does feel a little out of place. The reason it's there is effectively to make it easy to emit the waiting room state using the same infrastructure as other waiting room events. This is useful for the frontend because waiting room transitions are relatively infrequent so without a tick event we would need to add something like a new proxy endpoint for getting the current waiting room state. It also has the nice property that if something has silently changed the waiting room state somehow (which can happen due to a coding mistake as the events are bolted on after the fact, rather than being the source of truth) then  we will see it in the event stream on the next tick - and hopefully that should alert us that something fishy is going on.

One option would be to rename it to `Noop`. Another would be to parameterise all the mutating methods of the waiting room on an event type and value. That way assigning events becomes the responsibility of the calling code.

daemon/src/request/waiting_room.rs
```diff
@@ -128,6 +104,48 @@ where
...
+#[derive(Debug, PartialEq)]
+pub struct TransitionWithResult<T, R> {
+    /// The event that caused the state change
+    event: Event,
+    /// The time  the change occurred
+    timestamp: T,
+    /// The state before the change
+    state_before: HashMap<Revision, SomeRequest<T>>,
+    /// The state after the change
+    state_after: HashMap<Revision, SomeRequest<T>>,
```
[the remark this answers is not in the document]
alexjg: My reasoning here is that the purpose of this feature is to help with debugging so the less logic involved in producing transitions the better. Just copying the raw before and after state means that we can be relatively sure that it's a faithful representation of the data the program is working against.

I don't _think_ this should have serious performance implications. The waiting room is generally a reasonably small hashmap and production of these events is relatively infrequent - although as I say this it occurs to me that this may not be true for seeds.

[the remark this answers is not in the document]
FintanH: > My reasoning here is that the purpose of this feature is to help with debugging so the less logic involved in producing transitions the better.

Ya, I can see why that would be useful. My worry would be that it's conflating some debugging and the waiting room's logic. I'll always have the before and after states for debugging, because I have to make the call on it and I get the result.

You're right that the waiting room is generally small and it wouldn't impact it greatly. I'm not sure if a seed would need this code, but you never know :)

So maybe we should look for a way to separate out what the waiting room is doing and ways of tracing the state of affairs. I'm thinking something like:
```rust
pub struct Traced<T, U> {
  before: T,
  after: U,
}

impl<T, U> Traced {
  /// Convert a `T` into a `U` with the supplied function,
  /// but keep track of the before and after states.
  pub trace<T: Clone, U>(t: T, f: impl FnOnce(T) -> U) -> Traced<T, U> {
    let before = t.clone()
    Traced { before, after: f(t) }
  }
}
```

The only thing I'm not sure about here is the higher-order function. It would need to be polymorphic in owned, `&`, and `&mut` but maybe we just write a load of boiler-plate `trace`, `trace_ref`, and `trace_mut` :upside_down_face: 

What do you think of this? Do you see another way of separating the concerns?

daemon/src/request/event.rs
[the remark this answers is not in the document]
FintanH: But the caller _has_ the waiting room? It's not doing anything other than giving you back the thing you just had but cloned.

So it feels like this is only interesting based on the caller's use case, in our case the subroutines. So maybe we could shift the tracing of this event there?

daemon/src/request/waiting_room.rs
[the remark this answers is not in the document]
alexjg: I see what you're getting at. My main concern when implementing the `state_before` and `state_after` assignments in the waiting room was to avoid exposing the waiting room internals to the rest of the application. However, the fact that the `HashMap<PeerId, SomeRequest<T>>` type is currently exposed in the `WaitingRoomTransition` event emitted by the subroutines somewhat defeats that objective. I think there are two approaches:

1. Expose the `state_before` and `state_after` in the event as `serde_json::Value`s and leave the actual content of the up to the waiting room. This achieves the original implementation-hiding intent and would justify leaving the actual production of those fields within the waiting room. It does make consuming events nicely more complicated because you don't really know what's in them, but as this is primarily for debugging that may not be a concern.
2. Perform the event generation in the subroutines as using your `Traced` type in combination with a helper method to remove some of the repetitiveness. This would probably also mean moving the `Event` enum out of `waiting_room` and into `subroutines`.

I think I'm inclined to the latter approach.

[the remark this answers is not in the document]
FintanH: Ya, to be honest, I think 2. sounds more reasonable to me. It just seems like there's a layering approach where the waiting room is the pure logic, then there's an execution layer in the `subroutines`, and we fit a piece in-between that helps with debugging and tracing. 

What might help the debugging component is having the `WaitingRoom` give enough information about the transitions so that these can be easily convert to `Event`s, if that makes sense?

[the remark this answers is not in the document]
alexjg: > What might help the debugging component is having the WaitingRoom give enough information about the transitions so that these can be easily convert to Events, if that makes sense?

Are you referring here to being able to produce whatever the `before` and `after` fields of `Traced` should be for some given type?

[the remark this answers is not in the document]
FintanH: No sorry, I was referring to providing the caller with checking the `state` of the `Request`. I'm not sure how "inspectable" it currently is so that you can generate the `Event` :)

alexjg:
I've changed the approach here quite a bit. Rather than changing the `WaitingRoom` I've added a wrapper called `RunningWaitingRoom`. This wrapper does the job of populating a `state_before` and `state_after` field on a transition. I also moved most of the waiting room related code from `RunState` into this new `RunningWaitingRoom` which now has the responsibility of translating calls to the `WaitingRoom` into a `Vec<Command>` for the `RunState`.

librad/src/git/types/remote.rs
```diff
@@ -427,7 +427,7 @@ impl Remote<LocalUrl> {
                 specs,
                 Some(
                     git2::FetchOptions::new()
-                        .prune(git2::FetchPrune::On)
+                        .prune(git2::FetchPrune::Off)
```
xla, line 430: Is this bool flip intentionally part of this PR?

daemon/src/peer/run_state.rs
```diff
@@ -594,7 +572,11 @@ mod test {
         let peer_id = PeerId::from(SecretKey::new());
         let addr = "127.0.0.0:80".parse()?;
 
-        let status = Status::Online { connected: 0 };
+        let connected_peers = one_connected_peer(PeerId::from(SecretKey::new()));
+        let status = Status::Online {
+            connected: 0,
```
xla, line 577:
```suggestion
            connected: 1,
```

Should be equal to the number of elements in `connected_peers`.

daemon/src/peer/run_state.rs
```diff
@@ -138,6 +150,8 @@ pub enum Status {
     Online {
         /// Number of connected peers.
         connected: usize,
```
xla, line 152: This seems superfluous with the addition of `connected_peers`.

daemon/src/peer/run_state.rs
```diff
@@ -138,6 +150,8 @@ pub enum Status {
...
+        /// Connected peers
+        connected_peers: HashMap<PeerId, Vec<SocketAddr>>,
```
xla, line 154: Are the values of the map used somewhere? Or is this in preparation for another change which surfaces this for the up.

xla (commented):
Generally this is a worthwhile strategy to tame the massive `RunState` impl. As it represents the "protocol" of the running daemon, modularising it with sub-protocols by encapsulating, in this case the request handling, appears sound. Going forward we might want to think about a structured approach to that.

librad/src/git/types/remote.rs
alexjg: Nope, reverted.

daemon/src/peer/run_state.rs
alexjg: This map is used to provide the data for the "Peers" panel in the network diagnostics in Upstream.

daemon/src/peer/run_state.rs
alexjg: Yup, I've removed the whole `connected` field. This will require a few small changes in upstream.

daemon/src/peer/run_state.rs
alexjg: See below.

xla (dismissed):
✏️ ⛈ 📦 🌔

daemon/src/peer/run_state/running_waiting_room.rs
```diff
@@ -0,0 +1,345 @@
...
+    pub const fn new(waiting_room: WaitingRoom<SystemTime, Duration>) -> Self {
+        Self { waiting_room }
+    }
+
+    pub fn cancel(
+        &mut self,
+        urn: Urn,
+        timestamp: SystemTime,
+        sender: Sender<Result<Option<SomeRequest<SystemTime>>, WaitingRoomError>>,
+    ) -> Vec<Command> {
```
FintanH, line 106: Not to do this with this particular PR, but:

Something that bothers me is the amalgamation of _all_ `Command`s into one structure. Something we could consider for another patch is that there is `Command` local to the `RunningWaitingRoom` and this can be added as a variant to a larger `Command` enum closer to the run state.

A `RunningWaitingRoom`s transitions and commands should be testable separated from the `RunState`.

daemon/src/peer/run_state/running_waiting_room.rs
```diff
@@ -0,0 +1,345 @@
...
+                ))),
+                Command::PersistWaitingRoom(self.waiting_room.clone()),
+            ],
+        }
+    }
+
+    pub fn request(
+        &mut self,
+        urn: Urn,
+        time: SystemTime,
```
FintanH, line 140: nit: we have `timestamp`, `time`, and `now` namings for these parameters. Can we settle on one? :)

daemon/src/peer/run_state/running_waiting_room.rs
```diff
@@ -0,0 +1,345 @@
...
+        urn: &Urn,
+        event: Event,
+        timestamp: SystemTime,
+        f: F,
+    ) -> Vec<Command>
+    where
+        F: FnOnce(&mut WaitingRoom<SystemTime, Duration>) -> Result<(), WaitingRoomError>,
+    {
+        let state_before = waiting_room.requests();
+        // FIXME(alexjg): Come up with a strategy for the results returned by the
```
FintanH, line 296: What do you mean here?

FintanH (commented):
This is looking good :) I like the separation and it makes the `run_state` cleaner :ok_hand:

daemon/src/peer/run_state/running_waiting_room.rs
alexjg: This is actually a copy of a `FIXME` @xla had in `run_state.rs`. I've assumed responsibility for it because I'm  touching this code and not fixing it. The basic problem is that we don't actually report the error in this result to the client, we just log it and carry on, it seems like we should probably have some API for telling the client that something has gone wrong.

daemon/src/peer/run_state/running_waiting_room.rs
alexjg: Fixed

daemon/src/peer/run_state/running_waiting_room.rs
FintanH: Ahhh that makes sense. Could you maybe change the wording to mention reporting the error back? Or even move the comment to the error case.

daemon/src/peer/run_state/running_waiting_room.rs
alexjg: It's something to consider. Splitting `Command` isn't without costs. The problem is that most of the "subprotocols" of the `RunState` need to emit commands related to other subprotocols. I'm not convinced that adding indirection to make each subprotocol unaware of the other would make the code easier to follow. Regardless, probably the best way to do this would be to first split each subprotocol into it's own bit of state as I've done here, and then split the `Command` to see how it looks.

As to the testing point, it feels to me like the natural place to test this stuff is on the `RunState`, that way we're free to change implementations significantly (as this PR does) without having to update all our tests. I have a bias towards more integration-ey tests though.

daemon/src/peer/run_state/running_waiting_room.rs
alexjg: Done

PR #595, opened 2021-04-09
replication: ensure we don't prune tracked peers

FintanH:
In the clone path of replication we only kept the delegate peers. We
would also want to keep the tracked peers to preserve the tracking
graph.

This patch also adds a test in our menage scenario to confirm a 3rd
peers sees the other 2 peers collab'ing.


xla (approved):
> This patch also adds a test in our menage scenario to confirm a 3rd
peers sees the other 2 peers collab'ing.

Came here to find a variable called voyeur, left duly disappointed.

🗻 🈁 🍙 📹

FintanH:
> Came here to find a variable called voyeur, left duly disappointed.

PRs welcome :upside_down_face:

PR #596, opened 2021-04-09
replication: Prevent fetch of non-existing repo 

xla:
Up until fetches would go through even if the project doesn't exist on
the server. Due to the use of pruning during the fetch this would leave
remotes in a broken state with a tree looking like:

```
 └── hyygmx7fm9zaneg64d7gjuohwwmo1obduwk4s6jku3yau9k8y3dgjs
    ├── heads
    │   └── main
    └── rad
        └── ids
```

### DIFF
--- a/librad/src/git/fetch.rs
+++ b/librad/src/git/fetch.rs
@@ -547,7 +547,7 @@ impl<'a> DefaultFetcher<'a> {
             .to_string(),
         )?;
         remote.connect(git2::Direction::Fetch)?;
-        let remote_heads = remote
+        let remote_heads: RemoteHeads = remote
             .list()?
             .iter()
             .filter_map(|remote_head| match remote_head.symref_target() {
@@ -620,7 +620,7 @@ impl<'a> DefaultFetcher<'a> {
                 &refspecs,
                 Some(
                     git2::FetchOptions::new()
-                        .prune(git2::FetchPrune::On)
+                        .prune(git2::FetchPrune::Off)
                         .update_fetchhead(false)
                         .download_tags(git2::AutotagOption::None)
                         .remote_callbacks(callbacks),
--- a/librad/tests/smoke/clone.rs
+++ b/librad/tests/smoke/clone.rs
@@ -21,6 +21,59 @@ use librad_test::{
 
 const NUM_PEERS: usize = 2;
 
+/// Fetching from a peer that does not have the identity should leave the
+/// `rad/*` refs intact.
+#[tokio::test]
+async fn not_present() {
+    logging::init();
+
+    let peers = testnet::setup(3).await.unwrap();
+    testnet::run_on_testnet(peers, 3, |mut peers| async move {
+        let maintainer = Host::init(peers.pop().unwrap()).await;
+        let contributor = Leecher(peers.pop().unwrap());
+        let voyeur = peers.pop().unwrap();
+
+        let urn = maintainer.project.project.urn().clone();
+        let maintainer_id = maintainer.peer.peer_id();
+
+        contributor.clone_from(maintainer, true).await;
+
+        let cfg = contributor.0.protocol_config().replication;
+        contributor
+            .0
+            .using_storage(move |storage| {
+                // check rad/self of maintainer exists
+                assert!(
+                    storage
+                        .has_ref(&Reference::rad_self(Namespace::from(&urn), maintainer_id))
+                        .unwrap(),
+                    "`refs/remotes/<maintainer>/rad/self` should exist before"
+                );
+
+                let res = replication::replicate(
+                    &storage,
+                    cfg,
+                    None,
+                    urn.clone(),
+                    voyeur.peer_id(),
+                    voyeur.listen_addrs().iter().copied(),
+                );
+                assert!(res.is_ok());
+
+                // check rad/self of maintainer exists
+                assert!(
+                    storage
+                        .has_ref(&Reference::rad_self(Namespace::from(&urn), maintainer_id))
+                        .unwrap(),
+                    "`refs/remotes/<maintainer>/rad/self` should exist after"
+                );
+            })
+            .await
+            .unwrap();
+    })
+    .await;
+}
+
 #[tokio::test]
 async fn when_connected() {
     logging::init();


librad/src/git/p2p/server.rs
```diff
@@ -70,6 +71,10 @@ impl GitServer {
             Ok(Header { service, repo, .. }) => match *service {
                 Service::UploadPack => {
                     tracing::info!("upload pack");
+                    if !self.has_repo(&repo).await? {
+                        tracing::error!("Client asked for non-existing repo: {}", repo);
+                        return send_err(&mut send, &format!("missing repo {}", repo)).await;
```
alexjg, line 76: There's no way for us to surface this particular error on the client side right? By the time it gets to `git::Remote::Error` we just have a string.

FintanH: Can you elaborate on what you mean a bit? I'm not sure I follow the question

alexjg: I'm trying to imagine how we would use this to change `librad::git::replicate` so that it can return a more informative error when the remote doesn't have the project. As I understand thing this change will end up as a `git2::Error` value when we call `DefaultFetcher::fetch`, so if we did want to do something with this particular error case we would need to match on the string in `git2::Error`.

FintanH: Well this needs to talk over the wire so I think it'd be pretty tricky to send something over and parse it on the other end. Probably not worth the hassle and just let it fail. I don't think there's much we can do if the fetcher does fail in this scenario anyway, right?

alexjg: That's fair, and it's consistent with the way that the pack protocol returns an error if the client issues a want for an object it doesn't recognise, so it probably is the right thing to do.

It seems unsatisfying that we have no way of letting the client know that the project didn't exist on the remote. I feel that such information would be helpful in debugging misbehaving servers (e.g a peer who advertises a project, then loses the project before we request it from them). I guess in such cases we'll still see the fetch errors and that's probably sufficient.

FintanH: At least we don't lose refs this way :upside_down_face:

alexjg (approved):

kim:
Can you describe what this fixes in more detail? It seems to me like the problem is more related to pruning.

librad/src/git/p2p/server.rs
```diff
@@ -78,6 +83,10 @@ impl GitServer {
                 },
                 Service::UploadPackLs => {
                     tracing::info!("upload pack ls");
+                    if !self.has_repo(&repo).await? {
```
kim, line 86: Why do this twice, when we could return early before matching on the service?

librad/src/git/p2p/server.rs
kim: The problem is due to libgit2 bailing with “Early EOF” or something, instead of passing the error through. I consider it a bug.

kim (changes requested):
While I would still like to understand how the issue manifests, I think this patch applies the band-aid to the wrong limb regardless:

1. The “server” end is never trusted, so cannot be relied upon to “do the right thing” — instead, the fetch end needs to look at the advertised refs and return early if the fetchspecs we compute would never actually match
2. The fact that we’re pretending to fetch from a single repo (instead of a namespaced gigarepo) is due to the lack of protocol v2 support — once we have that, we can fetch many namespaces in one go by giving a prefix filter to the refs advertisement.

kim:
(2. means that the server would never be able to check if the requested namespaces “exist” unless the protocol commands would be interpreted fresh off the wire)

FintanH:
Thanks for the pointers @kim 

> While I would still like to understand how the issue manifests

I'll shed some light on how this was causing issues. During some QA testing of Upstream we had the following scenario:
1. The person would be onboarded
2. They would create a project that would land on the `setzling` node
3. The person would remove `active_profile` and onboard again
4. They would replicate the project (this would include the remote from the peer in step 1)
5. They switch the seed node address to `sprout` -- which doesn't have the project
6. A fetch would be attempted for the project to `sprout` (admittedly this is also a band-aid, but probably a good thing we found this case due to it :))

Step 6 would result in the `remote/<peer>/rad/*` references being removed and the state of the storage being broken, due to the project not existing and the pruning option being on.

> The “server” end is never trusted, so cannot be relied upon to “do the right thing” — instead, the fetch end needs to look at the advertised refs and return early if the fetchspecs we compute would never actually match

Ah, I see your point. I thought the server guarding against this would make sense, but in a trustless environment, this makes more sense. My bad :hand: 
We can check the advertised refs at the point where we list the remote heads, right? Or is it at a different point in the interaction?

kim:
> Step 6 would result in the `remote/<peer>/rad/*` references being removed and the state of the storage being broken, due to the project not existing and the pruning option being on.


That’s the part I don’t quite understand, and looking at the code doesn’t help much — shouldn’t we just prune the difference between the tracked remotes and the existing ones?

alexjg:
The issue here was the pruning option being `On` when doing git fetch. This causes git to remove references which match the respecs but which are not in the remote. E.g this code in `librad/src/git/fetch.rs`

```
            self.remote.fetch(
                &refspecs,
                Some(
                    git2::FetchOptions::new()
                        .prune(git2::FetchPrune::On)
                        .update_fetchhead(false)
                        .download_tags(git2::AutotagOption::None)
                        .remote_callbacks(callbacks),
                ),
                None,
            )?;
```

I'm thinking we should probably remove that altogether, it feels risky to have git silently changing things underneath us. My understanding is that we would then need to arrange for garbage collection to be run in some other manner.

FintanH:
> That’s the part I don’t quite understand, and looking at the code doesn’t help much — shouldn’t we just prune the difference between the tracked remotes and the existing ones?

Sorry, I should have clarified since the terminology is overloaded. As @alexjg pointed out, the pruning we're referring to here is essentially the [`--prune`](https://git-scm.com/docs/git-fetch#Documentation/git-fetch.txt---prune) option during a fetch.

From the docs linked above:
> Before fetching, remove any remote-tracking references that no longer exist on the remote.

This seems to be triggered in this case, since the server doesn't have the project, the client prunes them.

This option is enabled [here](https://github.com/radicle-dev/radicle-link/blob/master/librad/src/git/fetch.rs#L623). Iirc this is enabled so that we enable the `gc` to do its thing. So I opted for, wrongly, the server to be defensive. And now in the latest commit, the client to be defensive.

kim:
Uh, yeah, that should be off (see, one can’t get booleans the right way,
even when they have different names)

FintanH:
Mystery solved 😂

kim (dismissed):
👍

Issue #584, opened 2021-04-05
Allow advertising predefined address


viraptor:
Not sure if this belongs just to -bins or does it extend to -link, but just like the webserver has a listen and a public address, radicle should have the same option. Currently I'm seeing advertisements from my seed like this:
```
 2021-04-05T15:38:20.333 INFO  librad::net::protocol::membership::hpv      > apply; remote_peer=hybk76pim74znj9tapyarr3qmxt1qyc9u1rfk9cda63pdsq9rrz8o4 remote_addr=35.197.169.21:12345 rpc=Neighbour { info: PeerAdvertisement { listen_addrs: {10.152.0.4:12345}, capabilities: {} }, need_friends: None }
```
However that seed is never going to be reachable for others on the `listen_addrs`.
(Bonus points for discovery via STUN https://crates.io/crates/rustun)

kim:
> just like the webserver has a listen and a public address, radicle should have the same option

Yeah, not a bad idea. If you want to give it a spin, the net::peer::Config would be a good place to start.

> However that seed is never going to be reachable for others on the listen_addrs.

That’s not universally true, though: a regular p2p node could very well wish to be reachable on the local network, and NAT hairpinning behaviour is not something we could reliably detect. Peers will add the address they see the other end as before propagating peer information, so the advertised address is not the only option.

That being said, a config option to turn off advertisement of a private address in case you know what you’re doing isn’t a bad idea either.

> Bonus points for discovery via STUN

STUN only works when both ends agree on a well-known server — which is not exactly what we mean when we say “decentralised”. So... if you have a proposal of how to run an open STUN network which incentivises people to participate, converges towards a set of honest, reliable, and properly configured nodes, and yields overall lower latency and higher success rate than the brute force method of just trying all addresses a node _appears_ to have (considering that STUN is basically a brute force method of trying all addresses a node _appears_ to have), I would be very interested to hear it ;)

viraptor:
> a regular p2p node could very well wish to be reachable on the local network

Correct, I wouldn't want to remove that option. I meant *that specific* node will never be reachable that way.

> the net::peer::Config would be a good place to start

Thanks, I will have a look.

> STUN only works when both ends agree on a well-known server

We could still use a very limited part of stun. Only the "query for address" functionality, where it's up to the local node to decide who to ask. (With some sane default)

kim:
> We could still use a very limited part of stun. 

[...]

PR #597, opened 2021-04-10
Allow configuring advertised address

viraptor:
In many configurations the address which can be seen by other nodes will
be completely different than the socket address used for listening. For
example a seed with address 0.0.0.0 will advertise localhost, all docker
interfaces, link-local ipv6 addresses, etc. but not the public address
it's unaware of.

This change introduces the minimal first step towards a better
advertisement. A static advertised address is exposed in configuration.

Future direction may include public address autodiscovery using STUN
endpoints, or asking other radicle nodes to report back the seen remote
address.

Fixes https://github.com/radicle-dev/radicle-link/issues/584

### DIFF
--- a/librad/src/net/protocol.rs
+++ b/librad/src/net/protocol.rs
@@ -19,6 +19,7 @@ use futures::{
     stream::StreamExt as _,
 };
 use governor::{Quota, RateLimiter};
+use nonempty::NonEmpty;
 use nonzero_ext::nonzero;
 use parking_lot::Mutex;
 use rand_pcg::Pcg64Mcg;
@@ -66,6 +67,7 @@ mod tick;
 pub struct Config {
     pub paths: Paths,
     pub listen_addr: SocketAddr,
+    pub advertised_addrs: Option<NonEmpty<SocketAddr>>,
     pub membership: membership::Params,
     pub network: Network,
     pub replication: replication::Config,
@@ -264,8 +266,13 @@ where
 {
     let local_id = PeerId::from_signer(&signer);
     let git = GitServer::new(&config.paths);
-    let quic::BoundEndpoint { endpoint, incoming } =
-        quic::Endpoint::bind(signer, config.listen_addr, config.network).await?;
+    let quic::BoundEndpoint { endpoint, incoming } = quic::Endpoint::bind(
+        signer,
+        config.listen_addr,
+        config.advertised_addrs,
+        config.network,
+    )
+    .await?;
     let (membership, periodic) = membership::Hpv::<_, SocketAddr>::new(
         local_id,
         Pcg64Mcg::new(rand::random()),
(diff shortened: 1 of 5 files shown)


librad/src/net/protocol.rs
```diff
@@ -66,6 +66,7 @@ mod tick;
 pub struct Config {
     pub paths: Paths,
     pub listen_addr: SocketAddr,
+    pub advertised_addr: Option<SocketAddr>,
```
kim, line 69: Should be a `Vec`, to be able to match `LOCALHOST_ANY` semantics

librad/src/net/protocol.rs
```diff
@@ -66,6 +66,7 @@ mod tick;
 pub struct Config {
     pub paths: Paths,
     pub listen_addr: SocketAddr,
+    pub advertised_addrs: Option<Vec<SocketAddr>>,
```
kim, line 69: Sorry for nitpicking, but `Option<Vec<T>>` is not a terribly useful type: if it’s `None`, we know the list has no elements, but if it’s `Some` it may also not have any elements.

Since the list is allowed to be empty, `Vec` would do, otherwise we usually use `NonEmpty`

viraptor: That's a very good point. I want to keep option so that I can use .map and others. I'll make it NonEmpty instead.

librad/src/net/protocol.rs
```diff
@@ -66,6 +67,7 @@ mod tick;
 pub struct Config {
     pub paths: Paths,
     pub listen_addr: SocketAddr,
+    pub advertised_addrs: Option<NonEmpty<SocketAddr>>,
```
FintanH, line 70: `Option<NonEmpty<T>` is the same thing as `Vec<T>` -- you either don't have any elements, i.e. `None === Vec::empty()`, or you do, i.e. `Some(NonEmpty { head, tail }) === vec![head].append(tail)` :) So I'm not entirely sure this is buying us anything :thinking:

viraptor: I have strong opinions about this being significant in theory. Specifically not having an explicit list of addresses to advertise is different than having an empty list (so not advertising anything) - even if it's not currently allowed. It makes the difference more explicit when reading code.

In practice as you said, it's all the same, so ¯\_(ツ)_/¯ happy to go with popular vote if I'm not convincing :-)

FintanH: > Specifically not having an explicit list of addresses to advertise is different than having an empty list

This makes sense to me, so you can consider me convinced :grin:

FintanH (approved):

kim:
Closed via 2bddee7c8f603d7717acab32c85d8a6044d6fd76

librad/src/net/protocol.rs
kim: > Specifically not having an explicit list of addresses to advertise is different than having an empty list (so not advertising anything

What would be the semantics of this?

PR #599, opened 2021-04-12
ci: fix conditionals for jobs to be skipped

kim:
Found the right incantations to skip build steps on the two CI services,
depending on whether a PR is from a fork or not.

Note that the "required checks" feature doesn't support conditionals, so
becomes meaningless. That should be okay, though, as Teh Green Button is
verboten anyways.


geigerzaehler:
In upstream we’re using the following expression
```
if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != 'radicle-dev/radicle-upstream'
```

https://github.com/radicle-dev/radicle-upstream/blob/47ced24620841fb74d35c9862dc8c9617fb7366e/.github/workflows/build.yaml#L8-L10

PR #600, opened 2021-04-12
ci: don't skip buildkite steps if not a PR

kim:
Also, GHA seems to not correctly evaluate `== true|false`.

### DIFF
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -3,8 +3,9 @@ steps:
   # triggering `n` image builds when updating the `DOCKER_IMAGE`. Note that it
   # still saves about 20sec to pin the image to a SHA256 after an upgrade.
   - label: "Build container"
-    # Nb: `if` skips the step if the expression evaluates to false
-    if: build.pull_request.repository.fork == false
+    if: |
+        build.pull_request.repository.fork == null ||
+        build.pull_request.repository.fork == false
     commands:
       - ".buildkite/env"
       - ".buildkite/build-container"
@@ -18,7 +19,9 @@ steps:
   - wait
 
   - label: "Build + Test"
-    if: build.pull_request.repository.fork == false
+    if: |
+        build.pull_request.repository.fork == null ||
+        build.pull_request.repository.fork == false
     commands:
       - ".buildkite/env"
       - "ci/build-test"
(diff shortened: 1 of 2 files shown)


kim:
je ne comprend pas les booleans

.buildkite/pipeline.yaml
```diff
@@ -4,7 +4,7 @@ steps:
   # still saves about 20sec to pin the image to a SHA256 after an upgrade.
   - label: "Build container"
     # Nb: `if` skips the step if the expression evaluates to false
-    if: build.pull_request.repository.fork == false
+    if: build.pull_request.repository.fork != true
```
FintanH, line 7: Wait... WHAT?!

kim: Please consider donating to "ASSOCIATION OF COMPUTER SCIENTISTS AND PRACTICIONERS FOR THE ABOLISHMENT OF YAML AND BOOLEAN LOGIC" (AOCSAPFTAOYABL)

FintanH: I'll donate some RAD

kim:
I think this does actually work now, but the checks display on the front matter gets confused because the companion PR has the same SHA

.github/workflows/fast.yaml
```diff
@@ -2,7 +2,7 @@ name: fast
 on: [pull_request]
 jobs:
   fmt:
-    if: github.event.pull_request.head.repo.fork == true
+    if: ${{ github.event.pull_request.head.repo.fork }}
```
FintanH, line 5: Assuming `null` means `false` here :smile:

FintanH (approved):

FintanH:
> I think this does actually work now, but the checks display on the front matter gets confused because the companion PR has the same SHA

You could try test it by opening a PR from a fork? :shrug:

kim:
> You could try test it by opening a PR from a fork?

That's what I did: #601 

If you switch to the "Checks" tab, you'll see two actions, one from this branch and the other from the fork. This branch correctly skipped the redundant jobs, the other one ran them. Supposedly because both PRs have the same SHA, the wrong checks are shown on the PR front page.

kim:
Oh, and the BK check shown on the fork PR links to the build for THIS PR. I guess because of the same reason.

PR #603, opened 2021-04-12
replication: Surface errors

xla:
Don't swallow to give more context when fetch and sign errors occur.

### DIFF
--- a/librad/src/git/replication.rs
+++ b/librad/src/git/replication.rs
@@ -62,10 +62,10 @@ pub enum Error {
     #[error(transparent)]
     Track(#[from] tracking::Error),
 
-    #[error("signer error")]
+    #[error("signer error: {0}")]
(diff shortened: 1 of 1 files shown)


alexjg (approved):

kim:
This pattern has been condemned to be frowned upon by [Radicle Decision 0002](https://github.com/radicle-dev/radicle-decisions/blob/master/proposals/0002.md). If we do this anyways, I would request to not use the debug formatter, as defaulting to that (with no option to change it) is basically what makes `tracing` less-than-useful for logging.

xla:
@kim I can't really see what 0002 is proposing for a case like the one we have at our hands here. Reckon doesn't hurt to not use Debug formatting.

kim:
https://github.com/radicle-dev/radicle-decisions/blob/master/proposals/0002.md#the-behavior-of-display:

> It should never include information provided by the error returned from the source.

kim (dismissed):

PR #605, opened 2021-04-12
spec: make rendered html the GH pages index

### LINKED ISSUE #531: spec: Stylez

kim:
Rendering the RFCs would just be too laborious. Perhaps they could be
added as appendices to the spec at some point.

Closes #531

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -1,7 +1,1 @@
-<html>
-    <head></head>
-    <body>
-        <h1>Radicle Link</h1>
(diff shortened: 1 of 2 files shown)


FintanH (approved):

xla (approved):
🚐 🍽 🎗 🎋

PR #606, opened 2021-04-13
e2e: simple docker-compose network

kim:
Dumb down the procfile, and use compose to spin up a more elaborate
stack. Note also that mDNS appears to have stopped working, so is no
longer used.


e2e/README.md
```diff
@@ -2,19 +2,28 @@
...
+the `quick` script. This will start `$NUM_PEERS`, plus a bootstrap node, using
+[overmind].
 
-You can adjust the number of peers by setting the `NUM_PEERS` environment
-variable.
+A more elaborate setup is provided via [docker-compose]. Run
 
-_Note that this is currently simply a debugging aid. More sophisticated
-orchestration may be added in the future (contributions welcome!)_
+    DOCKER_BUILDKIT=1 docker-compose up
```
FintanH, line 19: When trying to run this I got:
```
ERROR: 
        Can't find a suitable configuration file in this directory or any
        parent. Are you in the right directory?

        Supported filenames: docker-compose.yml, docker-compose.yaml
```

Renaming `compose.yaml` to `docker-compose.yaml` allowed it to work. Maybe a versioning issue?

FintanH (approved):
Was able to get the `quick` script and the `docker-compose` running and played with some graphs :)

I just ran into one issue around the file naming.

e2e/README.md
kim: You should be able to tell it which yamls to look at using `docker-compose -f the.yaml`. I'd rather change the readme than the filename, because it's more likely that we will have more than one compose file.

e2e/ephemeral-peer.dockerfile
```diff
@@ -0,0 +1,20 @@
+# syntax=docker/dockerfile:1.2
+# Nb: the "latest" tag needs to be managed manually, CI doesn't update it
+FROM gcr.io/opensourcecoin/radicle-link-seedling-build:latest AS build
+WORKDIR /build
+RUN --mount=type=bind,source=.,target=/build,rw \
```
kim, line 5: You know what's really taking all the fun out of this? If we _don't_ add `target/` to `.dockerignore`, `docker build` will send like 60GiB to the doge daemon. _However_, it apparently also respects `.dockerignore` when mounting, so we never see any caching.

e2e/README.md
FintanH: Cool, good to know :+1:

xla (dismissed):
🎇 💀 💬 🈴

PR #609, opened 2021-04-13
data: kickstart the radicle-data crate

kim:
To give a home to those generally useful datastructures which are not
found elsewhere, or are not sophisticated enough to begin living a life
of their own, we introduce the `radicle-data` crate.

For starters, a couple of non-empty container types are exported.

### DIFF
--- a/data/Cargo.toml
+++ b/data/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "radicle-data"
+version = "0.1.0"
+authors = ["The Radicle Team <dev@radicle.xyz>"]
+edition = "2018"
+license = "GPL-3.0-or-later"
+
+[dependencies]
+nonempty = "0.6"
+
+[dependencies.minicbor]
+version = ">= 0.6, 0"
+features = ["std"]
+optional = true
+
+[dependencies.serde]
+version = "1.0"
+optional = true
--- a/data/README.md
+++ b/data/README.md
@@ -1,0 +1,1 @@
+Assorted datastructures.
--- a/data/src/nonempty.rs
+++ b/data/src/nonempty.rs
@@ -1,0 +1,384 @@
+// Copyright © 2019-2020 The Radicle Foundation <hello@radicle.foundation>
+//
+// This file is part of radicle-link, distributed under the GPLv3 with Radicle
+// Linking Exception. For full terms see the included LICENSE file.
+
+use std::{
+    borrow::Borrow,
+    collections::{BTreeMap, BTreeSet, HashMap, HashSet},
+    hash::{BuildHasher, Hash},
+    ops::Deref,
+};
+
+pub use nonempty::NonEmpty as NonEmptyVec;
+
+/// Alias for a [`NonEmpty`] backed by a [`HashSet`]
+pub type NonEmptyHashSet<T> = NonEmpty<HashSet<T>>;
+
+/// Alias for a [`NonEmpty`] backed by a [`BTreeSet`]
+pub type NonEmptyOrderedSet<T> = NonEmpty<BTreeSet<T>>;
+
+/// Alias for a [`NonEmpty`] backed by a [`HashMap`]
+pub type NonEmptyHashMap<K, V, S> = NonEmpty<HashMap<K, V, S>>;
+
+/// Alias for a [`NonEmpty`] backed by a [`BTreeMap`]
+pub type NonEmptyOrderedMap<K, V> = NonEmpty<BTreeMap<K, V>>;
+
+/// Types which may be empty.
+///
+/// A [`NonEmpty`] can only be constructed from implementors when `is_empty`
+/// returns `false`.
+pub trait MaybeEmpty {
+    /// `true` if this type does not contain any elements
+    fn is_empty(&self) -> bool;
+}
+
+impl<T> MaybeEmpty for BTreeSet<T> {
+    fn is_empty(&self) -> bool {
+        self.is_empty()
+    }
+}
+
+impl<T> MaybeEmpty for HashSet<T> {
+    fn is_empty(&self) -> bool {
+        self.is_empty()
+    }
+}
+
+impl<K, V> MaybeEmpty for BTreeMap<K, V> {
+    fn is_empty(&self) -> bool {
+        self.is_empty()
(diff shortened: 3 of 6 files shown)


data/src/nonempty/map.rs
```diff
@@ -0,0 +1,251 @@
...
+/// contains at least one element.
+///
+/// Non-mutating methods of the underlying map are available through the
+/// [`Deref`] impl. Additionally, mutating methods which cannot leave an empty
+/// map behind are provided. To construct a [`NonEmptyMap`], use the [`TryFrom`]
+/// impl.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct NonEmptyMap<K, V, Inner = BTreeMap<K, V>> {
+    inner: Inner,
+    _marker: PhantomData<(K, V)>,
```
FintanH, line 31: The `PhantomData` is here to make the compiler not complain about unused `K, V`, right?

FintanH (dismissed):

data/src/nonempty/map.rs
kim: Yeah Rust. What I'd actually wanted was something like this:

```rust
trait IsEmpty {
    fn is_empty(&self) -> bool;
}

pub struct NonEmpty<T>(T);

impl<T: IsEmpty> TryFrom<T> for NonEmpty<T> {
    type Error = TheEmptinessInside;

    fn try_from(t: T) -> Result<Self, Self::Error> {
        if t.is_empty() { Err(TheEmptinessInside) } else { Ok(Self(t)) }
    }
}
```

but that seems to create many-a-headache with coherence.

data/src/nonempty/map.rs
```diff
@@ -0,0 +1,251 @@
...
+        self.inner.iter_mut()
+    }
+
+    pub fn values_mut(&mut self) -> impl Iterator<Item = &mut V> {
+        self.inner.values_mut()
+    }
+}
+
+impl<K, V> TryFrom<BTreeMap<K, V>> for NonEmptyMap<K, V> {
+    type Error = ();
```
alexjg, line 79: Is having an empty error type going to lead to confusion? I know it's pretty obvious that there's only really one way to fail to construct a non empty collection, but if you're looking at a trace from a program that's just crashed I feel like it will be much quicker to figure out what's going wrong if the error explicitly says that you tried to construct a non empty collection without values rather than having no message.

data/src/nonempty/map.rs
kim: Hah, wait. If we don't use `TryFrom` and add specialised `Extend` impls, this seems to work.

data/src/nonempty/map.rs
kim: Ya. I think it should be a contructor returning `Option<Self>`, really. Wait for reroll.

data/src/nonempty.rs
```diff
@@ -0,0 +1,293 @@
...
+
+impl<T> NonEmpty<T>
+where
+    T: MaybeEmpty,
+{
+    /// Construct a [`NonEmpty`] from a possibly empty type.
+    ///
+    /// If the argument is empty, ie. [`MaybeEmpty::is_empty`] evaluates to
+    /// `true`, [`None`] is returned.
+    pub fn from_maybe_empty(maybe_empty: T) -> Option<Self> {
```
kim, line 155: `try_from_maybe_empty_perhaps` :thinking:

data/src/nonempty.rs
```diff
@@ -0,0 +1,293 @@
...
+///
+/// Mutating methods which either grow the container or don't change its size
+/// are provided via the [`Set`] and [`Map`] impls, respectively. Additionally,
+/// [`Extend`] is implemented if the container implements it.
+///
+/// To construct a [`NonEmpty`], use [`NonEmpty::from_maybe_empty`].
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct NonEmpty<T>(T);
+
+impl<T> NonEmpty<T>
```
FintanH, line 147: I think one thing that would be useful for using this is having some kind of `pub fn new` that takes the first element. Sometimes you know you're going to have the first element at hand and then `insert` the rest. Otherwise, we'll likely end up doing `from_maybe_empty(set).expect("set is not empty. qed.")` :upside_down_face: 

It could be in terms of a `trait Insertable` and so we would have:
```rust
pub fn new(v: T::Value) -> Self where T: Insertable + Default {
  let mut container = T::default();
  container.insert(v);
  Self(container)
}
```

data/src/nonempty.rs
FintanH: `maybe_nonempty`? :eyes:

data/src/nonempty.rs
kim: Good point. I don't see a way to overload this method, though.

FintanH: https://github.com/radicle-dev/radicle-link/pull/610/commits/b1c97d14420aef159f58e1a66ae75e3a238f1fc1

FintanH: :shrug: I guess they're more used to `expect`ing the worst :)

data/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "radicle-data"
+version = "0.1.0"
+authors = ["The Radicle Team <dev@radicle.xyz>"]
```
xla, line 4: The license headers mention the foundation, should the info match or is there merit to have different contact points? Also totally unrelated to this change so can capture in an issue if worth discussing further.

data/README.md
```diff
@@ -0,0 +1 @@
+Assorted datastructures.
```
xla, line 1: Quality!

xla (approved):
LGTM

🔽 🍝 🤖 ⛵️

data/Cargo.toml
kim: I... don't know. Did we change the policy at some point? Did we even _think_ about it?

My IANAL take would be: the foundation is the copyright holder by default, but the developers of radicle-link are not necessarily all affiliated with the foundation. Authors of substantial contributions do have the right to add their own name to the license header, or not mention the foundation at all.

PR #611, opened 2021-04-14
net: bump tracked conn epoch when not reaped

kim:
There is a potential race when tickles get delayed for some reason
(maybe because we're waiting on I/O?).

Also, use SeqCst ordering throughout, until I feel I grok the C++ memory
model, which might be never.

### DIFF
--- a/librad/src/net/quic/connection/tracking.rs
+++ b/librad/src/net/quic/connection/tracking.rs
@@ -6,7 +6,7 @@
 use std::{
     hash::BuildHasherDefault,
     sync::{
-        atomic::{AtomicUsize, Ordering::*},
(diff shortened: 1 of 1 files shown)


librad/src/net/quic/connection/tracking.rs
```diff
@@ -240,13 +238,28 @@ fn spawn_gc(
                 None => break,
                 Some(epoch) => {
                     let prev_epoch = epoch.fetch_add(1, SeqCst);
+                    let curr_epoch = epoch.load(SeqCst);
```
kim, line 241: You know why I did this? Because I **appreciate** vertical alignment! But then, automatic code formatters were invented ... :crying_cat_face:

PR #613, opened 2021-04-15
seed: migrate Rust code back

FintanH:
Some Rust code was living in radicle-bins which made churn in some
librad components annoying to iterate on. This patch migrates that
code back here as the first step.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -204,4 +204,5 @@ allow-registry = ["https://github.com/rust-lang/crates.io-index"]
 # List of URLs for allowed Git repositories
 allow-git = [
     "https://github.com/ZcashFoundation/ed25519-zebra",
+    "https://github.com/radicle-dev/radicle-avatar",
 ]
--- a/seed/Cargo.toml
+++ b/seed/Cargo.toml
@@ -6,10 +6,20 @@ edition = "2018"
 license = "GPL-3.0-or-later"
 
 [dependencies]
+argh = "0.1"
 async-trait = "0.1"
 futures = "0.3"
 librad = { path = "../librad" }
+num_cpus = "1"
 radicle-keystore = "0"
+serde = "1"
 thiserror = "1"
 tokio = { version = "1.1", features = ["time"] }
+tokio-stream = "0.1"
 tracing = "0.1"
+tracing-subscriber = "0.2"
+warp = { version = "0.3", default-features = false }
+
+[dependencies.radicle-avatar]
+git = "https://github.com/radicle-dev/radicle-avatar"
+branch = "main"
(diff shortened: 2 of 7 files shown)


xla (changes requested):
So we are left with radicle-bins being a glorified Cargo.lock?

The cli part is misleading it's a single binary that serves one purpose it's not really exposing an interface which is used to achieve heterogeneous tasks.

FintanH:
> So we are left with radicle-bins being a glorified Cargo.lock?

That's the goal.

> The cli part is misleading it's a single binary that serves one purpose it's not really exposing an interface which is used to achieve heterogeneous tasks.

Open to suggestion. `exe`?

kim:
Does that mean we stabilise the API for any UI frontend one might want (or not)?

FintanH:
> Does that mean we stabilise the API for any UI frontend one might want (or not)?

Right, was just gonna come back to mention that the `ui` code is still in `bins`. So ya, I believe that means the API would have to be stabilised for UIs.

xla:
> Right, was just gonna come back to mention that the `ui` code is still in `bins`. So ya, I believe that means the API would have to be stabilised for UIs.

It might be beneficial to have the UI including its API self-contained, or make the API general purpose. Currently it serves the one specific seed dashboard and not much else.

xla:
> It might be beneficial to have the UI including its API self-contained, or make the API general purpose. Currently it serves the one specific seed dashboard and not much else.

Otherwise we just end up with spooky breakage at a distance in subtle ways and at runtime.

kim:
Yeah, so, I have some concerns about that:

1. There is no reason a seed node needs to have a UI at all
2. There is no reason a seed node must have a _canonical_ UI
3. There is no reason all seed nodes must have the same UI
4. There is no reason a seed node must have a UI which is limited to what it currently does (cf. @cloudhead's plans for "radicle on the web")
5. I am not very keen on moving maintainership responsibilities to this repo. Esp. in light of 4., this could easily extend to a lot of support code which is rather specific to a particular _seed **product**_
6. None of this is tested, so nobody will notice if it breaks
7. If a seed node has some kind of generic API, this must be well-defined, and may not at all be tailored to a web frontend

So, can we either

* Not do this right now, esp. while #576 is in-flight

or

* Only move the code that pertains to "headless" mode here, and recompose in `bins`

or

* Define what the seed API is, and add tests for it?

or all of the above?

xla:
> or all of the above?

I agree on these points, that's why I think esp. the frontend part should not be lightly moved over. We expose a `Node` which is imho the right boundary, if we want it to have it encapsulate more it should be selective.

kim:
It would also be useful to define what the difference between a `daemon` and a
`seed` is, really.

Right now, the only conceptual difference is that a `seed` may auto-track. But
that's not something a `daemon` couldn't also provide.

FintanH:
Ya, fair points. I was being hasty moving this over. I'll close it for now and revisit when we've gone over the `api` migration :ok_hand:

PR #614, opened 2021-04-15
data: add bounded types

kim:
Adds newtypes witnessing the "length" of a type to be within a range.

### DIFF
--- a/data/src/bounded.rs
+++ b/data/src/bounded.rs
@@ -1,0 +1,393 @@
+// Copyright © 2019-2020 The Radicle Foundation <hello@radicle.foundation>
+//
+// This file is part of radicle-link, distributed under the GPLv3 with Radicle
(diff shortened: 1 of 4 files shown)


data/src/bounded.rs
```diff
@@ -0,0 +1,393 @@
...
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error("minimum length {min} not reached: {len}")]
+    TooSmall { min: usize, len: usize },
+
+    #[error("maximum length {max} exceeded: {len}")]
+    TooLarge { max: usize, len: usize },
+}
+
+/// Newtype wrapper for types which have a [`Length`], witnessing that then
```
FintanH, line 79:
```suggestion
/// Newtype wrapper for types which have a [`Length`], witnessing that the
```

FintanH (dismissed):
Big type-level energy

PR #615, opened 2021-04-15
net: promote periodic membership tickle

xla:
To avoid connections being gc'ed when idling, there was a mechanism to
tickle all active ones in the current partial view. So far that was
placed after the await point of the stream processing, so it would not
be executed. To remedy the change tries to be barry algebraic by
introducing a new variant explicitely for the tickler. It does the job
but doesn't quite hit the mark as the current accept::periodic
implementation is very much tailored to the tick tocking - Likely
warrants a bit of a rework.

### DIFF
--- a/librad/src/net/protocol/accept.rs
+++ b/librad/src/net/protocol/accept.rs
@@ -79,20 +79,34 @@ where
                     .into(),
                 }])
             },
+
+            membership::Periodic::Tickle => {
+                // Tickle connections in the partial view.
+                //
+                // This is mostly to keep passive connections from being collected. Note
+                // that we're not checking actual liveness, nor interfering with the
+                // membership protocol.
+                tracing::debug!("initiating tickle");
+                for peer in state.membership.known() {
+                    if let Some(conn) = state.endpoint.get_connection(peer) {
+                        conn.tickle();
+                    }
+                }
+
+                // There are no tocks to evaluate for this case, there is opportunity to improve
+                // this part as sugested in [comment]:
+                //
+                //      This is fine. If it looks funny to you, then because accept also
+                // evaluates      the tocks. More cleansily, this could be two
+                // functions, one which transforms      the periodic events into
+                // tocks, and another one feeds them to the tock-evaluator.
+                //
+                // [comment]: https://github.com/radicle-dev/radicle-link/pull/615#discussion_r614402283
+                stream::iter(vec![])
+            },
         })
         .for_each(|tock| tick::tock(state.clone(), tock))
         .await;
-
-    // Tickle connections in the partial view.
-    //
-    // This is mostly to keep passive connections from being collected. Note
-    // that we're not checking actual liveness, nor interfering with the
-    // membership protocol.
-    for peer in state.membership.known() {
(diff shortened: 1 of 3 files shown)


librad/src/net/protocol/membership/periodic.rs
```diff
@@ -56,7 +57,17 @@ where
         }
     });
 
-    if let Err(e) = stream::select(shuffle, promote).map(Ok).forward(tx).await {
+    let tickle = Interval::new(Duration::from_secs(15), Duration::from_secs(5))
+        .filter_map(|_| future::ready(Some(Periodic::Tickle)));
+
+    // FIXME(xla): There should be a more elegant way to merge these streams into
```
kim, line 63: No I don't think there's a better way

librad/src/net/protocol/accept.rs
```diff
@@ -79,20 +79,28 @@ where
...
+                // that we're not checking actual liveness, nor interfering with the
+                // membership protocol.
+                tracing::debug!("initiating tickle");
+                for peer in state.membership.known() {
+                    if let Some(conn) = state.endpoint.get_connection(peer) {
+                        conn.tickle();
+                    }
+                }
+
+                // FIXME(xla): This indicates that while tickling is a periodic task this method
```
kim, line 96: This is fine. If it looks funny to you, then because `accept` also _evaluates_ the tocks. More cleansily, this could be two functions, one which transforms the periodic events into tocks, and another one feeds them to the tock-evaluator.

kim (dismissed):
:+1:

kim:
That facepalm moment...

librad/src/net/protocol/membership/periodic.rs
```diff
@@ -56,7 +57,17 @@ where
         }
     });
 
-    if let Err(e) = stream::select(shuffle, promote).map(Ok).forward(tx).await {
+    let tickle = Interval::new(Duration::from_secs(15), Duration::from_secs(5))
```
kim, line 60: I think we will need to ensure that this value is always smaller than `net::quic::MAX_IDLE_TIMEOUT`. I know this kind of crosses domains, but perhaps we could just say `MAX_IDLE_TIMEOUT / 2`?

xla: Good catch, was inclined to expose another param similar to promotion and shuffle. Given that this is easy to misconfigure and has a strong relation to other consts in the system For that I'd not expose and turn to the scientif method of half of the max as well.

librad/src/net/protocol/membership/periodic.rs
FintanH: Using another language could be a good option :upside_down_face:

FintanH (commented):
Nice work on tracking this down!

:muscle: :pinching_hand: <- the tickle spot

librad/src/net/protocol/membership/periodic.rs
kim: Consumer! Just get your hands dirty on implementing RFC 424242: Statically Dispatchable Existentially Quantified Types already!

xla: #sauvage

PR #617, opened 2021-04-16
net: use stateless retry

kim:
Popped off the reading list: QUIC stateless retry is essentially a
defense against UDP reflection attacks, adding one (small) roundtrip to
the handshake.

### DIFF
--- a/data/Cargo.toml
+++ b/data/Cargo.toml
@@ -7,6 +7,8 @@ license = "GPL-3.0-or-later"
 
 [dependencies]
 nonempty = "0.6"
+thiserror = "1.0"
+typenum = "1.13"
 
 [dependencies.minicbor]
 version = ">= 0.6, 0"
--- a/data/src/bounded.rs
+++ b/data/src/bounded.rs
@@ -1,0 +1,393 @@
+// Copyright © 2019-2020 The Radicle Foundation <hello@radicle.foundation>
+//
+// This file is part of radicle-link, distributed under the GPLv3 with Radicle
+// Linking Exception. For full terms see the included LICENSE file.
+
+#[cfg(any(feature = "serde", feature = "minicbor"))]
+use std::iter::FromIterator;
+use std::{
+    collections::{BTreeMap, BTreeSet, HashMap, HashSet},
+    marker::PhantomData,
+    ops::Deref,
+};
+
+use thiserror::Error;
+use typenum::{IsLessOrEqual, Unsigned, U0};
+
+/// Types which have a length.
+pub trait Length {
+    fn length(&self) -> usize;
+}
+
+impl<T> Length for Vec<T> {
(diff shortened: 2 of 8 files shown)


FintanH (commented):
I'm finding it hard to understand the section linked. Is the idea that the server knows how to retry packets without doing the cryptographic handshake again? Or am I off the mark?

kim:
It's about requiring the initiator (client) to verify its network address: the client sends `ClientHello`, but instead of a large-ish `ServerHello`, the server sends a small retry token which the client cannot guess, and doesn't commit to any handshake state yet. The client must echo this token back, and so the server can verify that the source address tuple is indeed owned by the client.

Remember that it is very easy to spoof the source address in UDP, which can be used to trick the receiver into sending the response to an innocent third machine. A classic DoS attack.

FintanH:
> It's about requiring the initiator (client) to verify its network address: the client sends ClientHello, but instead of a large-ish ServerHello, the server sends a small retry token which the client cannot guess, and doesn't commit to any handshake state yet. The client must echo this token back, and so the server can verify that the source address tuple is indeed owned by the client.

Ah, that's _WAY_ more clearer :smile: 

> Remember that it is very easy to spoof the source address in UDP, which can be used to trick the receiver into sending the response to an innocent third machine. A classic DoS attack.

So it'd be like `A` asks for a large project from `B` but then `A` malevolently puts `C`'s address in as the one `B` should respond to?

FintanH:
Looking up "reflection attack" helped :) Should have done that first!

FintanH (approved):

FintanH:
> Looking up "reflection attack" helped :) Should have done that first!

For anyone else who might be unfamiliar, the start of this is a good primer https://blog.cloudflare.com/reflections-on-reflections/

kim:
> large project

I don't think this would be possible. To get past the handshake, the client must own the source address. QUIC doesn't allow to change the source address midway without going through a connection migration protocol (which has pretty much the same purpose).

Stateless retry happens _before_ the handshake, so the payload in a reflection attack would be the `ServerHello` (which is not nothing).

FintanH:
> Stateless retry happens before the handshake, so the payload in a reflection attack would be the ServerHello (which is not nothing).

Right, after reading up on it the amplification seems to come from asking more than 1 server to respond to the victim

kim:
Can someone repro the test failure (`smoke::gossip::fetches_on_gossip_notify`) locally? I'd be surprised if it would be related to the change.

kim:
Ah nevermind. I think _someone_ was lazy and thought they'd could get away without minute `spawn_blocking`.

xla (dismissed):
🌎 🕚 ❄️ 🔋

FintanH (dismissed):

PR #619, opened 2021-04-16
smoke: saturate a peer

FintanH:
Trying to replicate a dead-lock that we witness in a scenario between
a client and a seed node.

The real world scenario:
The client periodically creates a test every few seconds. On the 12th
project the seed does not replicate it. There's no output from logging
nor GIT_TRACE_PACKET.

This test scenario:
Peer 1 creates 12 projects and Peer 2 tries to replicate them all in a
loop. The dead-locking is not witnessed and the test passes.

### DIFF
--- a/librad/src/git/p2p/server.rs
+++ b/librad/src/git/p2p/server.rs
@@ -22,7 +22,7 @@ use std::{
 
 use futures::{
     self,
-    io::{AsyncBufReadExt, AsyncRead, AsyncWrite, AsyncWriteExt, BufReader},
+    io::{AsyncBufReadExt, AsyncRead, AsyncReadExt as _, AsyncWrite, AsyncWriteExt, BufReader},
 };
 use git2::transport::Service;
 use git_ext::{into_io_err, RefLike, References, UPLOAD_PACK_HEADER};
@@ -213,17 +213,30 @@ impl UploadPack {
                 let mut stdout = child.stdout.take().unwrap().compat();
 
                 send.write_all(UPLOAD_PACK_HEADER).await?;
-                futures::try_join!(futures::io::copy(&mut stdout, &mut send), child.wait())
-                    .and_then(|(_, status)| {
-                        if !status.success() {
-                            Err(io::Error::new(
-                                io::ErrorKind::Other,
-                                format!("upload-pack ls exited non-zero: {:?}", status),
-                            ))
-                        } else {
-                            Ok(())
-                        }
-                    })
+                let res =
+                    futures::try_join!(futures::io::copy(&mut stdout, &mut send), child.wait());
+
+                // Drive the `RecvStream` to its end to ensure all control messages are received
+                // and corresponding responses are scheduled. If this is not done crucial finish
+                // transitions are missing which determine if control messages to increase the
+                // mutual max stream counter are sent. It will manifest in one end not being
+                // able to open new streams, which in turn is not surfaced as
+                // hard error at this point by the quic lib.
+                //
+                // As there is no data expected on the `RecvStream` the expectation is that this
+                // completes with an ordinary EOF.
+                let mut buf = [0; 1];
+                recv.read(&mut buf).await?;
+
+                let (_, status) = res?;
+                if !status.success() {
+                    Err(io::Error::new(
+                        io::ErrorKind::Other,
+                        format!("upload-pack ls exited non-zero: {:?}", status),
+                    ))
+                } else {
+                    Ok(())
+                }
             },
 
             Self::UploadPack(mut child) => {
--- a/librad/tests/smoke/saturation.rs
+++ b/librad/tests/smoke/saturation.rs
@@ -1,0 +1,115 @@
+// Copyright © 2019-2020 The Radicle Foundation <hello@radicle.foundation>
+//
+// This file is part of radicle-link, distributed under the GPLv3 with Radicle
+// Linking Exception. For full terms see the included LICENSE file.
+
+use std::{ops::Index as _, time::Duration};
+
+use futures::StreamExt;
+
+use librad::{
+    git::{identities, tracking},
+    identities::payload,
+    net::protocol::{
+        event::{self, upstream::predicate::gossip_from},
+        gossip,
+    },
+};
+use librad_test::{
+    logging,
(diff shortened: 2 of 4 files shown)


kim:
I think you forgot to add a file there..

librad/tests/smoke/saturation.rs
```diff
@@ -0,0 +1,55 @@
...
+    logging,
+    rad::{identities::TestProject, testnet},
+};
+
+#[tokio::test]
+async fn saturate_a_peer_with_projects() {
+    logging::init();
+
+    const NUM_PEERS: usize = 2;
+    const NUM_PROJECTS: usize = 11;
```
xla, line 17: This needs to be 12+.

kim: Is 12 a magic number there, yes?

FintanH: There's the initial project too :) So it adds up to 12. Can bump it though if you think that'd make more sense?

FintanH:
> I think you forgot to add a file there..

I did :grimacing: pushed now though :joy:

librad/tests/smoke/saturation.rs
FintanH: > Is 12 a magic number there, yes?

Yup, the seed won't have any fetching logs nor git tracing when the 12th project was announced

kim: My bets are on fd exhaustion...

FintanH:
Going to try to emulate peer 1 gossiping the new projects and peer 2 replicating when they hear the gossip

librad/tests/smoke/saturation.rs
FintanH: I believe @xla took a look at the fd's and didn't see anything out of the ordinary, but I'll let him chime in

librad/tests/smoke/saturation.rs
```diff
@@ -0,0 +1,119 @@
...
+                    }
+                    Ok(())
+                }
+            })
+            .await
+            .unwrap()
+            .unwrap();
+
+        for proj in projs.iter() {
+            let branch = RefLike::try_from(
```
kim, line 80: FYI: I removed this in my checkout. This branch is never created, so the gossip receiver will say “announced non-existent ref”, and refuse to forward the gossip message.

FintanH: Ah :facepalm: thanks! I removed it here as well

librad/src/git/p2p/server.rs
```diff
@@ -211,19 +211,24 @@ impl UploadPack {
...
-                                io::ErrorKind::Other,
-                                format!("upload-pack ls exited non-zero: {:?}", status),
-                            ))
-                        } else {
-                            Ok(())
-                        }
-                    })
+                futures::try_join!(
+                    futures::io::copy(&mut stdout, &mut send),
+                    recv.read_to_end(&mut buf),
```
kim, line 219: How about something like `io::copy(&mut recv, &mut io::sink())`?

librad/src/git/p2p/server.rs
```diff
@@ -211,24 +211,31 @@ impl UploadPack {
...
+                // and corresponding responses are scheduled. If this is not done crucial finish
+                // transitions are missing which determine if control messages to increase the
+                // mutual max stream counter are sent. It will manifest in one end not being
+                // able to open new streams, which in turn is not surfaced as
+                // hard error at this point by the quic lib.
+                //
+                // As there is no data expected on the `RecvStream` the expectation is that this
+                // completes with an ordinary EOF.
+                let mut buf = [0; 0];
+                recv.read(&mut buf).await?;
```
alexjg, line 229: This won't happen if the `try_join` fails right? Should we save the `?` on the `try_join` result until after running `recv.read`?

alexjg (approved):

librad/src/git/p2p/server.rs
```diff
@@ -213,17 +213,32 @@ impl UploadPack {
...
-                        if !status.success() {
-                            Err(io::Error::new(
-                                io::ErrorKind::Other,
-                                format!("upload-pack ls exited non-zero: {:?}", status),
-                            ))
-                        } else {
-                            Ok(())
-                        }
-                    })
+                let res =
```
kim, line 216: Alright, trying to spell out what we want here:

* Drive `child` to completion
    
   There is `kill_on_drop`, but that's not a guarantee. That is, if `child` never terminates, the `server` never terminates. Applying some kind of backpressure on the creation of new git streams would not be a bad idea, but that's beyond the scope of this patch.

* Drive `copy(&mut stdout, &mut send)` to completion.

   Same thing: we need to drain `stdout` of the process.

* Drive `recv`, such that it receives either EOF or one byte of data (which, technically, would be a protocol violation).

* Drop all futures, or none at all


So, I think the `try_join` is good. The `copy(&mut recv, &mut devnull)` would continue to receive if the other end is a tarpit. If the other end is a tarpit, however, `receiving_unknown_size()` will be false, so dropping `recv` SHOULD be fine (in that, the stream gets deregistered).

This means that the previous attempt was better, except that the buffer should be 1 byte in length. If that 1 byte is not zero after `AsyncRead::poll_read`-ing once, this function should call the bad manners police immediately with the remote peer's id card.

I'm ok to merge this as-is for now, and follow up with a more elaborate thing.

kim:
Only question here: do we really gain anything by running the test? I'm not sure it is the case that it would catch the underlying issue if it was re-introduced as a regression. As it's also not a cheap test, it means that it just creates heat.

If it's valuable to have it as a diagnostic/forensic tool, perhaps it could go into `e2e` as an executable?

xla:
> Only question here: do we really gain anything by running the test? I'm not sure it is the case that it would catch the underlying issue if it was re-introduced as a regression. As it's also not a cheap test, it means that it just creates heat.

The smoke test does not complete since the number of projects was upped, which seems to force to hit the upper limit on the streams. I'd still consider it an accurate gauge if this problem is addressed or not, it doesn't hover show the relationship properly which is communicates poorly and will likely bamboozle new and current maintainers in six months alike.

kim (approved):

PR #623, opened 2021-04-19
testing: allow pointing integration tests to any bootstrap node(s)

kim:
This is mainly intended to be able to bootstrap against a large-ish
compose network, and will refuse to join any other network than
"localtestnet". Observe that some tests indeed fail when doing so.

The testnet module is also reworked to use the notorious RAII pattern,
instead of a safe bracket. This makes tests less robust, but is closer
to reality. Observe that it indeed surfaces a few cancellation issues.

Lastly, the example is removed from librad. This had been forgotten, as
it was an intermediate step before the e2e "ephemeral-peer" was
introduced.


librad-test/src/rad/testnet.rs
```diff
@@ -127,54 +134,171 @@ where
...
+    /// None`] is returned.
+    ///
+    ///
+    /// If the variable is not set, the [`Default`] value is returned (which is
+    /// [`Bootstrap::Prev`]).
+    ///
+    /// # Panics
+    ///
+    /// This method panics if `LIBRAD_TEST_BOOTSTRAP` contains a string of the
+    /// form `peer-id@socketaddr`, but that doesn parse/resolve successfully
```
FintanH, line 178:
```suggestion
    /// form `peer-id@socketaddr`, but that doesn't parse/resolve successfully
```

librad/tests/smoke/gossip.rs
```diff
@@ -181,11 +188,10 @@ async fn fetches_on_gossip_notify() {
...
     logging::init();
 
-    let peers = testnet::setup(NUM_PEERS).await.unwrap();
-    testnet::run_on_testnet(peers, NUM_PEERS, |mut peers| async move {
-        let peer1 = peers.pop().unwrap();
-        let peer2 = peers.pop().unwrap();
-
+    let net = testnet::run(config()).await.unwrap();
+    {
+        let peer1 = &net.peers()[0];
```
FintanH, line 193: nit: should we consistently use `index`?

FintanH (approved):

kim:
> nit: should we consistently use `index`?

This is only needed to prevent auto-dereferencing down to the inner `Peer`,
which is only needed for some of the helper stuff. I'd just leave it to the
compiler :shrug:

FintanH:
> This is only needed to prevent auto-dereferencing down to the inner `Peer`,
which is only needed for some of the helper stuff. I'd just leave it to the
compiler

Fair enough :) Thanks for the explanation :bow:

PR #624, opened 2021-04-19
peer: handle task cancellation a bit more gracefully

kim:
Cancellation isn't a reason to panic, and we should not unwind while
shutting down normally.


librad/src/net/peer/storage.rs
```diff
@@ -98,16 +97,12 @@ impl Storage {
             },
         })
         .await
-        .expect("`Storage::git_has` panicked")
+        .unwrap_or(false)
```
FintanH, line 100: Why does it make sense to return `false` instead of propagating the error?

kim:
> Why does it make sense to return `false` instead of propagating the
> error?

That's what we do already, storage errors evaluate to false. This may not have
been the best idea, but is not what this patch is about: the only difference is
that it will not panic if the task was cancelled (and panic in all other cases).

FintanH (dismissed):

xla (dismissed):
🚐 ❗️ 🐳 🚞

PR #625, opened 2021-04-20
git: use dedicated runtime for p2p transport

kim:
This removes nesting of executors, which should make things less
susceptible to hangs because the right incantation of
spawn/spawn_blocking was not adhered to. At least, it improves
reasoning, and allows for some instrumentation.

It may or may not fix the 12 monkeys issue. For that, it may be
advisable to manage _all_ runtime instantiations from librad, instead of
relying on ambient executors.


FintanH (commented):
Does this mean there's a _specific_ executor for the transport and it would be used instead of the one produced by `tokio::main`?

kim:
> Does this mean there's a _specific_ executor for the transport and it
> would be used instead of the one produced by `tokio::main`?

No, it means there is a _single_ executor for the transport, instead of using
the thread-local executor of each thread which creates a transport.

Preferable would be to have only one event loop per process, but since we need
to make the runtime available as a static to the transport, this would mean that
anything built _on top_ of librad would need to use librad's runtime (or else,
be subject to undocumented behaviour when polling futures from different runtime
instances. That is, the issue just moves one layer up).

PR #630, opened 2021-04-22
data: improve construction for known-to-be-within bounds values

kim:
Instead of the lawless Default trait, use std::iter structs with the
desired properties.


FintanH (approved):
Very nice :ok_hand:

PR #632, opened 2021-04-22
net: use a bounded vec for propagated address info

kim:
This is a long-standing open issue: we need to make sure that collection
types are always bounded on the wire.

Note that, while morally a set, set semantics were not captured
precisely:

1. For establishing a connection, we take the union of the seen and
   advertised addresses, with a bias towards the former.
2. For maintaining the seen addresses, we bias towards the more recent
   information.

1. needs to be computed at runtime anyways, while 2. is much simpler to
maintain using a (small) Vec instead of a specialised insertion-ordered
set type. We thus spare ourselves the nuisance of having to thread
through an Ord or Hash constraint on the datatype level.

### DIFF
--- a/librad/Cargo.toml
+++ b/librad/Cargo.toml
@@ -18,6 +18,7 @@ dyn-clone = "1.0"
 futures-timer = "3.0"
 globset = "0.4"
 governor = ">=0.3.2"
+indexmap = "1.6"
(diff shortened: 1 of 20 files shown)


FintanH (approved):
:+1:

PR #634, opened 2021-04-23
net: use if-watch and allow dual-stack sockets

### LINKED ISSUE #532: windows: Figure out if we could do without WinPcap
`pnet_datalink` requires some gymnastics in order to get the linker to see `Packet.lib` from WinPcap on Windows. I don't believe we actually need this, because we're using this library only to enumerate network interfaces.

Maybe there is some other crate to do this, or some feature flags could be introduced downstream to avoid this.

kim:
When binding to INADDR(6)_ANY, use the `if-watch` crate to dynamically
discover and update the listen addresses. This simplifies building on
windows, as we don't need winpcap anymore. It also allows to obtain the
current listen addresses without handling IO errors.

When binding to an ipv6 address, allow the socket to operate in
dual-stack mode if the platform supports it.

Closes #532


FintanH (dismissed):
Cool :sunglasses:

PR #636, opened 2021-04-24
librad: introduce prop feature flag

FintanH:
We introduce a prop feature flag so that we can expose proptest
strategies, outside of tests, in other workspace packages.

This includes adding a new secret key mascot RESKey the risky key. It
uses SmallRng from the rand package for quicker testing.

We move all Arbitrary instances and gen_* functions in to `gen` modules
and make them public under the "prop" flag.

### DIFF
--- a/librad/Cargo.toml
+++ b/librad/Cargo.toml
@@ -5,6 +5,9 @@ authors = ["The Radicle Team <dev@radicle.xyz>"]
 edition = "2018"
 license = "GPL-3.0-or-later"
 
+[features]
+prop = [ "proptest" ]
+
 [dependencies]
 async-stream = "0.3"
 async-trait = "0.1"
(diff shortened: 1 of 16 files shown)


kim:
I like the RESkey stuff.

Buuut... I am rather certain that the feature gymnastics won’t behave as intended. The reason I’m saying this is that I tried this before with the keystore KDF params. In a workspace, cargo computes the largest set of features in order to not having to build crates multiple times. The net effect was that, no matter which target, the “special” feature always ended up being enabled.

As I said elsewhere: should we just stop fighting a broken build system, and move **all** tests and support code to a single workspace crate?

FintanH:
> As I said elsewhere: should we just stop fighting a broken build system, and move all tests and support code to a single workspace crate?

Let's DO IT! What do you think about `librad-test` being the candidate for moving the tests + support into since it already contains a lot of support code already?

kim:
> What do you think about librad-test

Ya, let's rename it to `test`, `radicle-link-test`, `asdfg`, or `malkovich`

PR #637, opened 2021-04-24
git: prune non-conforming references in `fetched_peers

### LINKED ISSUE #500: Pruning in fetched_peers
In https://github.com/radicle-dev/radicle-link/pull/478/files#diff-fe1d1a45efacfdf43a7b387a93ab0db0b6eeab3e68b613e5f190fcbcdafb6ff0R772-R799 we introduced parsing `fetched_peers` when we perform a shotgun fetch for a first time `replicate`. If there was an error in parsing then we `continue`, but this leaves the non-conforming reference lying around. It would be better to attempt to `prune` the reference.

therealyingtong:
Closes #500.

### DIFF
--- a/librad/src/git/replication.rs
+++ b/librad/src/git/replication.rs
@@ -350,12 +350,14 @@ where
     F::Error: std::error::Error + Send + Sync + 'static,
 {
     if !storage.has_urn(&urn)? {
-        let updated = fetcher
+        let mut updated = fetcher
             .fetch(fetch::Fetchspecs::PeekAll { limit })
             .map_err(|e| Error::Fetch(e.into()))?;
-        let fetched_peers = project::fetched_peers(&updated)?;
+        let (fetched_peers, pruned_tips) = project::fetched_peers(&updated)?;
+        updated.updated_tips = pruned_tips;
 
         let mut tips = updated.updated_tips;
+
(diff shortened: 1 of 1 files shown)


xla (commented):
utACK

⛱ 👐 🌼 🌃

FintanH:
Thanks for jumping in on this @therealyingtong. I'm not sure if this change might be necessary as we're going to revisit the replication. However, @kim might have a better idea about what the plan is there.

kim:
tbh, I stared at this a few times, but couldn't fully understand if it makes sense... so much for "needs to be revisited"

FintanH:
> tbh, I stared at this a few times, but couldn't fully understand if it makes sense... so much for "needs to be revisited"

:sweat_smile: 

So the issue the `FIXME` wants to prevent is that `PeekAll` fetches a wildcard for remotes, and we try and parse the peer in this call. However, it couldn't parse then we shouldn't really keep the reference around, and rather prune it. 

But actually, as usual, I jumped the gun and I'm realising this change doesn't necessarily do that. It just removes them from the updated tips.

I think `PeekAll` is more trouble than it's worth...

FintanH:
Thanks for the PR @therealyingtong, but I'm going to close this as we're revisiting the implementation of replication soon. Thanks for your time :bow:

PR #639, opened 2021-04-26
daemon: basic support for listing tagged patches

geigerzaehler:
We’re adding basic support for listing patches based on Git tags with a special prefix.

I’d love to get some feedback on how much of this can be extracted and moved to other parts of the codebase

### DIFF
--- a/daemon/src/patch.rs
+++ b/daemon/src/patch.rs
@@ -1,0 +1,169 @@
+//! [`list`] all the [`Patch`]es for project.
+
+use either::Either;
+use librad::{
+    git::types::{namespace::Namespace, Reference, RefsCategory},
+    peer::PeerId,
+    signer::BoxedSigner,
+};
+use radicle_git_ext::RefspecPattern;
+
+use crate::{
+    project::{self, peer},
+    state, Person,
+};
+
+const TAG_PREFIX: &str = "radicle-patch/";
+
+/// A patch is a change set that a user wants the maintainer to merge into a projects default
+/// branch.
+///
+/// A patch is represented by an annotated tag, prefixed with `radicle-patch/`.
+#[derive(Debug, Clone)]
+pub struct Patch {
+    /// ID of a patch. This is the portion of the tag name followin the `radicle-patch/` prefix.
+    pub id: String,
+    /// Peer that the patch originated from
+    pub peer: project::Peer<peer::Status<Person>>,
+    /// Message attached to the patch. This is the message of the annotated tag.
+    pub message: Option<String>,
+    /// Head commit that this
+    pub commit: git2::Oid,
+    /// The merge base of [`Patch::commit`] and the head commit of the first maintainers default
+    /// branch.
+    pub merge_base: Option<git2::Oid>,
+}
+
+impl Patch {
+    /// Returns true if [`Patch::commit`] is contained in the base branch history. This is the case
+    /// when [`Patch::commit`] is the same as [`Patch::merge_base`].
+    #[must_use]
+    pub fn merged(&self) -> bool {
+        self.merge_base == Some(self.commit)
+    }
+}
+
+/// List all patches for the given project.
+///
+/// # Errors
+/// * Cannot access the monorepo
+/// * Cannot find references within the monorepo
+pub async fn list(
+    peer: &crate::net::peer::Peer<BoxedSigner>,
+    project_urn: crate::Urn,
+) -> Result<Vec<Patch>, state::Error> {
+    let mut patches = Vec::new();
+    let monorepo_path = state::monorepo(peer);
+    let monorepo = git2::Repository::open(monorepo_path)?;
+    let namespace = Namespace::from(project_urn.clone());
+    let default_branch_head_commit = {
+        let project = state::get_project(peer, project_urn.clone())
+            .await?
+            .ok_or_else(|| state::Error::ProjectNotFound(project_urn.clone()))?;
+        let maintainer = project
+            .delegations()
+            .iter()
+            .flat_map(|either| match either {
+                Either::Left(pk) => Either::Left(std::iter::once(PeerId::from(*pk))),
+                Either::Right(indirect) => {
+                    Either::Right(indirect.delegations().iter().map(|pk| PeerId::from(*pk)))
+                },
+            })
+            .next()
+            .expect("missing delegation");
+        // the `remote` for `get_branch` is set to the first maintainer, if the current `peer`
+        // is that maintainer, `get_branch` will catch that and search the local peers directories.
+        // The `branch` is set to `None` as `get_branch` will then fall back to the default branch.
+        let default_branch =
+            state::get_branch(peer, project_urn.clone(), Some(maintainer), None).await?;
+        monorepo
+            .find_reference(&default_branch.to_string())?
+            .peel_to_commit()?
+            .id()
+    };
+
+    for project_peer in state::list_project_peers(peer, project_urn.clone()).await? {
+        let remote = match project_peer {
+            project::Peer::Local { .. } => None,
+            project::Peer::Remote { peer_id, .. } => Some(peer_id),
+        };
+        let ref_pattern = Reference {
+            remote,
+            category: RefsCategory::Tags,
+            name: format!("{}*", TAG_PREFIX)
+                .parse::<RefspecPattern>()
+                .expect("invalid refspec pattern"),
+            namespace: Some(namespace.clone()),
+        };
+        let refs = ref_pattern.references(&monorepo)?;
+        for r in refs {
+            let r = r?;
+            match patch_from_ref(
+                &monorepo,
+                default_branch_head_commit,
+                project_peer.clone(),
+                &r,
+            ) {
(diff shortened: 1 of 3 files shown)


daemon/src/patch.rs
```diff
@@ -0,0 +1,169 @@
...
+    signer::BoxedSigner,
+};
+use radicle_git_ext::RefspecPattern;
+
+use crate::{
+    project::{self, peer},
+    state, Person,
+};
+
+const TAG_PREFIX: &str = "radicle-patch/";
```
FintanH, line 16: Adding a `pub fn rad_patch` to `impl`s [here](https://github.com/radicle-dev/radicle-link/blob/master/librad/src/git/types/reference.rs#L160) would help with interacting with these. 

Bikeshedding moment: the path could be `refs/namespaces/<ns>/refs/rad/patches`

daemon/src/patch.rs
```diff
@@ -0,0 +1,169 @@
...
+
+/// A patch is a change set that a user wants the maintainer to merge into a projects default
+/// branch.
+///
+/// A patch is represented by an annotated tag, prefixed with `radicle-patch/`.
+#[derive(Debug, Clone)]
+pub struct Patch {
+    /// ID of a patch. This is the portion of the tag name followin the `radicle-patch/` prefix.
+    pub id: String,
```
FintanH, line 25: Maybe this could be the full `RefLike` path?

daemon/src/patch.rs
```diff
@@ -0,0 +1,169 @@
...
+    PeelToTagFailed(#[source] git2::Error),
+    #[error("failed to determine merge base")]
+    FailedToDetermineMergeBase(#[source] git2::Error),
+    #[error("tag target object is not a commit")]
+    InvalidObjectType,
+    #[error("tag name is not valid UTF-8")]
+    InvalidName(#[source] std::str::Utf8Error),
+}
+
+fn patch_from_ref<'repo>(
```
FintanH, line 129: I'm wondering if some of this could expressed in terms of `Storage` rather than using the raw `monorepo`

alexjg:
How does one "close" a patch in this model? We basically need the tag to be deleted from the patchers remote right?

daemon/src/patch.rs
geigerzaehler: We need the monorepo to call `monorepo.merge_base()` and only for that. `Storage` does not provide that functionality yet. We could add it to `Storage` or use `Storage::as_raw()`.

In general, would you prefer, if we follow the pattern established in `project::checkout` where all functions take an `impl CanOpenStorage + 'static` parameter to access the monorepo?

daemon/src/patch.rs
geigerzaehler: > Bikeshedding moment: the path could be `refs/namespaces/<ns>/refs/rad/patches`

If patches don’t live under the `tags` ref category then pushing will be slightly more complicated: `git push tags/rad/patches/*:rad/patches/*` instead of `git push --tags`. The upside is that we’re not messing with the regular `tags` category.

I have no strong preference so I’m happy to change this.

FintanH: Ah yes, this is a good point :) 

> `git push tags/rad/patches/*:rad/patches/*`

I wonder if we could extend the `include` machinery to do this for us :thinking:

daemon/src/patch.rs
FintanH: > In general, would you prefer, if we follow the pattern established in project::checkout where all functions take an impl CanOpenStorage + 'static parameter to access the monorepo?

Nope, that's not necessary because that's used for when interacting with remotes, which we're not doing here.

I think `as_raw` is available to any `librad::git` modules so that might be the best route to go down.

geigerzaehler:
> How does one "close" a patch in this model?

At the moment we mark a patch as closed if it has been merged into the maintainers default branch. This isn’t ideal of course and doesn’t covered squash or rebase merges. But it’s a start that should work well for a large part of the use cases and something that we can extend.

> We basically need the tag to be deleted from the patchers remote right?

As far as I know we don’t prune references when fetchting. This means that if a remote removes a reference that reference will not be removed from `remotes` in our monorepo.

alexjg:
I think it's fine to not implement these things as a first iteration, but closing, rebasing, squashing etc. are all very common workflows so it seems to me that we should have some idea of how we will achieve those things. That way we improve our chances of having a forward compatible protocol.

I can imagine that one might do something like have a version counter on the tag name, so it becomes something like `radicle-patch/<id>/version`, along with the remote ID this would give you a clock you can use for updating the patch. You could then store some state in the annotation message. Is this the kind of thing you had in mind?

It's completely possible I've missed discussion of all this which happened elsewhere, in which case please point me in the direction of said discussion :slightly_smiling_face: .

kim:
Is this the level of abstraction the daemon aims to provide? If so, I might have misunderstood something. All of the tag-based workflow can be implemented using surf or some other repository navigation tooling.

daemon/src/patch.rs
kim: `as_raw` is a crutch to avoid having to re-implement low-level git functionality for internal use. It is not intended to be part of the surface API.

geigerzaehler:
> Is this the level of abstraction the daemon aims to provide? If so, I might have misunderstood something. All of the tag-based workflow can be implemented using surf or some other repository navigation tooling.

That’s a good point. The daemon does indeed not seem to be quite the right place for this. But `radicle-surf` or `radicle-source` are not sufficient either. We need to get the tracked peer and the maintainers for a project. It should be possible to split the code here between `radicle-surf`, `radicle-source` and what is currently the internal `api` crate in Upstream. When I implemented this the general convention seemed to be to have this kind of logic in `coco` and for the `api` to be just a thin translation layer.

I’d be happy to try out splitting the logic. Do you or @FintanH and @xla have some concrete separation of the logic in mind. Of course this will also likely require us to integrate `radicle-source` into the proxy (or it’s upcoming incarnation).

> I think it's fine to not implement these things as a first iteration, but closing, rebasing, squashing etc. are all very common workflows so it seems to me that we should have some idea of how we will achieve those things. That way we improve our chances of having a forward compatible protocol.

Our basic idea for extending this is to start adding meta data to the tags (or objects) that represent patches. (Message trailers seem a good candidate.) For now, patches aren’t really a protocol since we just present tags in a certain way and we don’t have protocol specific data that requires interpretation (except for the patch prefix). This makes them automatically forwards compatible. Any future iteration will have some metadata (and a version number) that disambiguates new versions from the current version. 

This is also the reason I opted not to include the version number in the patch path. I might be missing something though. And if you feel that we should be more careful I’m happy to add the version number to the path.

daemon/src/patch.rs
geigerzaehler: > I wonder if we could extend the `include` machinery to do this for us thinking

Are you thinking about leveraging the [`remote.*.push` option](https://git-scm.com/docs/git-config#Documentation/git-config.txt-remoteltnamegtpush)? This might be dangerous because setting it messes with the users workflow. `git push rad` (without any refspecs) will use the configured value instead of pushing the current branch. Especially if we do this on checkout in the background users will not be aware.

kim:
> `radicle-source`

What's `radicle-source`?

> Do you or @FintanH and @xla have some concrete separation of the logic in mind.

I have at least as many questions around that as you :)

geigerzaehler:
> What's `radicle-source`?

It’s the source code browsing logic extracted from `coco` and [lives](https://github.com/radicle-dev/radicle-surf/tree/master/source) in the `radicle-surf` repo.

daemon/src/patch.rs
FintanH: Well currently for pushing branches we have:

```
[remote "rad"]
	url = rad://hnrkqyoojqyxtm8kcwokr6r4rfc3ey18xyeho.git
	fetch = +refs/heads/*:refs/remotes/rad/*
```

I thought it would have been possible to do the following:
```
[remote "rad"]
	url = rad://hnrkqyoojqyxtm8kcwokr6r4rfc3ey18xyeho.git
	fetch = +refs/heads/*:refs/remotes/rad/*
	fetch = +refs/rad/patches/*:refs/tags/*
```

However, after testing it, it just sends them to `refs/tags` in the monorepo regardles :/ Likely `git` Doing What I Mean :upside_down_face:

FintanH:
> We need to get the tracked peer and the maintainers for a project. It should be possible to split the code here between radicle-surf, radicle-source and what is currently the internal api crate in Upstream.

Ja, so I see what yis mean that this could be done in surf. @geigerzaehler I think this would mean that the Upstream client would need to combine calls:
1. List projects + peers (via daemon)
2. Get patches (via source+surf)

But surf won't know anything about projects or maintainers. That will just have to be opaque information and the client needs to contextualise the data, hence combining the values from 1. and 2. So the `Patch` type won't be as-is here but a restriction of that data, and the `Patch` type as-is here would be represented in the front-end types.

Does that make sense?

geigerzaehler:
> Does that make sense?

Yes, that makes sense. I’ll close this PR then since I don’t expect any changes to land in link

PR #641, opened 2021-04-26
net: announce after rere

kim:
The refs updated as a side-effect of rere shall be announced. This is a
prerequisite for grafting, which happens independent of gossip
membership.

Note that this will cause amplification while client-initiated announce
is still in effect.

### DIFF
--- a/librad/src/net/protocol/io/recv/git.rs
+++ b/librad/src/net/protocol/io/recv/git.rs
@@ -6,7 +6,7 @@
 use std::{io, net::SocketAddr};
 
 use futures::{
-    future::TryFutureExt as _,
+    future::{self, TryFutureExt as _},
     io::{AsyncRead, AsyncWrite},
     stream::{FuturesUnordered, StreamExt as _},
 };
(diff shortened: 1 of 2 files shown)


kim (commented):
Flipped with #650

xla (approved):
📀 📭 🍽 📺

librad/src/net/protocol/io/recv/git.rs
```diff
@@ -110,7 +112,30 @@ where
             .map(|ReplicateResult { updated_tips, .. }| updated_tips);
             match updated_tips {
                 None => tracing::info!("rere skipped"),
-                Some(xs) => tracing::info!("rere updated {} refs", xs.len()),
+                Some(xs) => {
+                    tracing::info!("rere updated {} refs", xs.len());
```
alexjg, line 116: Could we log the actual tips here?

kim: Hm, really? At what log level? There could be a lot.

alexjg: Yeah maybe at trace, mainly because this kind of information is really useful in ruling things out when debugging why things aren't replicating.

kim: How about instrumenting `control::gossip`, so it catches all broadcast messages, and the span path shows where it originated from?

alexjg: Yeah that would serve the same purpose I think. As long as we don't think it's likely we'll add additional conditionals before the gossip.

kim: Turns out that doesn't yield much useful output due to `Debug` being too, ehm, inspecting. Went with your suggestion instead.

FintanH (approved):

PR #642, opened 2021-04-28
replication: Account for missing refs prefix

xla:
When a peer announces without `refs` it would cause the logic to check
if the change is present locally to fail.

### DIFF
--- a/git-ext/src/reference/name.rs
+++ b/git-ext/src/reference/name.rs
@@ -403,6 +403,8 @@ impl From<RefLike> for Qualified {
     fn from(RefLike(path): RefLike) -> Self {
         if path.starts_with("refs/") {
             Self(path)
+        } else if path.starts_with("heads/") || path.starts_with("tags/") {
+            Self(format!("refs/{}", path))
         } else {
             Self(format!("refs/heads/{}", path))
         }
--- a/librad/src/net/peer/storage.rs
+++ b/librad/src/net/peer/storage.rs
@@ -283,6 +283,13 @@ mod tests {
         }
 
         #[test]
+        fn missing_refs() {
+            let urn = Urn::new(*ZERO_OID).with_path(reflike!("heads/main"));
+            let ctx = urn_context(*LOCAL_PEER_ID, Left(urn.clone()));
+            assert_eq!(urn.with_path(reflike!("refs/heads/main")), ctx)
+        }
+
+        #[test]
         fn remote_empty() {
             let urn = Urn::new(*ZERO_OID);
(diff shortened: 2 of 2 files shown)


FintanH (commented):
I think I have a more DWIM fix that I'm prepping. My worry is that this is too specific for these cases

kim:
Obviously begs the question: what generates refs where only the “refs/“
prefix is stripped?

The issue is: “pu” is equivalent to “refs/heads/pu”, yet “heads/pu” is
equivalent to “refs/heads/heads/pu”, which is clearly not the same. Iow,
“heads” is not a reserved word, and so the current behaviour is correct
(well, adheres to the git conventions).

We can obviously enforce that refs are always “fully qualified” on the
wire, but I have a feeling that this wouldn’t have prevented this issue,
because one can get confused easily about what format the git operation at
hand will yield (and thus how the ref may or may not need rewriting).

FintanH:
> The issue is: “pu” is equivalent to “refs/heads/pu”, yet “heads/pu” is
equivalent to “refs/heads/heads/pu”, which is clearly not the same. Iow,
“heads” is not a reserved word, and so the current behaviour is correct
(well, adheres to the git conventions).

Right, I was exploring if we could notice that `heads/pu` can be mapped to `refs/heads/pu`, but this requires us to reject `pu` because it isn't qualified with a category. Coming across [this test](https://github.com/radicle-dev/radicle-link/blob/master/librad/src/git/types/reference.rs#L642-L651) actually surprised me but it's clear now with your point above.

> We can obviously enforce that refs are always “fully qualified” on the
wire, but I have a feeling that this wouldn’t have prevented this issue,
because one can get confused easily about what format the git operation at
hand will yield (and thus how the ref may or may not need rewriting).

I think this would mean that `Urn` would take a `Qualified` path rather than `RefLike`, I think. This actually doesn't seem like the worst thing to me at first glance, but maybe there's a reason I'm not seeing where having an arbitrary `RefLike` is useful.

All that being said, the correct fix for the issue we saw is to qualify the path in upstream when announcing.

kim:
>
> I think this would mean that Urn would take a Qualified path rather than
> RefLike, I think.
>

Yeah, there is another subtlety, however: if we say Qualified means it’s
the directory hierarchy of the loose refs storage, then what about
namespaces?

I think we want a type there which removes “refs/namespaces/x/“ if it’s
there, but enforces that it starts with “refs/permissible-category/“. Which
is _not quite_ what the Reference type does.

Strings are hard, let’s work on caching instead.

>

PR #646, opened 2021-04-30
p2p: allow any in upload-pack

FintanH:
We allow `allowanysha1inwant`. This allows the server to request a
reference as a `want`, since one side might have updated its tips during
a mutual fetch.


xla (approved):
🕶 😉 👩 🌶

librad/src/git/p2p/server.rs
```diff
@@ -139,6 +139,8 @@ impl UploadPack {
 
         let mut git = Command::new("git");
         git.args(&["-c", "uploadpack.hiderefs=refs/"])
+            .arg("-c")
+            .arg("uploadpack.allowanysha1inwant=true")
```
kim, line 143: Not necessary here, as there are no wants being considered :)

kim (dismissed):

PR #651, opened 2021-05-05
Add a reason to request::states::Status::Failed

alexjg:
In addition to adding a reason to the request state enum we also modify
`librad::git::storage::fetcher::Fetcher::fetch` to return a more
descriptive error in the case of exceeded fetch limits.

Signed-off-by: Alex Good <alex@memoryandthought.me>


librad/src/git/storage/fetcher.rs
```diff
@@ -262,6 +262,20 @@ pub mod error {
         #[error(transparent)]
         Pool(#[from] super::PoolError),
     }
+
+    #[derive(Debug, Error)]
+    pub enum FetchError {
+        #[error("Fetch limit {limit} exceeded. Amount fetched: {amount_fetched} from {remote}. Fetch specs: {fetch_specs:?}, refspecs: {refspecs:?}")]
```
kim, line 268: I'm not sure it is a good default to dump the specs in the `Display` impl -- these can be very large, and most likely only provide value when at least `DEBUG` is enabled. Since the `Debug` impl will amount to the same information, maybe leave it to the call site to log it out at the appropriate level?

librad/src/git/storage/fetcher.rs
```diff
@@ -416,7 +430,13 @@ mod imp {
             urn: Urn,
             remote_peer: PeerId,
         ) -> Result<Self, git2::Error> {
-            let mut remote = storage.as_raw().remote_anonymous(url.as_str())?;
+            let mut remote = match storage.as_raw().remote_anonymous(url.as_str()) {
+                Ok(r) => r,
+                Err(e) => {
+                    tracing::error!(?url, error=?e, "Error opening remote");
```
kim, line 436: nit: `err=?e` is used elsewhere. If/when we decide to employ some kind of log indexing or trace collector, it might come in handy to be able to query for that field.

librad/src/git/storage/fetcher.rs
alexjg: Good shout, fix pushed

librad/src/git/storage/fetcher.rs
alexjg: Fix pushed

daemon/Cargo.toml
```diff
@@ -6,6 +6,7 @@ authors = ["Radicle Team <dev@radicle.xyz>"]
 edition = "2018"
 
 [dependencies]
+anyhow = "1.0.40"
```
FintanH, line 9: Does it have to be specific to the patch version? Or could we use `"1"`?

librad/src/git/storage/fetcher.rs
```diff
@@ -262,6 +262,20 @@ pub mod error {
         #[error(transparent)]
         Pool(#[from] super::PoolError),
     }
+
+    #[derive(Debug, Error)]
+    pub enum FetchError {
+        #[error("Fetch limit {limit} exceeded. Amount fetched: {amount_fetched} from {remote}")]
```
FintanH, line 268: We try and follow these rules for displaying errors https://github.com/radicle-dev/radicle-decisions/blob/master/proposals/0002.md#the-behavior-of-display
```suggestion
        #[error("fetch limit {limit} exceeded, fetched {amount_fetched} from {remote}")]
```

daemon/src/request/states.rs
```diff
@@ -80,15 +80,18 @@ pub struct Requested {
...
 pub enum Status {
     /// The status of the related data is: ready to go!
     Available,
     /// The status of the related data is: in progress!
     InProgress,
     /// The status of the related data is: failed :(
-    Failed,
+    Failed {
+        #[allow(missing_docs)]
+        reason: String,
```
FintanH, line 93: Would it be better to use `Box<dyn Error>` here too?

daemon/src/request/states.rs
```diff
@@ -102,13 +105,24 @@ impl Status {
     /// progress. And finally, the last case is that both `Status`es agree
     /// the `Status` is `Available`.
     #[must_use]
-    pub const fn join(self, other: Self) -> Self {
-        match (self, other) {
-            (Self::Failed, _) | (_, Self::Failed) => Self::Failed,
+    pub fn join(&self, other: &Self) -> Self {
+        match (self, &other) {
+            (Self::Failed { reason: reason1 }, Self::Failed { reason: reason2 }) => Self::Failed {
+                reason: format!("{} and {}", reason1, reason2),
```
FintanH, line 111: Ah, but it might be more complicated here. Can we chain reasons?

FintanH (commented):
I like where this is going :)

daemon/Cargo.toml
kim: @FintanH I had misunderstood how this works in the cargo world. This being "^1.0.40", and cargo being mucho semantico, the meaning is ">= 1.0.40 && <2.0.0". If the first digit was a zero, it would mean ">= 0.0.40 && <0.0.41".

kim: (please do not ask "what about the digit in the middle?")

FintanH: Dios mio :scream: I think you're right. `1.0.40` includes the hidden `^` and this is the table they list:
```
^1.2.3  :=  >=1.2.3, <2.0.0
^1.2    :=  >=1.2.0, <2.0.0
^1      :=  >=1.0.0, <2.0.0
^0.2.3  :=  >=0.2.3, <0.3.0
^0.2    :=  >=0.2.0, <0.3.0
^0.0.3  :=  >=0.0.3, <0.0.4
^0.0    :=  >=0.0.0, <0.1.0
^0      :=  >=0.0.0, <1.0.0
```

I really don't understand this... So _any_ non-backward compatible change in `>=1` has to be a major change rather than minor.

So this is saying that we need changes from `1.0.40` and above but nothing below. Is that the expectation @alexjg?

FintanH: > I really don't understand this... So any non-backward compatible change in >=1 has to be a major change rather than minor.

Ah wait, I've confused myself...

kim: No, no, that's correct, as you can convince yourself of here: https://semver.org/lang/ar/#%D8%A8%D8%A7%D8%AE%D8%AA%D8%B5%D8%A7%D8%B1

Haskellers have the luxury of an additional digit in the version specifier, but it's not clear if that reduces the total amount of confusion

alexjg: I've changed it to `"1.0.0"`. There is an argument that maybe we shouldn't depend on `anyhow` at all as it's generally intended for applications rather than libraries. But `daemon` is more like an application anyway so I think it's okay.

daemon/src/request/states.rs
alexjg: That would require us to have a custom `Serialize` implementation I think, which to me seemed like it wasn't worth it given that this `reason` field is for diagnostic purposes anyway rather than for creating new errors from.

daemon/Cargo.toml
xla: FWIW the rust "error community" considers `anyhow` deprecated and advises alternatives like `eyre`.

librad/src/git/storage/fetcher.rs
alexjg: Fixed

daemon/Cargo.toml
kim: <sup>[citation needed]</sup>

librad/src/git/storage/fetcher.rs
```diff
@@ -262,6 +262,20 @@ pub mod error {
...
+
+    #[derive(Debug, Error)]
+    pub enum FetchError {
+        #[error("Fetch limit {limit} exceeded, fetched: {amount_fetched} from {remote}")]
+        FetchLimitExceeded {
+            limit: usize,
+            amount_fetched: usize,
+            remote: PeerId,
+            fetch_specs: Fetchspecs<PeerId, Revision>,
+            refspecs: Vec<String>,
```
FintanH, line 274: Do we need `fetch_specs` _and_ `refspecs` out of curiosity? It seems `refspecs` can be derived from `fetch_specs`

librad/src/git/storage/fetcher.rs
```diff
@@ -262,6 +262,20 @@ pub mod error {
...
     }
+
+    #[derive(Debug, Error)]
+    pub enum FetchError {
+        #[error("Fetch limit {limit} exceeded, fetched: {amount_fetched} from {remote}")]
+        FetchLimitExceeded {
+            limit: usize,
+            amount_fetched: usize,
+            remote: PeerId,
+            fetch_specs: Fetchspecs<PeerId, Revision>,
```
FintanH, line 273:
```suggestion
            fetchspecs: Fetchspecs<PeerId, Revision>,
```
nit: To keep in line with the naming elsewhere :)

FintanH (approved):

daemon/Cargo.toml
xla: I sufficiently bamboozled myself and mistook the `anyhow` crate for the `anomaly` crate - which is being sunset. In the process of digging for citation I found this: https://github.com/rust-lang/project-error-handling

librad/src/git/storage/fetcher.rs
alexjg: Fixed

daemon/src/request/states.rs
alexjg: I'm not sure I know what you mean? Something like elaborate the type so that it can have multiple reason fields rather than concatenating strings?

librad/src/git/storage/fetcher.rs
alexjg: My inclination is to have both. Yes the `refspecs` are derived from the `fetchspecs`, but the logic is non-trivial. In my experience when things have gone wrong (which is when you would be looking at an error) you want to expose as much as possible of the intermediate state of the program so that it's easier to rule out hypotheses about which part of the program have misbehaved.

daemon/src/request/states.rs
FintanH: Sorry, this was in relation to this comment https://github.com/radicle-dev/radicle-link/pull/651#discussion_r626638113. So this can be resolved :)

librad/src/git/storage/fetcher.rs
FintanH: Sounds reasonable to me :+1:

FintanH (approved):

kim (approved):

PR #652, opened 2021-05-05
net: run rere after serving the fetch

kim:
Running both concurrently makes things harder to reason about. Better
opportunities for optimisation and interleaving may arise with git
protocol v2.

### DIFF
--- a/librad/src/net/protocol/io/recv/git.rs
+++ b/librad/src/net/protocol/io/recv/git.rs
@@ -11,11 +11,9 @@ use futures::{
     stream::{FuturesUnordered, StreamExt as _},
 };
 use thiserror::Error;
-use tracing::Instrument as _;
 
 use crate::{
-    executor,
-    git::{p2p::header::Header, replication::ReplicateResult, Urn},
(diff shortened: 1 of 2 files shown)


xla (approved):
Bikeshed: debug could be more appropriate for the rere logs.

ℹ️ 🚑 👊 ⭐️

FintanH (approved):
This will also limit the case where we see history rewrites since `rere` won't go before the replication :)

xla:
> This will also limit the case where we see history rewrites since `rere` won't go before the replication :)

Does that not still depend on respecting the force flag as addressed in libgit2/libgit2#5854

FintanH:
> Does that not still depend on respecting the force flag as addressed in libgit2/libgit2#5854

Not necessarily. Emphasis on the word _limit_. We were running into the issue because the seed would `rere` _before_ `replicate` and we would witness the rewrite. Since we're enforcing `replicate` goes before `rere` here it means that the history will be preserved. But it doesn't rule out _all_ history rewrites, of course.

FintanH (approved):

PR #655, opened 2021-05-06
daemon: run all async tests on the multi-threaded runtime

kim:
In conjunction with #650, we see async tests intermittently hanging
indefinitely when the per-test runtime shuts down. This may have to do
with some blocking operations in `daemon` not being properly
`spawn_blocking`-ed (and thus blocking the single runtime thread), or
some mysterious interaction with threads spawned from within the async
program.

In any case, using the multi-threaded runtime appears to fix it.

### DIFF
--- a/daemon/src/seed.rs
+++ b/daemon/src/seed.rs
@@ -94,7 +94,7 @@ mod tests {
 
     use pretty_assertions::assert_eq;
 
-    #[tokio::test]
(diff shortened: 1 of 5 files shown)


xla (approved):
📞 📼 🕞 🦁

FintanH (approved):
`flavor` :yum:

PR #656, opened 2021-05-07
net: send disconnect on unsolicited broadcast

kim:
Another oversight when switching to unidirectional streams: the
unsolicited state occurs when partial view states diverge, but the
remote end will never know because closing the stream has no effect
anymore. Thus, send a DISCONNECT message in this case to sync states.

### DIFF
--- a/librad/src/net/protocol/io/recv/gossip.rs
+++ b/librad/src/net/protocol/io/recv/gossip.rs
@@ -11,19 +11,23 @@ use futures::{
 };
 use futures_codec::FramedRead;
 
-use crate::net::{
-    connection::RemotePeer,
-    protocol::{
-        broadcast,
-        gossip,
-        info::PeerInfo,
-        io::{codec, peer_advertisement},
-        membership,
-        tick,
-        ProtocolStorage,
-        State,
+use crate::{
+    net::{
+        connection::RemotePeer,
+        protocol::{
+            broadcast,
+            event,
+            gossip,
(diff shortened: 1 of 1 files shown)


librad/src/net/protocol/io/recv/gossip.rs
```diff
@@ -68,22 +67,68 @@ pub(in crate::net::protocol) async fn gossip<S, T>(
                 )
                 .await
                 {
-                    Err(e) => {
-                        tracing::warn!(err = ?e, "gossip error");
+                    // Partial view states diverge apparently, and the stream is
+                    // (assumed to be) unidirectional. Thus, send a DISCONNECT
+                    // to sync states.
+                    Err(broadcast::Error::Unsolicited { remote_id, .. }) => {
+                        tracing::warn!(
```
alexjg, line 74: Is this a `warn` or is this an expected part of the protocol? i.e do we want to scare operators with these messages?

kim: It can happen because:

1. The remote end went crazy somehow
2. There are broadcasts in-flight on the remote end while it notices that the local end is no longer in the active set
3. Something went wrong, and both ends are in some kind of split-brain state

So, I would think it's a piece of information one would like to know, and should def not be below info.

xla (approved):
🔬 🖐 🏧 🌾

alexjg (dismissed):

PR #661, opened 2021-05-10
seed: Improve resilience

xla:
See single commits for rationale and context.

### DIFF
--- a/seed/src/handle.rs
+++ b/seed/src/handle.rs
@@ -3,8 +3,13 @@
 // This file is part of radicle-link, distributed under the GPLv3 with Radicle
 // Linking Exception. For full terms see the included LICENSE file.
 
-use futures::{channel::mpsc as chan, sink::SinkExt as _, stream::StreamExt as _};
+use std::time::Duration;
+
 use thiserror::Error;
+use tokio::{
+    sync::{mpsc, oneshot},
+    time,
+};
 
 use librad::{net::protocol::event::downstream::MembershipInfo, peer::PeerId};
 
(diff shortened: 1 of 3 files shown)


seed/src/lib.rs
```diff
@@ -135,15 +137,15 @@ impl Default for NodeConfig {
...
     /// Receiver end of user requests.
-    requests: chan::UnboundedReceiver<Request>,
+    requests: mpsc::Receiver<Request>,
 }
 
 impl Node {
     /// Create a new seed node.
     pub fn new() -> Result<Self, Error> {
-        let (handle, requests) = chan::unbounded::<Request>();
+        let (handle, requests) = mpsc::channel::<Request>(64);
```
alexjg, line 148: Is there any particular logic to choosing this size for the buffer?

seed/src/lib.rs
```diff
@@ -222,28 +226,27 @@ impl Node {
     /// Handle user requests.
     async fn handle_request(request: Request, api: &Peer<Signer>) -> Result<(), Error> {
         match request {
-            Request::GetMembership(mut reply) => {
+            Request::GetMembership(reply) => {
                 let info = api.membership().await;
-                reply.send(info).await?;
+                reply.send(info).map_err(|_| Error::Reply)
```
alexjg, line 231: Does this need the closure?

seed/src/lib.rs
xla: Power of 2, as meaningful as 42. Magic constant until a better heuristic has been found or it is surfaced as config param.

seed/src/lib.rs
alexjg: Ah, my bad, I misread the code and thought it was passing argument to `Error::Reply`, not enough coffee!,

alexjg (approved):

seed/src/lib.rs
FintanH, line 148: Should we parameterise this number?

seed/src/lib.rs
```diff
@@ -70,11 +72,11 @@ pub enum Error {
     #[error(transparent)]
     Node(#[from] NodeError),
 
-    #[error(transparent)]
-    Channel(#[from] chan::SendError),
-
     #[error(transparent)]
     Profile(#[from] profile::Error),
+
+    #[error("sending reply failed")]
```
FintanH, line 78: It would useful to say what the reply was for

FintanH (changes requested):
Looks good :ok_hand:

FintanH (dismissed):
Meant to approve :sweat_smile:

seed/src/lib.rs
xla: Maybe, what is the knob called? How is it exposed and contextualise for an operator? What would be a sanerisher default than 64?

FintanH: It represents the number of tasks that the `Node` can handle right? So it's kind of like a backpressure knob or something along those lines. 

> How is it exposed and contextualise for an operator?

How did you contextualise the need for it? :)

> What would be a sanerisher default than 64?

I would think that it would be some function of:
```
#expected_requests_per_second / #average_seconds_to_complete_request
```

seed/src/handle.rs
```diff
@@ -14,59 +19,71 @@ use crate::Project;
...
+        let (tx, rx) = oneshot::channel();
         self.channel
-            .send(Request::GetMembership(tx))
-            .await
-            .map_err(|_| NodeError::RequestFailed)?;
+            .try_send(Request::GetMembership(tx))
+            .map_err(|_| NodeError::RequestSendFailed)?;
 
-        rx.next().await.ok_or(NodeError::RequestFailed)
+        time::timeout(Duration::from_secs(3), rx)
```
cloudhead, line 51: Shouldn't the caller decide on the timeout?

xla: That's an interesting alternative, it would complicate the API and I'd assume we want a default so we would need to expose an `Option` param or have a way for the NodeHandle to be configured. I'd be curious how the caller decides what the right value is. 3 is a pretty solid number.

seed/src/lib.rs
xla: > How did you contextualise the need for it? :)

It's to address an unbounded queue, as contextualised in the commit message we've seen requests just pilling up because of polling nature of the UI when the tincans and machinery in the background was stuck, to then just explode with a release of all resources. It's one of the efforts to make the seed operation more resilient.

> I would think that it would be some function of:

I don't think either of those numbers are known and/or depend heavily on the local setup. While I think it's right to have a knob, it should mature more before we know what the knob does exactly and how it can be tuned. The surface of the seed is already expanding and I'd like to avoid a scenario where you need experts to tune the params right (JVM, aehm).

FintanH: > I don't think either of those numbers are known and/or depend heavily on the local setup. While I think it's right to have a knob, it should mature more before we know what the knob does exactly and how it can be tuned.

How can it mature without being easily tuned from the outside? The only way to change the value now would be to change this line of code, re-compile, and start the executable again. But obviously, we can start off with the "sensible" default of 64.

I don't feel too strongly about this though :)

seed/src/handle.rs
cloudhead: Yeah, either it's set on the handle, eg `handle.set_timeout(x)`, or you have two functions, one which takes a timeout, eg. `foo` and `foo_timeout`, or also just force the caller to specify a timeout, since it's a low level library function..

Just throwing some other options, though 3 is a good number.

cloudhead: The thing I don't like about using an `Option` is when you read `handle.get_memberships(None)` it's like wut.

FintanH (approved):
Thanks for thinking about the magic numbers :bow:

cloudhead (dismissed):

FintanH (approved):

PR #662, opened 2021-05-10
wip: RFC: Collaborative Objects

alexjg:
This is a very  rough, incomplete, first draft of an RFC for collaborative objects. Details on exactly what the API will look like, and discussions of downsides and alternatives are still missing.


docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,215 @@
...
+expensive and there are opportunities to do some indexing ahead of time. A more
+pressing concern is that we want to ensure that the schema is respected, more
+on this later.
+
+
+### Updating objects
+
+To make a change to an object we load the existing messages for an object. The 
+application developer provide us with the binary representation of the change
+to that object. We apply the change and ensure that the new object state still
```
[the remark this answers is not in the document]
alexjg: Unfortunately not, the change the application developer would provide is (from our perspective) just some bytes. It has to be applied to a document to mean anything. This does open some attack vectors via creating large changes which take a long time to apply.

In this model checking the schema before applying the change would be the responsibility of the CRDT implementation. I'm not aware of any implementation which does this right now, although it is on a list of future work for Automerge so it would be possible to implement it in this manner in the future (and without any protocol changes to boot).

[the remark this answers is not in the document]
FintanH: Could you maybe give me an example where a change comes in and modifies an object but invalidates the schema? I think that would help me understand this issue a bit better.

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,215 @@
...
+matches the object schema. At this point the state of the object may depend on
+many contributions from the tracking graph - not just the ones in our own view
+of the project. This is a natural point at which to compact the document using
+automerge's compacted document format. We bundle all the changes into one
+compacted change file and replace all the changes in our view of the object
+with this compacted change file - then create a new commit, update our signed
+refs, and announce.
```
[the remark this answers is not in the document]
alexjg: In the Automerge model the computations _are_ the object. In general with CRDTs you are actually passing the history of the whole document around, this is how we're able to do commutative merges.

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,215 @@
...
+
+### Strange Perspectives
+
+This model introduces some counter-intuitive properties. For example, I might
+"create an issue" in a repository and anyone who is tracking me would see that
+issue, but people who are tracking the project but don't have me in their
+tracking graph will only see the issue if the maintainer replies to it. It's
```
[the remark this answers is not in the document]
alexjg: Yup, exactly.

docs/rfc/0003-collaborative-objects.md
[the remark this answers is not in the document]
alexjg: Imagine we have a schema like the following:

```json
{
    "type": "object",
    "properties": {
        "title": {
            "type": "string",
            "maxLength": 100
        }
    }
}
```

And the current state of the document in our view of the repo is:

```json
{
    "title": "A big imporatnt issue"
}
```

In this case let's say our object directory looks like this:

```
.
|--<schema>
|  |--<schema change>
|--<change 1>
|--<manifest>
```

i.e there is a single binary change file which when loaded resolves to the "current state" I mentioned above.

From `librad`s perspective what we do is this:

- Load the schema changes using automerge to resolve the schema. Effectively we have `schema = Automerge.load(<schema change bytes>)`
- Load the existing state of the object using automerge, `object = Automerge.load(<change 1 bytes>)`
- Pass the automerge document to the application, which has it's way with it and returns a new change, let's call it `<change 2 bytes>`
- The application passes the change back to us, we apply it to the the document so we have `updatedObject = Automerge.apply(object, <change 2 bytes>)`
- Now we check that `updatedObject` is valid with respect to the schema
    - If it is, we compact all the changes and save them in the object directory
    - If it is not then we return an error to the application

So if the application were to return a change which set the title of the object to a string longer than 100 characters then we would return an error.

docs/rfc/0003-collaborative-objects.md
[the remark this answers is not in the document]
FintanH: Any references to get more up-to-speed with this concept?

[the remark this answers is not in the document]
alexjg: Bartosz Sypytkowski (who also has an excellent post about HyParView) has a blog series on CRDTs which starts at https://bartoszsypytkowski.com/the-state-of-a-state-based-crdts/

For more academic work there's an overview of CRDTs (as of 2011, lot's has changed since then) here https://hal.inria.fr/file/index/docid/555588/filename/techreport.pdf

There's also a paper on Automerge specifically at https://arxiv.org/pdf/1608.03960.pdf. The algorithm has changed somewhat but it's still a decent overview.

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,215 @@
...
+## Message storage and transmission
+
+The property of CRDTs that we care about is strong eventual consistency: any
+two nodes which have received the same set of messages will merge to the same
+state. A major part of the design of collaborative objects therefore is how we
+store and transmit these messages. Messages in this context are just binary
+blobs which need to be passed to the CRDT implementation in any order.
+
+We add a new entry to the `signed_refs` of a project at
+`refs/namespaces/<project>/remotes/refs/remotes/<peer>/rad/collaborative-objects`.
```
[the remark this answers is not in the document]
alexjg: Yeah hashing the schema is the approach taken by the Avro Schema Registry. It works very nicely if you don't have forks in the schema evolution. In a distributed setting it's harder to see how this would work, although if we only allowed schemas to be updated via the `ProjectPayload` then it might be possible.

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Message storage and transmission
+
+The property of CRDTs that we care about is strong eventual consistency: any
+two nodes which have received the same set of messages will merge to the same
+state. A major part of the design of collaborative objects therefore is how we
+store and transmit these messages. Messages in this context are just binary
+blobs which need to be passed to the CRDT implementation in any order.
+
+We add a new entry to the `signed_refs` of a project at
```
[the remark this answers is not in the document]
alexjg: Yeah good point, anything under `refs/namespaces/<project>/remotes/refs/remotes/<peer>/rad/*` is already signed by `signed_refs` right?

docs/rfc/0003-collaborative-objects.md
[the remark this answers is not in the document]
FintanH: It might be useful for easily detecting when you're on the same schema. Could be interesting to think about it via a merkle-chain where you could detect which side is in the past, or as you said if it's a fork. Purely spit-balling here :)

kim:
@alexjg Do you want feedback on this already?

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,421 @@
...
+refs, and announce.
+
+### Strange Perspectives
+
+This model introduces some counter-intuitive properties. For example, I might
+"create an issue" in a repository and anyone who is tracking me would see that
+issue, but people who are tracking the project but don't have me in their
+tracking graph will only see the issue if the maintainer replies to it. It's
+hard to see how you would do things like "link to an issue" under these
+constraints. This is inherent to the network model though, rather than being a
```
kim, line 345: As you say, it's inherent. Consider, however, the role of seed nodes in our model: modulo their local access policy, they act as supernodes for the social network. Ie. a libral seed (which auto-tracks every peer id it learns about) would converge to a globally consistent view.

This poses two problems:

1. One might wish to _interrogate_ a seed's view of the object, but not fetch 10k remotes to do so
2. A maintainer might wish to apply _some_ of the updates the seed has applied, but leave out others (e.g because they contain spam)

This seems to go against the proposal to store _all_ objects in a single `tree` object, even though tempting due to the consistency guarantees. Not that it is _impossible_ to perform "shallow" fetches, but it'd be considerably more complicated. Also consider that we can't tell what was included in a partial merge other than evaluating the automerge, unlike actual merge (multi-parent) commits.

docs/rfc/0003-collaborative-objects.md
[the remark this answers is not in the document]
FintanH: Ach, whoops! My bad - thanks for the correction

docs/rfc/0003-collaborative-objects.md
alexjg: Yeah ibagree with these concerns, im working on what this will look like for a tree-per-object, possibly with a path element in the ref for the type name of the object. This would allow things like fetching just the state of one object, or the state of all objects of one type.

it does increase the size of the initial ref advertisment during upload pack

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,421 @@
...
+        "type": "string"
+    },
+    "description": {
+        "type": "string"
+    },
+    "author": {
+        "type": "string",
+        "description": "The radicle ID of the author of the issue"
+    },
+    "signature": {
```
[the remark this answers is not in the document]
alexjg: In this simple schema you could sign over a canonical JSON encoding of the "title", "description", "author", and "id". To make this more realistic we could add an ID field to each comment and the sign over ("comment ID", "comment text", "parent_id"), where "parent_id" is the ID of the previous comment according to the author at the time they wrote it. Validating such a signature would involve asserting that the previous comment appears before the given comment in the document.

If maintaining a stricter invariant than precedence is necessary then you could encode the comments as a tree structure rather than a list.

There are obviously a whole bunch of other problems - specifying what the signature algorithm is, what key was used etc. Hopefully these can all be concerns of the particular data type implementation and the collaborative objects infrastructure is agnostic about them. One thing I have thought about that might be useful would be to add a custom schema type which verifies that a signature is a valid signature with respect to another URN in the schema. Maybe something like this:

```json
{
    "signature": {
        "type": "rad_signature",
        "signing_urn_path": "/author",
        "over": [
            "id",
            "title",
            "description",
            "author",
        ]
        "description": "A base64 encoded signature of the issue"
    },
}
```

I feel like in the process of trying to implement the issues feature we will encounter problems which will help flesh out what such features should look like.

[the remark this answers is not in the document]
FintanH: > In this simple schema you could sign over a canonical JSON encoding of the "title", "description", "author", and "id". To make this more realistic we could add an ID field to each comment and the sign over ("comment ID", "comment text", "parent_id"), where "parent_id" is the ID of the previous comment according to the author at the time they wrote it. Validating such a signature would involve asserting that the previous comment appears before the given comment in the document.

To me, this seems like responsibilities are being mixed. I understand that having such a signature would be useful but I would imagine it being placed somewhere _above_ the payload, e.g. a git trailer. And so the signature is always over the payload itself -- where a reply to a particular comment could be in the schema itself.

I'm also wondering if that would be covered by the `rad/signed_refs` (after my comment blunder :))?

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,421 @@
...
+        "type": "string",
+        "description": "A base64 encoded signature of the issue"
+    },
+    "comments": {
+        "type": "array",
+        "items": {
+            "type": "object",
+            "properties": {
+                "text": {"type": "string"},
+                "author": {
```
FintanH, line 90: I'm curious if there'll be ways to define common components, for example `author` is repeated here. Or will we have to reach for something like Dhall to do that? :)

docs/rfc/0003-collaborative-objects.md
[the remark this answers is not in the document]
kim: Yes, I also feel like this would put too much burden on the application layer. I read the proposal as we never actually store the JSON representation, only opaque automerge changes, so it's not so easy to see how to attribute a particular change to an author -- but otoh, an author can change everything about the object, including authorship fields.

I think this can probably be solved if we're not using `tree`s, but the actual git history. So I'll wait for that and inquire again later =]

> covered by the rad/signed_refs

Nope, that has a different purpose. It attests the _origin_ of a set of branches, not authorship.

docs/rfc/0003-collaborative-objects.md
kim: > im working on what this will look like for a tree-per-object

cool, I'll ask again then

> it does increase the size of the initial ref advertisment during upload pack

ya, we'll have some scalability issues with that either way. git protocol v2 will help (`ref-prefix`), but I'm thinking that we may want to prefix-compress refs on the wire at some point.

alexjg: Alternatively, and what I have written in a draft at the moment, we could add a query parameter to radicle URNs which specifies what collaborative objects you want to fetch. If that query parameter is omitted then no refs for them are advertised. Obviously this is a bit of a hack - but given that collaborative objects are Radicle specific things anyway that feels acceptable.

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,421 @@
...
+|  |  |--<schema change 1>
+|  |  |--<schema change 2>
+|  |--<change 1>
+|  |--<change 2>
+|--<object 2 ID>
+|  ...
+```
+
+We refer to a directory in this tree as an "object directory" and the tuple 
+`(manifest, [schema changes], [changes])` as an "object state".
```
FintanH, line 296: So I'm not clear on what the storage is across peers. Taking the example of the link team collaborating we would have a the following structure in our monorepo:

```
├── namespaces
│   ├── hnrkc3expus8az3s5iehfyi5qegyh3841nwxo
│   │   └── refs
│   │       ├── rad
│   │       │   ├── id
│   │       │   ├── ids
│   │       │   │   ├── alex-person
│   │       │   │   ├── finto-person
│   │       │   │   ├── kim-person
│   │       │   │   └── xla-person
│   │       │   ├── self
│   │       │   └── signed_refs
│   │       └── remotes
│   │           ├── alex-peer
│   │           │   ├── heads
│   │           │   │   └── main
│   │           │   └── rad
│   │           │       ├── id
│   │           │       ├── ids
│   │           │       │   ├── alex-person
│   │           │       │   ├── finto-person
│   │           │       │   ├── kim-person
│   │           │       │   └── xla-person
│   │           │       └── signed_refs
│   │           ├── kim-peer
│   │           │   ├── heads
│   │           │   │   └── main
│   │           │   └── rad
│   │           │       ├── id
│   │           │       ├── ids
│   │           │       │   ├── alex-person
│   │           │       │   ├── finto-person
│   │           │       │   ├── kim-person
│   │           │       │   └── xla-person
│   │           │       └── signed_refs
│   │           └── xla-peer
│   │               ├── heads
│   │               │   └── main
│   │               └── rad
│   │                   ├── id
│   │                   ├── ids
│   │                   │   ├── alex-person
│   │                   │   ├── finto-person
│   │                   │   ├── kim-person
│   │                   │   └── xla-person
│   │                   ├── self
│   │                   └── signed_refs
```

Now we're adding a new reference `rad/collaborative_objects`. How would that be split between my top-level `rad` tree and `alex-peer`, `kim-peer`, and `xla-peer`? Would we maintain a consistent view of all the changes we're aware of or just our own changes? If the latter, do we store the computed document? In the current iteration, I don't see where that would fit in `(manifest, [schema changes], [changes])`.

alexjg: I've pushed a change which slightly modifies the layout, but this question is still basically the same. My current thinking is that the changes each remote is aware of stay in their respective remotes - we don't store the computed document in the repository (though we may want to cache it somewhere). When we modify an object then we copy all the changes that the modified object depends on into our view of that object and announce it.

FintanH: > we don't store the computed document in the repository (though we may want to cache it somewhere).

Ah, I was thinking automerge takes the whole document, but looked back and saw it loads the document from a set of changes. I suppose applications could decide to cache the computed document.

> I've pushed a change which slightly modifies the layout, but this question is still basically the same. My current thinking is that the changes each remote is aware of stay in their respective remotes

Something that's brewing in my head is what happens when there's a number of hops to something like a reply. Say you created an issue that an Ink-n-Switch collaborator replies to and their colleague replies to that. Which one of those replies would _I_ see and how would those changes reach me?

alexjg: I _think_ this is what I address in the "strange perspectives" section. It's effectively the same model as for receiving changes in the code in a repository (this is a deliberate choice, hopefully we only have to educate users about it once that way). The problems and solutions will likely be the same - things like seed nodes which listen for changes you might be "interested" in but which are not part of your tracking graph and notify you of them.

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,506 @@
...
+
+### Updating objects
+
+To make a change to an object we load the existing messages for an object. The 
+application developer provides us with the binary representation of the change
+to that object. We apply the change and ensure that the new object state still
+matches the object schema. At this point the state of the object may depend on
+many contributions from the tracking graph - not just the ones in our own view
+of the project. We now create a commit with our new change in it, referencing
+all the commits containing the direct dependencies of the change as parents.
```
FintanH, line 425: Now when you say "all the commits containing the direct dependencies", does that mean we can "reference" other peer's commits? For example, is replying to a comment adding that comment's respective commit to the new commit as a parent?

alexjg: Yep, exactly.

FintanH: Cool, so there is a way that I might see your comment if I'm not tracking you nor the maintainer replies, since someone I'm tracking might reply to you :)

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,649 @@
...
+- `history_type`, always `"automerge"`, this is here to allow for different
+  CRDT implementations in future.
+
+
+Each object is also created with a JSON schema. The schema is represented by an
+initial `schema.json` and a series of schema migrations which extend that
+initial schema. Schema migrations will not be addressed in detail in this RFC
+but we will show their feasibility.
+
+#### `schema`
```
FintanH, line 334: I think it would be good to cover what can go in a `schema` by either listing them here, or linking to other documentation, or both

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,649 @@
...
+#### `change`
+
+This is the automerge change which this commit introduces. It is a binary file
+which must contain a single change and it's dependents must be the dependents
+referenced by the parents of the commit.
+
+
+#### `author`
+
+This is a file containing the SHA of the commit which references the authors
```
FintanH, line 356: I'm wondering if a symbolic ref might be a better choice here, similar to how `rad/ids/<hash>` and `rad/self` are simply symbolic refs to stay up-to-date

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,649 @@
...
+identity. We need this so we know to ignore this commit when walking the
+history to look for changes.
```
FintanH, line 358: I'm not sure I understand this, could you expand on it?

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,649 @@
...
+### Reconstructing Collaborative Objects
+
+Assuming we have replicated a number of collaborative objects from our tracking
+graph, we can now view the merged state of those objects. To do this we search
+through every `/rad/collaborative-objects/<typename>/<object ID>` reference for
+every remote we have and collect the change files for each object ID.
+
+At this point we have the hash linked graph of automerge changes, but we need
+to make sure that the merged document is authenticated and valid with respect
+to it's schema. To do this we start at the roots of the hash graph and walk
```
FintanH, line 374:
```suggestion
to its schema. To do this we start at the roots of the hash graph and walk
```

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,649 @@
...
+down the tree. As we encounter each change we check it's signature, apply it
```
FintanH, line 375:
```suggestion
down the tree. As we encounter each change we check its signature, apply it
```

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,649 @@
...
+repository - it can be performed in memory and may be cached. The result is
+that the user sees a single merged view of the object based on the contents
+of the remotes they have replicated. That is, there is no additional
+merge-then-commit step.
+
+### Fetching Collaborative Objects
+
+Each time a repository creates a collaborative object tree it creates a ref
+pointing to that object at `refs/namespaces/<project>/rad/collaborative-objects/<typename>/<object ID>`, 
+where `object ID` is a unique identifier generated at creation time.
```
FintanH, line 390: Is this the UUID in the `manifest`? I'm not sure if that is explicitly stated anywhere

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,649 @@
...
+        }
+    }
+  }
+}
+```
+
+An issue consists of a title, description, and author along with the author's
+signature; followed by a list of comments, each of which is signed by it's
+respective author. This is an extremely simplified model. Note the presence of
+the `automerge_type`, `frozen`, and `rad_signed_by`, and `rad_urn` keys.
```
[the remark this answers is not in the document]
FintanH: It would be good to link to the [Schema Extensions](https://github.com/radicle-dev/radicle-link/pull/662/files#diff-a302abf8a4acae0a3e0b65f64f9079fadd3ae105bff1978cf2736f498ba92be6R431) section here.

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,649 @@
...
+
+Many collaborative data structures will need to make statements about who is
+allowed to change what parts of a structure. To achieve this we extend the 
+json schema language with some custom metadata, the `rad_signed_by` property. This
+property can be placed on any `object` schema. It's value is an
+object with two keys, an array of properties which must be signed, and array of 
+radicle URNs who's signature must be present.
+
+This property implies a required `signatures` property which contains a mapping
+from URN to a description of the signature. (Maybe a JWS?).
```
FintanH, line 443: What do you mean by "description of the signature" here?

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,649 @@
...
+#### `frozen`
+
+Some attributes should never be changed, for example the ID of an issue, or a
+nonce on a comment. Any schema can add the metadata key, `frozen: true` to
+indicate that after it's initial creation any change which modifies it is
+invalid.
+
+#### `automerge_type`
+
+This can take the value `"text"` if  placed on `string` properties to indicate
```
FintanH, line 460: Is this the only type or are you simply giving an example?

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,649 @@
...
+This model introduces some counter-intuitive properties. For example, I might
+"create an issue" in a repository and anyone who is tracking me would see that
+issue, but people who are tracking the project but don't have me in their
+tracking graph will only see the issue if the maintainer replies to it. It's
+hard to see how you would do things like "link to an issue" under these
+constraints. This is inherent to the network model though, rather than being a
+specific problem of this architecture.
+
+We can work around some of this weirdness using seed nodes. If we consider
+seed nodes
```
FintanH, line 475:
```suggestion
seed nodes.
```

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,649 @@
...
+- retrieve an object with a particular ID as an Automerge document for
+  applications which wish to write data
+- update an object by providing the bytes of an automerge change which updates
+  the document
+- create a new object from a JSON object, a JSON schema, and a type name
+  
+Note that I am referring to "the binary representation of an automerge x" 
+because the automerge API works in terms of binary changes.
+
+This new api will live in a new top level module at
```
FintanH, line 493:
```suggestion
This new API will live in a new top-level module at
```

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,649 @@
...
+    signer: signer::Signer,
+}
+
+enum History {
+    Automerge(Vec<Vec<u8>>)
+}
+
+struct ObjectId(String);
+struct TypeName(String);
+struct Schema(..);
```
FintanH, line 508: `..`? :eyes: Is this also just bytes or does the `..` mean you're still figuring that out?

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,649 @@
...
+struct CollaborativeObject {
+    typename: TypeName,
+    schema: Schema,
+    id: ObjectId,
+    author: Person, 
+    json: serde_json::Value,
+    history: History, 
+}
+
+struct NewObjectSpec {
```
FintanH, line 519: What's this struct for? :thinking:

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,649 @@
...
+    typename: TypeName,
+    history: History,
+    schema_json: serde_json::Value,
+}
+
+impl CollaborativeObjectStore {
+    fn retrieve_objects(&self, typename: String) -> Result<_, Vec<CollaborativeObject>>
```
FintanH, line 526: Have you been writing some Haskell? :smiling_imp: Or are all these objects errors? :P

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,649 @@
...
+hard to see how you would do things like "link to an issue" under these
+constraints. This is inherent to the network model though, rather than being a
+specific problem of this architecture.
+
+We can work around some of this weirdness using seed nodes. If we consider
+seed nodes 
+
+## APIs
+
+The APIs librad will provide:
```
FintanH, line 479: I'm wondering if it's worth adding retrieval of schemas to the API? This will likely be needed for modifying them in the future proposal as well.

FintanH (commented):
I left a few grammar fixes for you :smile: 

But I also have some clarifications and suggestions for the structure of the document.

One thing I'm wondering is that the recommendations are all-in with the git implementation, but is there room to discuss the implementation in a more general fashion ala the `Identity` spec -- that used an abstract `Revision`, for example, and then gave the concrete git implementation.

docs/rfc/0003-collaborative-objects.md
alexjg: :facepalm:, fixed.

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,649 @@
...
+
+The `automerge_type` key indicates that this field should be stored as a
+special data type in the automerge document which tracks individual character
+inserts and deletions, this allows us to merge edits to the same piece of text
+in a non surprising manner.
+
+The `frozen` key indicates that any change which modifies this field should be
+ignored.
+
+The `rad_signed_by` key is is more interesting it tells librad to validate that
```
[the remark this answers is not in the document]
alexjg: This `rad_signed_by` schema extension is the part of this RFC I am least sure of. My intention here was that the `rad_signed_by` key references a URN and validation checks that a signature from a key attached to that URN is present. In this sense I don't think I am advocating for just using `Person`?

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,649 @@
...
+|--author
+```
+
+This tree contains a single change to a collaborative object. We will go into
+more details shortly. Any direct dependencies of this change are encoded in the 
+same manner and become the parents of this commit. This allows us to
+reconstruct the automerge depdency graph. 
+
+Along with the dependencies of the commit we also need to add the commit of the
+identity which created this commit. We need this identity to validate
```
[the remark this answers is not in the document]
alexjg: Identity here refers to the head of the identity which is the author of this change. This is the same commit whose hash is referenced in the `author` entry. I've explicitly used the term `identity` because I don't _think_ we need to specify whether it's a person or a project.

docs/rfc/0003-collaborative-objects.md
alexjg: You mean linking to the json schema docs?

docs/rfc/0003-collaborative-objects.md
alexjg: The purpose of representing each change as a hash is to encode the dependency graph of the automerge document into Git. This has a few nice properties:

- We know that if we have `fetch`ed a change then we have also fetched all of the dependencies of that change
- We can deduplicate storage of common dependencies, even across remotes. This is useful if many people are (for example) responding to an issue, they can all depend on each others changes without copying them on disk.

However, when we're validating the hash graph we also need to ensure that the identity of the author of the change is present so that we can check the signature on the change. Therefore we add the commit of the head of the author identity as a parent of the change commit to ensure that it is fetched along with the change. 

Now, we want to be able to reconstruct the automerge change graph from the git change graph. We do this by starting at the latest change and walking the commit tree, recursively adding each parent to the automerge change graph. As we do this we want to exclude the parent commits which refer to the signing identity rather than an automerge change, but there's no way to tell whether a commit is an automerge change or not without examining it, so the `author` file contains the hash of the commit which is the identity so we can exclude it from our tree walk as we construct the automerge change graph.

docs/rfc/0003-collaborative-objects.md
alexjg: Added a reference.

docs/rfc/0003-collaborative-objects.md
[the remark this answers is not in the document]
FintanH: In the Identity work, there are two kinds of delegations `Direct` and `Indirect`. The former means that you are delegating directly to keys, whereas the latter means that you are delegating to a `Person` which in turn means you are using _one_ of the keys associated with that `Person`. So, when you say that the signature is using an `Urn` then I would think it's referring to a `Person`.

docs/rfc/0003-collaborative-objects.md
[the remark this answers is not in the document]
FintanH: What would it mean for the `author` to be a `Project`? :thinking:

docs/rfc/0003-collaborative-objects.md
FintanH: Yup!

docs/rfc/0003-collaborative-objects.md
FintanH: Ah that explanation was super helpful!

docs/rfc/0003-collaborative-objects.md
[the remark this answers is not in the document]
alexjg: Would it not be valid to say that `rad_signed_by` for an `Indirect` delegation is requiring that the signature be by one of the delegated `Person`s keys? The reason I want to keep this is that there will be situations where you want to do things like allow "any project maintainer" to change something - specifying the URNs as Persons means that new maintainers will be unable to make changes.

docs/rfc/0003-collaborative-objects.md
alexjg: I'm still working it out :slightly_smiling_face:. More to the point, it doesn't actually matter for the API as the purpose of the tuple struct is to hide the implementation, I guess I could do the same for the `ObjectId` and `TypeName` for consistency.

docs/rfc/0003-collaborative-objects.md
alexjg: It's just a parameter object for the `create_object` call.

docs/rfc/0003-collaborative-objects.md
alexjg: The difficulty here is that in principle each object can have a different schema. This means that retrieving a schema is the same as retrieving an object (as the object includes its schema). It would be nice to require that all schemas of a type have the same schema - however, I think this introduces a griefing opportunity where someone publishes objects of a given type with a different schema to an already agreed schema and this poisons that type for anyone who syncs with them before syncing with well behaved remotes. I am still trying to figure out a nice way of handling this situation.

docs/rfc/0003-collaborative-objects.md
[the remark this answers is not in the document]
FintanH: > Would it not be valid to say that rad_signed_by for an Indirect delegation is requiring that the signature be by one of the delegated Persons keys? 

That's correct :) 

> specifying the URNs as Persons means that new maintainers will be unable to make changes.

Right, so I don't think this exists yet. You can't delegate to a project. But maybe it's something to think about :) #orgs #teams

docs/rfc/0003-collaborative-objects.md
FintanH: > as the object includes its schema

Ah yes, that's a good point :) 

> I think this introduces a griefing opportunity where someone publishes objects of a given type with a different schema to an already agreed schema and this poisons that type for anyone who syncs with them before syncing with well behaved remotes.

Ya, I suppose this would fall under the umbrella of "consensus on schemas", where privileged persons can agree on the schema and a change would need validation from those persons.

docs/rfc/0003-collaborative-objects.md
FintanH: Ah right, sorry missed the use of it below :)

alexjg:
Vis a vis non-git implementations, it would certainly be possible to talk in terms of synchronizing and validating an abstract graph of CRDT changes. I think this would be valuable and it's something I am interested in doing but I'm not sure that it's useful for this  RFC. I think one of the values of these RFCs is hammering out the more fiddly implementation details so it makes sense to be more concrete.

docs/rfc/0003-collaborative-objects.md
alexjg: Yep. Now, one downside of not requiring that every object be of a particular schema is that the work of schema validation then falls on applications. This is not a deal breaker (as Mongodb has conclusively shown) but it's not ideal. One workaround might be to allow applications to specify a schema when they fetch objects, any object which doesn't match that schema would not be returned.

docs/rfc/0003-collaborative-objects.md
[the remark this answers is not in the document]
alexjg: Practically though, we do have all the keys necessary to implement this flow no? If I have the head commit of a `Project`, then I can load it using `Identities::some_identity` and then I can check if a given public key is in that `Project` using `Indirect::eligible` right?

docs/rfc/0003-collaborative-objects.md
FintanH: Ya, to be honest, I prefer the idea of a zoo of compatible schemas. It means that while a project might have its own schema I could add additional metadata that I find interesting but which may be ignored by the project maintainers.

FintanH: But I guess the question then is: How does one _propose_ a change to the "agreed upon" schema?

docs/rfc/0003-collaborative-objects.md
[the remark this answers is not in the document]
FintanH: Ya that sounds right :)

FintanH:
Another thought I just had was, "Does this RFC need to be particular about `Project`s?" I feel like it might make sense to have "collaboration objects" on a `Person`, but one would essentially just be collaborating with themselves. For example, I could have a list of "favourite projects" on my `Person` identity, that is tracked over my devices. It's a good chance for `Person` identities to have metadata that could change frequently, and so wouldn't be suitable to be kept in the `Doc` itself.

alexjg:
Interesting thought. We would need to add a similar `signed_refs` construct to `Person` identities as at the moments - as far as I'm aware - we don't replicate anything except the identity document itself for `Person` identities.

I think this is probably worth doing  but might be best left to a later RFC?

FintanH:
> We would need to add a similar signed_refs construct to Person identities as at the moments - as far as I'm aware - we don't replicate anything except the identity document itself for Person identities.

It's your lucky day, `Person` also has `signed_refs` :) Check it out on a file system near you!

alexjg:
But we don't replicate that at the moment right? Specifically we only call `replicate_signed_refs` in `librad::git::replication::project::ensure_setup` and not in `librad::git::replication::person::ensure_setup. Still, that seems like it would be pretty trivial to add. I would still like to leave this to a later RFC if possible. It seems like it would be a straightforward extension of the concepts in this RFC but I like to keep my scopes nice and focused :telescope:

FintanH:
Ah yes, that's a good point.

What first brought this thought on was that in your MVP PR you have `project_urn` as fields and I was thinking that it might not have to be particular to a project. In fact, it would probably be tricky to scope these things to _only_ project `Urn`s unless we check that it always resolves to a project.

docs/rfc/0003-collaborative-objects.md
kim: Is the hash of a `change` blob actually guaranteed to be stable?

There is also the opportunity to not having to follow the blob pointer of `author` by either encoding the SHA1 in the file name itself (like git notes), or adding a trailer to the commit (a la `x-rad-author:`).

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,649 @@
...
+
+Therefore we add two parameters to a radicle URN:
+
+- `collab_object_id`: specifies that a server should only consider references to the given object ID
+- `collab_type_name`: specifies that a server should only consider references to the given type name
+
+These parameters are emitted by the radicle p2p transport as headers in a 
+similar fashion to the `n=` parameter for the nonce or the `ls=true`
+parameter for selecting the `upload-pack-ls` service. The Radicle git server
+can then use these parameters to filter the refs it operates on.
```
kim, line 415: It is potentially the right idea to add parameters to a _URL_, but the mechanics for the _URN_ are a bit different: essentially, every `URN/PeerId` pair is a git remote, and whenever you tell link to get a URN, you're creating a (persistent) remote. To restrict which branches are fetched, git's standard mechanism is a _fetch spec_ (the `git.<remote name>.fetch` key). This is something we can consider, but note that it has an effect on availability of the data (you either replicate those objects, or you don't).

Of interest is a way to "peek" at potential updates / contributions from peers which are _not_ in your tracking graph. This can be done semi-efficiently with v2 (which we'll hopefully be able to use soon) by just issuing a `ls-refs` with a `ref-prefix`. We need a place to store-and-forward this information, though.

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,649 @@
...
+json schema language with some custom metadata, the `rad_signed_by` property. This
+property can be placed on any `object` schema. It's value is an
+object with two keys, an array of properties which must be signed, and array of 
+radicle URNs who's signature must be present.
+
+This property implies a required `signatures` property which contains a mapping
+from URN to a description of the signature. (Maybe a JWS?).
+
+Any schema which has this property will result in some additional validation.
+Librad will encode the relevant keys of the target object using canonical JSON
```
kim, line 444: > canonical JSON

Which implies that objects can't have floating-point values 🙃

kim (commented):
Coming along nicely! I also accept the defeat on `coq-embed` ... for now 😈

kim:
> Does this RFC need to be particular about Projects

No, there's no difference in that sense (or it's a bug :)). The more general term would be "URN".

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,649 @@
...
+
+
+## Blessed Data Types
+
+This project metadata mechanism is extremely broad, which has a lot of upsides
+but it runs the risk of running into XMPP style extension hell, where every
+peer is running a different set of extensions. It might be worthwhile to bundle
+a few core extensions with librad - issues for example.
+
+## Alternative Approaches
```
geigerzaehler, line 541: I would love to see a comparison with the event log style approach of [SSB](https://ssbc.github.io/scuttlebutt-protocol-guide/#feeds) and [DAT](https://datprotocol.github.io/how-dat-works).

kim: This is actually subsumed by “Domain Specific CRDTs” and “Blessed Datatypes” — SSB applies an arbitrary fold function to a fixed set of datatypes, and leaves it to applications to agree on what the function is. The prevalent “multiwriter” thing for DAT is, afaik, “hypermerge” (which, incidentally, is based on automerge).

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,649 @@
...
+```json
+{
+    "typename": "https://radicle.xyz/issue",
+    "schema": <the schema above>,
+    "data": {
+        "title": "Librad doesn't implement the metadata RFC",
+        "description": "It's in the name",
+        "author": "<some base64>",
+        "signatures": {
+            "<author URN>": "<some base64>"
```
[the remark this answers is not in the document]
alexjg: I've modified this to match the description in the `rad_signed_by` specification.

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,649 @@
...
+Given that automerge changes are a hash linked graph, we can map them to Git.
+We do so by wrapping each change in a commit. The commit points at a tree with
+the following layout
+
+```
+.
+|--manifest
+|--schema
+|  |--schema.json
+|  |--migrations
```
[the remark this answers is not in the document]
alexjg: Interesting point. I'll have a think about this. It may be worth moving schemas into their own hash graph and then referencing them in a commit trailer much like the author.

docs/rfc/0003-collaborative-objects.md
alexjg: My intent with these parameters was to avoid advertising all the refs of the collaborative objects unless the peer wants them. As I understand protocol V1 the fetch spec approach does not avoid this right?

docs/rfc/0003-collaborative-objects.md
alexjg: Ugh, good point, I've modified this to specify CBOR instead.

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,649 @@
...
+        "const": "<the authors URN>",
+        "frozen": true
+    },
+    "comments": {
+        "type": "array",
+        "items": {
+            "type": "object",
+            "rad_signed_by": {
+              "properties": ["text", "author"],
+              "keys": ["<commenters URN>"]
```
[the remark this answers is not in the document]
alexjg: Good spot. To address this I've specified that we include the [JSON data access vocabulary](https://gregsdennis.github.io/json-everything/usage/vocabs-data.html) which allows the schema to specify that the signature must belong to the `author` of a comment.

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,649 @@
...
+        "type": "string",
+        "automerge_type": "text"
+    },
+    "author": {
+        "description": "The radicle ID of the author of the issue",
+        "const": "<the authors URN>",
+        "frozen": true
+    },
+    "comments": {
+        "type": "array",
```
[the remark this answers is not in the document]
alexjg: It's more complex, I've reworked the example to represent comments as a tree structure with some validation that a comment is the child of a particular comment ID. This is via a `parent_id` on the comment which is then signed by the `author`. This still allows for reordering within the comments of a particular comment - but that's somewhat unavoidable in the face of concurrent modifications anyway.

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,649 @@
...
+us to place operations in a total order which respects causality, i.e. if I add
+an operation then no operation that I could have observed at the time I made
+the operation will come after it in the log.
+
+Automerge defines a number of operations along with merge semantics for those
+operations. More detail on that can be found in [the implementation](https://github.com/automerge/automerge)
+and in [the paper](https://arxiv.org/abs/1608.03960).
+
+Operations in automerge are transported in batches called "changes". Each 
+change references zero or more changes it depends on via their hash. In this
```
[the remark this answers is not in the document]
alexjg: I think this is explained in the "Change Commits" section, do you think there is anything missing there?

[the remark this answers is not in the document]
geigerzaehler: I was just wondering if it makes sense to explain it here in one sentence. When reading the document top to bottom this made me stop and think about this here. The explanation later doesn’t help with that unfortunately.

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,716 @@
...
+            "children": {
+                "type": "array",
+                "items": {
+                    "$ref": "#/definitions/comment"
+                }
+            }
+            "parent_id": {
+                "oneOf": [
+                    "data": {
+                        "const": "1/id"
```
geigerzaehler, line 89: I have trouble understanding how this would work. The [data vocabulary](https://gregsdennis.github.io/json-everything/usage/vocabs-data.html) says that `1/id` is a URI evaluated in the context of the document that is being validated. So I assume you mean `#/1/id` to treat it as a URI instead of just a JSON path. But then again `/1/id` is not a valid path for an issue object because the object itself is not an array. So do you mean `/comments/1/id`. But this would always the second comment.

docs/rfc/0003-collaborative-objects.md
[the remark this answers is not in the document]
geigerzaehler: Can you explain how this prevents anybody from removing a comment from the `comments` array? Could you may be give a concrete example?

docs/rfc/0003-collaborative-objects.md
geigerzaehler: > The prevalent “multiwriter” thing for DAT is, afaik, “hypermerge” (which, incidentally, is based on automerge).

Thanks for clarifying. I’m not intimately familiar with DAT.


> This is actually subsumed by “Domain Specific CRDTs”

I read this section as being more focused on CRDTs and canoncial representations where as SSB is a bit more loose with it. I’ll add some more specific comments there.

docs/rfc/0003-collaborative-objects.md
alexjg: `1/id` is a relative pointer, it means "go up one level from the current object and look at the `id` key. The vocabulary actually specifies that "The instance being validated" is the context for URI resolution, in this case the instance is the `"parent_id"` key (I think, the details of the relative pointer resolution stuff are defined in a somewhat indirected manner, at the end of the day it will depend on what implementations do).

docs/rfc/0003-collaborative-objects.md
kim: We don't have to worry about v1 anymore -- I received a walkthrough from @Byron yesterday, and what I saw is sufficient to start ripping out libgit2 for network access NAO.

What we _should_ consider, however, is to make the refs paths [ref-prefix](https://git-scm.com/docs/protocol-v2#_ls_refs) friendly. Since we want `[remotes/*/]rad/{id,ids/*,signed_refs}` for replication (where we know the values of `remotes/*`), it might be worth considering to put the collaborative objects under their own namespace, eg. `cob/<type>/<uuid>`.

docs/rfc/0003-collaborative-objects.md
kim: Mya, not a [silver bullet](https://datatracker.ietf.org/doc/html/rfc7049#section-3.9). Canonicalisation of floats specifically is only defined non-normatively.

alexjg: Ugh, any other suggestions?

docs/rfc/0003-collaborative-objects.md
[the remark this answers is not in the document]
kim: That someone who was a member of the project at that point in time made a change on behalf of the project. I’m not sure how much effort we’d be willing to make to resolve this to “clowntoe@hy... on behalf of sy@hy...“ for rendering, but it makes sense to me semantically =]

docs/rfc/0003-collaborative-objects.md
alexjg: Exciting! I'll rework this in that light then. Having a top level `cob/type/<uuid>` makes sense for prefixing purposes.

docs/rfc/0003-collaborative-objects.md
kim: It’s not entirely clear to me what the actual payload for the signature is — when it’s an update, would it be only the actual automerge change, or the result of applying the change? If automerge supports difficult-to-canonicalise types such as floats, how does it handle them? If a change is produced deterministically (ie. if I put a float field as 0.23417, something will prevent me to accidentally create the same change but with a rounding error), and changes have an identity, does canonicalisation actually matter?

alexjg: The payload is the given properties of the realization of the state of the automerge document after having applied the automerge change. i.e the result of applying the change. Automerge has both 32 and 64 floating point numbers which are little endian encoded in the document. The automerge libraries are designed so that these types round trip through serialization and deserialization, so in this sense they are deterministic. We could stipulate that we use canonical CBOR with the float type being the type of the underlying automerge value.

The intention of the `rad_signed_by` schema is to impose constraints on parts of the state of the CRDT, so we can't use the change  directly. For example, a user might update their comment by inserting a character, the underlying op log would have two operations: one inserting a character and one updating the siganature, but we want to validate that the signature is valid over the author and the whole comment, so we can't use the change itself as the item over which to validate the signature.

docs/rfc/0003-collaborative-objects.md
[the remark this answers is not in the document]
alexjg: I've added a `grow_only` key to the `children` of a comment, which should address this concern.

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,727 @@
...
+inserts and deletions, this allows us to merge edits to the same piece of text
+in a non surprising manner.
+
+The `frozen` key indicates that any change which modifies this field should be
+ignored.
+
+The `grow_only` key indicates that elements may not be removed from this 
+sequence. This allows us to ensure that comments cannot be removed.
+
+The `rad_signed_by` and is more interesting. The  `rad_signed_by` field tells
```
FintanH, line 152:
```suggestion
The `rad_signed_by` is more interesting. The  `rad_signed_by` field tells
```

docs/rfc/0003-collaborative-objects.md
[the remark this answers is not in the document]
geigerzaehler: Does `grow_only` also mean that existing items are immutable and does it also apply to the top-level `comments` property? If so, you can’t add nested comments because it would mutate a comment.

```
doc.get("comments).idx(0).get("children").push(...)
```
This wouldn’t work because it would mutate the first comment.

On the other hand, if the items are not immutable a comment could be completely replaced by a malicious actor. They just need to sign the replacement comment with their own signature.

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,727 @@
...
+--- schema.json
+```
+
+This can be extended in future by creating schema commits that reference this
+schema commit and add migrations.
+
+As with change commits the schema commit is signed and references an author
+commit, therefore the commit has two trailers:
+
+- `X-Rad-Auther-Parent`
```
FintanH, line 468:
```suggestion
- `X-Rad-Author-Parent`
```

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,727 @@
...
+### Schema Commits
+
+Schemas are important for the interoperability of the system. We need
+applications to be able to rely on the data they are working with being valid,
+otherwise we impose the problem of schema validation on application developers.
+
+Schemas will need to be able to change over time. Schema migration is out of 
+scope for this RFC but we need a minimal mechanism to support it in future. To
+this end schemas are represented using their own hash graph. For the purposes
+of this RFC a schema is a commit with a tree that contains a single `schema.json`
```
FintanH, line 455: Where are these commits going to be found?

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,727 @@
...
+### Reconstructing Collaborative Objects
+
+Assuming we have replicated a number of collaborative objects from our tracking
+graph, we can now view the merged state of those objects. To do this we search
+through every `/rad/collaborative-objects/<typename>/<object ID>` reference for
+every remote we have and collect the change files for each object ID.
+
+At this point we have the hash linked graph of automerge changes, but we need
+to make sure that the merged document is authenticated and valid with respect
+to it's schema. To do this we start at the roots of the hash graph and walk
```
geigerzaehler, line 390: Is the plural “roots” intentional? If I understand this correctly, there is only one root: the change that created the document. It would also help clarify to make explicit what the root is in the RFC.

docs/rfc/0003-collaborative-objects.md
```diff
@@ -0,0 +1,727 @@
...
+down the tree. As we encounter each change we check it's signature, apply it
+and check that the new document does not violate the schema. If it does violate
+the schema we discard the change and all dependent changes. Finally, we have an
```
geigerzaehler, line 393: This might be tricky with concurrent edits.

For example we could have a document `{ foo: ["x"] }` where the length of the array at `foo` must be less than 3. If two peers concurrently add an element to produce `{ foo: ["x", "y"] }` and `{ foo: ["x", "z"] }` then each individual change is ok. But as soon as a peer sees both of them and combines them they become invalid. Do we reject both of the changes now? This might allow people to sabotage documents.

I’d also like to know more about whether we want to reject updates at the gossip layer. I.e. do we forward messages that produce invalid documents?

docs/rfc/0003-collaborative-objects.md
geigerzaehler: Thanks for the explanation. I’m also referencing the [“Relative JSON Pointer”](https://tools.ietf.org/id/draft-handrews-relative-json-pointer-00.html) specification that you mentioned in our private conversation.

docs/rfc/0003-collaborative-objects.md
kim: Interesting. I suppose it’s potentially useful to retain the signatures in the document, so they can be validated on the frontend as well. The lossy typing through JSON requires to define canonicalisation rules, though, which interestingly are the opposite of what RFC 7049 says: we need to use the _largest_ representation.

Aside: it’s an interesting choice that automerge doesn’t seem to concern itself with Unicode normalisation, given that applications typically won’t display diffs which would show that U+212B, U+0041U+030A and U+00C5 are considered three different characters. This means that we shouldn’t normalise either, because that would increase the number of possible inputs to the signature function.

So, my take would be:

—-
To compute a signature, the values at the keys specified in `rad_signed_by.keys` are looked up from the document in the order given, and converted back-to-back into CBOR according to the following canonicalisation rules:

* `integer` values are encoded as major types 0 and 1, respectively, with additional value 27 (64-bit) 
* `number` types are encoded as major type 7 with 5-bit value 27 (IEEE 754 double-precision float)
* `object` types are encoded as major type 5 (maps), applying the canonicalisation rules as per RFC 7049, section 2.3
* The remaining schema types have unambiguous mappings to CBOR types


When the input is in the automerge data model, the mappings are as follows:

* f32,f64: major type 7,27
* int: major type 0 or 1,27
* uint: major type 1,27
* bytes: major type 2, where the length is represented as a 64-bit unsigned integer (major type 1,27)
* str: major type 3, where the length is represented as a 64-bit unsigned integer (major type 1,27)
* boolean: major type 7, 20 or 21 respectively 
* counter, timestamp: as their corresponding integer value
* null: major type 7, 22


Cursors, undefined values, and constructions which can not be represented as definite-length arrays / maps in CBOR are illegal for signed values.

Finally, the signature is computed over the binary CBOR representation and stored in the `signature` property. To verify the signature, the payload is computed in the same way, and verified against the `signature` and `key` properties.

docs/rfc/0003-collaborative-objects.md
kim: Perhaps JSON schema constraints are only applicable insofar as the underlying CRDT can guarantee them? As in, perhaps we need to define a vocabulary for the kinds of constraints which are valid?

> I’d also like to know more about whether we want to reject updates at the gossip layer. I.e. do we forward messages that produce invalid documents?

We do not have individual messages, so the only choice would be to stop replicating the entire history of the violating peer from the invalid commit on. Given that the violation may be because of a bug or unsound constraint, I'd rather leave this decision to the human. I also don't think that there should be any unattended eager evaluation: there could be errors, conflicts, or even timetravel due to newly-discovered remotes. All of these need to be surfaced to the user somehow, so evaluation would more naturally occur on read.

alexjg: I'm currently working on putting together an alternative JSON schema vocabulary because I think that's the most principled way to approach this. We would choose a subset of [JSON schema validation](http://json-schema.org/draft/2019-09/json-schema-validation.html) (the default vocabulary) which only includes constraints that are commutative (e.g `oneOf` for values, but not `maxItems` for arrays). This vocabulary would also include the schema extensions we will need for authenticating data structures (e.g `rad_signed_by` etc.).

docs/rfc/0003-collaborative-objects.md
alexjg: It's mentioned earlier in the document that they are referenced by a commit trailer on changes, I'll add a sentence to that effect here as well.

docs/rfc/0003-collaborative-objects.md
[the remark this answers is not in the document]
alexjg: These are all good points and lead me to think that the schema based approach for validating access control is a dead end. I'm investigating an alternative capability based access control system which would be separate to the schema. I think JSON schema still makes sense for the basic structure of the data though.

docs/rfc/0662-collaborative-objects.adoc
```diff
@@ -0,0 +1,789 @@
...
+of this object is unique. More details in <<Object IDs>>.
+
+==== `change`
+
+This is the automerge change which this commit introduces. It is a binary file
+which must contain a single change and it's dependents must be the dependents
+referenced by the parents of the commit.
+
+
+=== Object IDs
```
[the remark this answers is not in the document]
kim: The identities use the `tree` hash to enable a "multisig" scheme on the `commit` objects. The doc `blob` contains a hash link to the previous `tree`.

Iiuc, cobs could just be named after their initial `commit` (which contains sufficient entropy), or not? The histories of collaborators evolve independently anyways, and are merged only virtually.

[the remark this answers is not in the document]
alexjg: Yep using the initial `commit` works just as well, and probably better as it's authenticated. I was wary of doing this because it means we're depending on a bit more Git for the generation of the ID (tree hash + commit format), but on reflection I don't think being independent of Git is important for this.

docs/rfc/0662-collaborative-objects.adoc
```diff
@@ -0,0 +1,778 @@
...
+
+Along with the dependencies of the commit we also need to add the commit of the
+identity which created this commit. We need this identity to validate
+signatures and by making the commit a parent we ensure that git will replicate
+it for us. 
+
+A valid change commit must have three trailers:
+
+* `X-Rad-Signature`, as for identity documents
+* `X-Rad-Author-Parent`, this is the hash of the commit which references the
```
FintanH, line 326: Is this the root hash or a particular revision of `namespaces/<author>/rad/id`? So in other words, will this hash point to an author at a certain point in time?

docs/rfc/0662-collaborative-objects.adoc
```diff
@@ -0,0 +1,778 @@
...
+and other properties of the object.
+
+To have a single root we need an object ID which is derived from the attributes
+of the object, that way an attacker cannot manipulate the attributes of an
+object with the same ID. We achieve this by using the hash of the initial
+commit of the object as the ID.
+
+It is entirely plausible that a peer would create two distinct objects with
+identical initial states. Under many content addressing schemes this would lead
+to the two objects having the same ID, however, git Commits include a timestamp
```
FintanH, line 371:
```suggestion
to the two objects having the same ID, however, git commits include a timestamp
```

docs/rfc/0662-collaborative-objects.adoc
```diff
@@ -0,0 +1,778 @@
...
+=== Reconstructing Collaborative Objects
+
+Assuming we have replicated a number of collaborative objects from our tracking
+graph, we can now view the merged state of those objects. To do this we search
+through every `/cob/<typename>/<object ID>` reference for
+every remote we have and collect the change files for each object ID.
+
+At this point we have the hash linked graph of automerge changes, but we need
+to make sure that the merged document is authenticated and valid with respect
+to it's schema. To do this we start at the root of the hash graph and walk
```
FintanH, line 384:
```suggestion
to its schema. To do this we start at the root of the hash graph and walk
```

docs/rfc/0662-collaborative-objects.adoc
```diff
@@ -0,0 +1,778 @@
...
+parts of a document via a
+https://tools.ietf.org/id/draft-handrews-relative-json-pointer-00.html[relative
+JSON pointer] when expressing constraints.
+
+==== `rad_signed_by`
+
+Many collaborative data structures will need to make statements about who is
+allowed to change what parts of a structure. To achieve this we extend the 
+json schema language with some custom metadata, the `rad_signed_by` property. This
+property can be placed on any `object` schema. It's value is an
```
FintanH, line 476:
```suggestion
property can be placed on any `object` schema. Its value is an
```

docs/rfc/0662-collaborative-objects.adoc
```diff
@@ -0,0 +1,778 @@
...
+signature is computed over the given properties and checked against the given
+key and signature.
+
+To compute a signature, the values at the keys specified in rad_signed_by.keys
+are looked up from the document in the order given, and converted back-to-back
+into CBOR according to the following canonicalisation rules:
+
+* `integer` values are encoded as major types 0 and 1, respectively, with
+  additional value 27 (64-bit)
+* `number` types are encoded as major type 7 with 5-bit value 27 (IEEE 754
```
FintanH, line 512: Could you add references to "IEEE 754" and "RFC 7049"?

docs/rfc/0662-collaborative-objects.adoc
```diff
@@ -0,0 +1,778 @@
...
+
+This project metadata mechanism is extremely broad, which has a lot of upsides
+but it runs the risk of running into XMPP style extension hell, where every
+peer is running a different set of extensions. It might be worthwhile to bundle
+a few core extensions with librad - issues for example.
+
+
+== Further work
+
+This RFC limits participants in collaborative objects to project maintainers.
```
FintanH, line 616: I'm not following where this limitation comes from :thinking: Could you elaborate on why that is?

docs/rfc/0662-collaborative-objects.adoc
```diff
@@ -0,0 +1,778 @@
...
+
+== Blessed Data Types
+
+This project metadata mechanism is extremely broad, which has a lot of upsides
+but it runs the risk of running into XMPP style extension hell, where every
+peer is running a different set of extensions. It might be worthwhile to bundle
+a few core extensions with librad - issues for example.
+
+
+== Further work
```
FintanH, line 614: I think it's also worth calling out that schema migrations are also considered further work

docs/rfc/0662-collaborative-objects.adoc
alexjg: The reason is that we can't make assertions we would like to make about the document in general via schema validation. For example, we would like to assert that no one can remove a comment which they did not create, but this can't be expressed via JSON schema. Therefore we limit modification to maintainers who are probably somewhat trusted and we leave access control as a future problem (one which I have a lot of ideas about, I just think we should reduce the risk on this RFC and implement access control later).

docs/rfc/0662-collaborative-objects.adoc
alexjg: A particular revision.

docs/rfc/0662-collaborative-objects.adoc
FintanH: Ok, that makes sense. So will maintainers also be the only ones able to comment initially as well? Or in other words, what's the extent of "modification"?

alexjg: Yep, for issues initially only maintainers will be able to create and comment. Obviously this is not sufficient in general, but it allows a lot of usecases which _are_ useful off the bat and we can focus on access control next. I'm pretty sure access control is going to require a custom logic based language, so I want to isolate the work of specifying the syntax and semantics of that language to it's own RFC.

alexjg:
I've made a few significant changes to this document. Provided these changes pass muster I think this is ready for merge.

## `rad_signed_by`

I've removed the `rad_signed_by` key. This was intended to allow schema authors to express things like "comments must be signed by their author" in the schema. Unfortunately I couldn't find a simple way to make this work in the presence of conflicts. The problem eventually becomes equivalent to the more general access control problem which we have deferred to later work. Concretely this means that metadata on components of a collaborative object (such as an `author_urn` field on a comment on an issue) will not be authenticated - anyone can pretend to be anyone else. The impact of this should be minor as for this implementation of collaborative objects the only people who can modify collaborative objects are project maintainers, who should be relatively trustworthy.

## Custom json schema vocabulary

Some keywords in the standard json schema vocabularies do not distribute across merges (consider the `maxContains` keyword). I have defined a custom vocabulary at https://alexjg.github.io/automerge-jsonschema/spec which specifies a subset of keywords which are safe to use on Automerge documents and specified that schemas MUST use this vocabulary.

docs/rfc/0662-collaborative-objects.adoc
```diff
@@ -0,0 +1,789 @@
...
+to be able to receive changes to the same data from other people and this
+raises the problem of how to merge those changes. I see two options here:
+
+* CRDTs, data structures which allow conflict free merges
+* Application level merges. Much as Git requires user action to merge
+  conflicting changes, we could expose the merge points of collaborative
+  objects to the user to resolve.
+
+The latter here is undesirable for a lot of reasons but most relevant is that
+it conflicts with extensibility and interoperability. in order to provide a good
```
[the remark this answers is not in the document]
FintanH: Ping =]

FintanH (dismissed):
Overall, this looks good to me. I mostly have grammatical suggestions to fix up :)

docs/rfc/0662-collaborative-objects.adoc
```diff
@@ -0,0 +1,690 @@
...
+It is important to note that this merging is at this point not stored in the
+repository - it can be performed in memory and may be cached. The result is
+that the user sees a single merged view of the object based on the contents
+of the remotes they have replicated. That is, there is no additional
+merge-then-commit step.
+
+=== Fetching Collaborative Objects
+
+Each time a repository creates a collaborative object tree it creates a ref
+pointing to that object at `refs/namespaces/<project>/cob/<typename>/<object ID>`,
```
kim, line 371: There is one (scalability) aspect that is still not entirely clear to me: are collaborators on a specific object supposed to merge all changes by the other collaborators into their copy of the object? If so, when does this happen?

Just to clarify:

Suppose peers `P_a`, `P_b`, `P_c` are maintainers as per this document (for simplicity, let's leave out key delegations). `P_a` and `P_b` apply some changes to collaborative object `O` within the context of project `R`, while `P_c` does not. Peers `P_d` and `P_e` are tracking `R`.

Now, `P_e` establishes a network connection to `P_d` with the goal of retrieving any updates to `R`. Because of the rule that delegate peers (aka maintainers) must be replicated, the refs layout `P_d` presents looks like this:

```
refs/namespaces/R/refs/cob/frob/O
refs/namespaces/R/refs/remotes/P_a/cob/frob/O
refs/namespaces/R/refs/remotes/P_b/cob/frob/O
refs/namespaces/R/refs/remotes/P_c/cob/frob/O
```

Because commits can be applied in any order by the involved peers, those `O`s won't converge to the same commit SHA. The question I have is, do they converge to the same tree SHA? That is, do `P_{a..e}` all apply all changes they see to their own version of `O`? Or only if they themselves apply a change, which requires to record the dependencies?

In the latter case, I think `P_d`'s view would not exist (because only maintainers can collaborate under this proposal). But, `P_e` needs to collect past and present maintainers from `R`, and fetch all corresponding remote tracking branches.

docs/rfc/0662-collaborative-objects.adoc
```diff
@@ -0,0 +1,690 @@
...
+A valid change commit must have three trailers:
+
+* `X-Rad-Signature`, as for identity documents
+* `X-Rad-Author-Parent`, this is the hash of the commit which references the
+  author identity. We use this trailer to avoid following the author commit
+  reference when constructing the automerge change graph
+* `X-Rad-Schema-Parent`, this is the hash of the parent commit which contains 
+  the schema of this object. See [schema commits](#schema-commits).
+
+Furthermore a change commit MUST be signed by an identity which was at some point
```
FintanH, line 299: I wonder if this limitation is required. Perhaps we could use a similar model of how branches are handled in identities. That is, all peers can add changes to their view of the object, but it is the maintainers' view that is the canonical view (or most important view). So in essence, non-maintainers are proposing comments to an issue, for example.

docs/rfc/0662-collaborative-objects.adoc
alexjg: The intention here is that all known views of the CRDT are merged at query time. In the scenario you're laying out above `P_e` would merge the changes from `P_{a..c}` when querying and see that as the state of the object. This merge is not a git merge, there is no commit that corresponds to it, it's just the merge of all the views of the collaborative object that `P_e` knows about. If `P_e` contributes a change then this is the point at which they would create a "change commit", which records the dependencies of that change in the git commit graph.

"merge at query time" is a simplification, what we will most likely want to do is merge when receiving new changes and cache the merged state somewhere. The point is that it's not part of the git commit graph. Conceptually it's the same and allows for `P_{d,e}` to have a view on the object without having to record any changes.

kim: That makes sense, thanks.

So, when `ls-ref`ing `P_d`, `P_e` would see:

```
refs/namespaces/R/refs/remotes/P_a/cob/frob/O
refs/namespaces/R/refs/remotes/P_b/cob/frob/O
```

Which means we can't limit the response by `ref-prefix` :(

alexjg: Hmm, that's a good point. I guess we could do it if we had a separate replication step for collaborative objects once we've fetched the tracking graph?

docs/rfc/0662-collaborative-objects.adoc
alexjg: This does have some appealing characteristics. Mainly that it is simple to
implement and it builds on intuitions people already need to have about Radicle.

However, I am also not sure that this is a good solution when we start thinking
about what collaborative objects are for. The canonical branch solution works
because maintainers are actively maintining those branches. Specifically they
are very aware of the meaning of the difference between different branches. When
a non-maintainer proposes a branch the maintainer will look at what the diff of
the new branch is and decide whether to track it as a canonical branch.

The situation with collaborative objects is different. The reason we want to
constrain objects to maintainers only is that there are assertions we want to
make about a collaborative object without having to examine the underlying data
structure. Consider the case of attributing authorship on an issue comment. This
RFC proposes no mechanism for ensuring that the comment was actually authored by
the person who's URN is on the comment. This means that in order to accept a
proposed issue a maintainer would need to examing the before and after state of
the entire issue checking that each comments author URN is valid. This is in
some sense the same work as examining a proposed branch, except that now it's
for a data structure the user has no understanding of or tooling for. 

I think this approach _could_ work but we would need to provide tooling for
maintainers to examine the difference between states of collaborative objects
and ways for schema authors to indicate what parts of an object should be
checked. This might be worth investigating if the completely automated
authorization system I envision turns out not to be possible.

This approach also feels promising for schema migrations.

docs/rfc/0662-collaborative-objects.adoc
kim: > once we've fetched the tracking graph

Right, that's the case anyway -- we would fetch `refs/namespaces/R/refs/remotes/P_a/cob/frob/O` if and only if it is mentioned in `refs/namespaces/R/refs/remotes/P_a/rad/signed_refs`, and only `want` the oid ibid.

What is more annoying is how to enable collaborators outside the inner cabal in the future: say you want to interrogate a seed node to see if it got changes by random peers for some object or class thereof. Or you would want to fetch the `cob/frob/O` ref of any random peer, without tracking that peer (ie. pruning it later if it turns out sketchy).

I suppose the only way to do this in a scalable way is to maintain some sort of reverse index by peer id, ie. `refs/namespaces/R/cob/frob/O/{self, P_a, P_b}`, where `P_*` are symrefs to the respective remote tracking heads.

docs/rfc/0662-collaborative-objects.adoc
FintanH: Well regarding authorship, if the author points to a URN, this has a set of keys associated with it. Now, if there's also a signature trailer, one can verify the signature from the set of keys. Would this work for verification?

Also, that's specific to authorship so is there any other kind of authorization we would need to plan for?

alexjg: The issue is allowing collaborative object designers to express the
authorization constraints in a way which is general enough to accomodate any
automerge document. We could hard code an authorization rule about issue comment
URNs, but then we are doing something very application specific. We would end up
writing a rule something like "if a change modifies a comment (identified by a
path into the JSON document) then ensure that the change is signed by the user
who made the comment and that the URN on the comment is the URN of the user
making the change".

There are a bunch of such rules you might want for issues, but very different
rules for things like kanban boards, and we don't want any of those rules to be
in `radicle-link`.

This is the work I want to address in the next step, which is a sort of
combination of authorization and schema validation (these being the same problem
in a decentralised CRDT). I'm currently thinking that we include a datalog with
constraints based authorization DSL which is part of the first commit of the
object and might look something like this:

```
creator says that X can insert into <comments>
where X is a user

creator says that Y.urn must be urn(X)
where Y isChild(<comments>)
and X isAuthor(Y)
```

This is some kind of logic over paths and URNs, which currently is very fuzzy
but I think it's doable. It's inspired by https://ieeexplore.ieee.org/abstract/document/4271637

PR #666, opened 2021-05-12
radicle-link-test: creation of centralised test crate

### LINKED ISSUE #649: Move testing to single package
### Motivation

As we separate out packages in the workspace, it becomes harder to re-use testing code as it's generally hidden behind `cfg(test)` and so it can't be re-used from `librad` in another package. 

We would like to consolidate all testing into a single package in the workspace. This package would essentially be the home of all tests (unit + integration), it would also include any support code for testing, e.g. `Strategy` generators for `proptest`. 

### Plan

All library code will add the following line to their `Cargo.toml` files:
```toml
[lib]
test = false
```

Since `librad-test` already contains support code for testing, we will use it as the base for moving all the testing code, and rename it to `radicle-link-test`. The following conventions will be used for moving tests over.

#### Unit tests

When moving a set of unit tests over we will mimic the module path of the original path, but prefixed with the `unit` path. For example, if we are moving the tests in `librad/src/git/fetch/specs.rs` then the path for the tests will live in `radicle-link-test/src/unit/librad/git/fetch/specs.rs`.

The `unit.rs` will then declare all the sub-modules, and in turn, `radicle-link/src/lib.rs` will contain the following line:
```rust
mod unit;
```

#### Integration tests

When moving a set of integration tests over we place them under the `integration` path, i.e. `radicle-link-test/src/integration/*`. For example, if we are moving `librad/tests/smoke/clone.rs` then the new path will be `radicle-link-test/src/integration/smoke/clone.rs`.

The `integration.rs` will then declare all the sub-modules, and in turn, `radicle-link/src/lib.rs` will contain the following line:
```rust
mod integration;
```

#### Support code

The support code will mimic the path of where it lives in its respective library, prefixed by the library's name. For example, when moving the `gen_public_key` and `gen_secret_key` function from `librad/src/keys.rs` they will go into `radicle-linke/src/librad/keys.rs`. Any extra support code can also live in the module, for example, the introduction of fast seed `SecretKey`.

### Pros

* We will have one place for testing and support code can be re-used across the tests of several packages.
* Compile times should improve for packages and the test package, as mentioned in this [article][delete-it].

### Cons

* Due to circular dependencies this change needs to be done in one-go, otherwise, the compiler gets confused about where the types come from

### Open Questions

* How do we evaluate if the API needs to be changed? For example, the `Delegations` trait is `sealed` and the generators introduce `SomeDelegations` which implements this trait. Should it be unsealed and made public? Are there other cases where we move private to public types/traits?

### References

* [Delete Cargo Integration Tests][delete-it]

[delete-it]: https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html

FintanH:
Closes #649

radicle-link-test becomes the defacto crate for testing code. The
following changes are made:

Moving librad-test to radicle-link-test.
Moving librad/tests, git-helpers/test to radicle-link-test/src/integration.
Moving all librad, git-ext, and git-helpers unit tests to
radicle-link-test/src/unit.
Moving all librad property tests to radicle-link-test/src/properties.
Moving all (relevant) librad test support code into
radicle-link-test/src/librad.


FintanH:
Todo:
- [ ] I've left out `daemon` so far don't want to create conflicts with the work @kim is doing
- [ ] Remove dev-dependencies
- [ ] Audit for changes to tests in `master`

Also holy shit, just noticed this is #666, very ominous...

kim:
I am wondering if we should go for `tests/{unit,integration,properties}`, or stick the support code into `lib`, `support`, or something.

FintanH:
> I am wondering if we should go for tests/{unit,integration,properties}, or stick the support code into lib, support, or something.

Well so far I've been sticking them in their mirrored hierarchies. So for example, here's the [`urn`](https://github.com/radicle-dev/radicle-link/pull/666/files#diff-10cd46f917380585ed7083224b0265467f831541db593b7c471f8dffd2de5070) support code. 

Or are you thinking `lib`, `support` for this kind of harness code?

kim:
Yeah just a moment of being puzzled after an initial `ls` and wondering if I should move the daemon tests into `src/daemon` or `src/integration/daemon`. I think either one of these would work:

```
src
|-- support
|   |-- <crate1>
|   `-- <crate2>
|-- integration
|-- properties
`-- unit
```

or

```
src
|-- <crate1>
|-- <crate2>
`-- test
     |-- integration
     |-- properties
     `-- unit
```

Maybe the second one is better, because the support code may not be tied to a particular crate in the longer run.

FintanH:
My preference would be for option 2. Seems clear and also easier to handle the change :smile:

radicle-link-test/src/integration/git_helpers/remote.rs
```diff
@@ -130,13 +131,18 @@ fn credential_helper() -> String {
 }
 
 fn setup_path() -> anyhow::Result<PathBuf> {
-    let helper_path = env!("CARGO_BIN_EXE_git-remote-rad");
-    let helper_path = Path::new(helper_path).parent().unwrap();
+    // The git-remote-rad binary is located under radicle-link/target/debug
```
kim, line 134: This seems to work only _sometimes_ XD

I think we should just make the test crate build the remote helper binary, so cargo can find it just as before.

FintanH: Boooooo :(

I tried that, and I ran into some issues. I didn't keep track of what they were so I'll give it another shot.

FintanH: Do you know if there was anything special about how you triggered this?

And also I don't think we can use it as before because that's specifically for integration tests:

> CARGO_BIN_EXE_<name> — The absolute path to a binary target's executable. This is only set when building an integration test or benchmark. This may be used with the env macro to find the executable to run for testing purposes. The <name> is the name of the binary target, exactly as-is. For example, CARGO_BIN_EXE_my-program for a binary named my-program. Binaries are automatically built when the test is built, unless the binary has required features that are not enabled.

kim: > integration test or benchmark

Ah dang

kim: Oh I see why it didn't work: one has to invoke `cargo build` before `cargo test`, because the latter won't make sure the binary was built before.

FintanH: I thought "Cool, I'll just add the `[[bin]]` to the toml and `cargo test` should build it too", but nope :(

FintanH: Ok, so hear me out... What about:

```rust
    if !helper_path.join("git-remote-rad").exists() {
        std::process::Command::new("cargo")
            .args(&["build", "--bin", "git-remote-rad"])
            .output()?;
    }
```

FintanH: Turns out that that doesn't work either :grimacing:

FintanH: :roller_coaster: Oh wait it does! If we add it to the `Cargo.toml` _and_ do this it will build the binary if it's missing

kim:
Just one more nit: do we want to call the directory `radicle-link-test` while we
omitted `radicle(-link)*` so far?

kim (approved):
chipeet

PR #668, opened 2021-05-14
net: avoid error log when endpoint is shutting down

kim:
We log error conditions explicitly, and the auto-error just adds noise
when shutting down gracefully.


xla (approved):
🖲 🙉 🕥 ⚔

FintanH (approved):
:relieved:

PR #669, opened 2021-05-14
Point everything at the patched git2

alexjg:
The patched git2 can be found in radicle-dev/git2-rs. That repo in turn uses radicle-div/libgit via a submodule in it's `libgit2-sys/libgit` folder.

I've tested running this with a couple of upstreams and a seed built against it and the `rere` problem dissappears.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,3 +12,12 @@ members = [
   "seed",
   "std-ext",
 ]
+
+[patch.crates-io.git2]
+git = "https://github.com/radicle-dev/git2-rs.git"
+rev = "5971ca18d73eed39384193899177dfb7ac2d0a05"
+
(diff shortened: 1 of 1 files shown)


copy/Cargo.toml
```diff
@@ -17,7 +17,8 @@ thiserror = "1.0"
 tracing = "0.1"
 
 [dependencies.git2]
-version = ">= 0.13.12, 0.13"
+git = "https://github.com/radicle-dev/git2-rs.git"
```
FintanH, line 20: It might be cleaner to add a patch to the workspace toml and leave these files as-is

FintanH (dismissed):

xla (approved):
🖲 💺 ⌚️ ✅

PR #670, opened 2021-05-14
testing: shutdown runtime with timeout

kim:
The nested runtime for the git p2p transport appears to cause the global
runtime to get stuck on shutdown for not entirely clear reasons (some
`thread::unpark` seems to get lost). Adding a small timeout to keep
runaway tasks running seems to fix that, while still preventing blocking
tasks to access already-dropped temporary directories.

### DIFF
--- a/daemon/src/peer/announcement.rs
+++ b/daemon/src/peer/announcement.rs
@@ -163,7 +163,6 @@ fn save(store: &kv::Store, updates: Updates) -> Result<(), Error> {
     bucket.set(KEY_NAME, kv::Json(updates)).map_err(Error::from)
 }
 
-#[allow(clippy::panic)]
 #[cfg(test)]
 mod test {
     use std::{collections::HashSet, convert::TryFrom as _};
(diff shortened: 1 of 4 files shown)


FintanH (commented):
I believe this makes sense to me.

I ran it locally and didn't get the hanging I usually see, but I did get an error in the chaotic chimpanzee test:
```
---- test::integration::librad::smoke::fetch_limit::replication_does_not_exceed_limit stdout ----
thread 'test::integration::librad::smoke::fetch_limit::replication_does_not_exceed_limit' panicked at 'called `Result::unwrap()` on an `Err` value: reference 'refs/namespaces/hnrkkye58a511mz77mfmifzs4ogwyrqo9rs1o/refs/remotes/hyy6eyokbh5tax45yogykqwfoxupzgfhkux9b911t78mnibdojqt8k/rad/id' not found; class=Reference (4); code=NotFound (-3)', radicle-link-test/src/test/integration/librad/smoke/fetch_limit.rs:56:42
```

Does anyone else see this? Doesn't seem to show up on CI

kim:
Yeah there’s something else being racy — you’ll also occasionally see an
error where the signed refs commit (presumably) either doesn’t have a
parent where it expects one, or has one where it doesn’t expect it. On the
up side, this is now finally easier to see :)

FintanH (approved):
:chart_with_upwards_trend:

kim:
Not sure what's going on with the lint failure -- local clippy is happy with that, and I can't seem to find an incantation which makes it pass.

PR #676, opened 2021-05-18
fix git2-rs patch not getting used

kim:
Cargo patch won't use the patched version if it can satisfy the
dependency in other ways, e.g. via a newer version of the crate. Due to
the wonders of caret dependencies, we thus need to pin git2-rs to the
version used in the patch override.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -204,4 +204,5 @@ allow-registry = ["https://github.com/rust-lang/crates.io-index"]
 # List of URLs for allowed Git repositories
 allow-git = [
     "https://github.com/ZcashFoundation/ed25519-zebra",
+    "https://github.com/radicle-dev/git2-rs",
 ]
--- a/librad/Cargo.toml
+++ b/librad/Cargo.toml
@@ -80,7 +80,7 @@ version = "0.4"
 features = []
 
 [dependencies.git2]
-version = ">= 0.13.12, 0.13"
+version = "=0.13.18"
 default-features = false
 features = []
 


kim:
I don't see the cargo patch applied in bins nor upstream, is that deliberate? /cc @alexjg

alexjg:
Not exactly deliberate. I was waiting to merge the changes in link and then forgot to put the changes in to upstream and bins :grimacing: . Will sort that now.

FintanH (commented):
Is it safe to just update `librad`'s `git2` version and not the others? :thinking:

kim:
It is _probably_ safe to assume cargo will pick the most restrictive bound _iff_ all bounds are semver-compatible. This is _most likely_ a footgun due to the PoW rule (git2-rs’ version starts with a zero).

I’m thinking, however, that we might also get away with pinning the version for `librad` (that is, not use the cargo patch but a git dep) and live with duplicate versions in applications.

FintanH:
> It is probably safe to assume cargo will pick the most restrictive bound iff all bounds are semver-compatible. This is most likely a footgun due to the PoW rule (git2-rs’ version starts with a zero).

Ya, fair enough. That sounds right.

> I’m thinking, however, that we might also get away with pinning the version for librad (that is, not use the cargo patch but a git dep) and live with duplicate versions in applications.

:cowboy_hat_face:

FintanH (approved):

alexjg:
I'll switch this over to use the git dep then, that would also mean we don't need to do anything in downstream crates right?

kim:
> switch this over to use the git dep then

I just tried this and the following happened: cargo treats a git dep as both the
same crate and not the same crate!

That is, dependency resolution fails if there is some other dependency on the
same crate, but from a different repository (that is, crates.io). Presumably,
that's because it doesn't honor the version of the git dep, and it's not
possible to force it to a version by stating it in the dependencies stanza.

Iow, we'd need to update all direct dependencies on git2 in _this_ workspace to
use the git rev, and then upstream still fails to resolve because radicle-surf
has a normal version constraint.

What a clusterf**k.

alexjg:
Ouch. So to get this working all downstreams will need to add the git dependency as well?

kim:
Yeah pretty much. I think the cargo patch approach is probably better (only two places to update), and then hope we can move on to gitoxide quickly.

alexjg (approved):

Issue #678, opened 2021-05-25
Connection rate limit needed for reliability


viraptor:
At some point recently, a number of seed nodes have been spawned in the 188.120.224.0/20 (TheFirst-RU) range, which behave more aggressively with connections than other seeds. The result has been a flood of "new ingress stream" at around 1 per ms. This caused my seed to enter a loop of overloading and crashing.

Since there's no network-visible markers for new streams, I could not filter out that traffic with a firewall. (had to drop that range completely)
It looks like the rate-limiting needs to be implemented in radicle itself to keep it stable.

PR #679, opened 2021-05-26
RFC: Connection Sentinels

alexjg:
A small RFC positing an interface for applications to decide whether a
connection or stream should be accepted.

Signed-off-by: Alex Good

### DIFF
--- a/docs/rfc/0004-rate-limiting.md
+++ b/docs/rfc/0004-rate-limiting.md
@@ -1,0 +1,68 @@
+# RFC: Connection Sentinels
+
+* Author: @alexjg
+* Date: 2021-05-25
+* Status: draft
+
+## Motivation
+
+We have observed problems when running the seed wherein some misbehaving peers
+create many incoming connections to the seed node consuming all the CPU on the
+node and causing it to become unresponsive. There are multiple solutions to
+this - rate limiting, allow/deny lists, more complex reputation systems etc.
+To support these solutions `librad` must expose a mechanism for determining 
+whether to accept a connection.
+
+## What to filter
+
+The librad network stack runs several different services, each of these has
+different performance characteristics and is likely to warrant different
+filtering policies. On that basis we expose the following interface for
+applications to decide whether a connection or a stream should be accepted
+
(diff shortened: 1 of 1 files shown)


kim:
I suppose the idea here is to allow an impl of `ConnectionSentinel` to perform
its own bookeeping. But then, it would also need to be notified of close events,
which is not entirely practical because:

a. those objects might be closed by simply getting dropped
b. finalisation (eg. transmitting outstanding data) may, however, occur later
   (via waking)
c. we need to account for both incoming and outgoing connections / streams, and
   potentially treat them differently when computing a decision

I would suggest to instead confine bookeeping to the library entirely, and pass
a reference to some kind of statistics structure to the decision making
function. Also note that we _allow_ bidirectional streams to be used for
gossip/membership, although we initiate only unidirectional ones. A policy
decision for a large number of short-lived uni streams opened in rapid
succession is rather different to compute than a weighted decision over
longer-lived concurrent streams (which is, most likely, why we're seeing
`max_concurrent_uni_streams` to not have the desired effect).

So in summary, while abstracting over this concern would be useful in the longer
term, I think that we should first have an _implementation_ of a policy which
proves to work in order to be able to tell what inputs (which are, necessarily,
internals) it requires.

To solve the issue at hand (#678), I'd suggest to first expose some transport
parameters (there's a TODO for that), and lower `max_concurrent_bidi_streams` to
something like 5 (instead of the default 100). Gossip uni streams can simply be
dropped when they hit a rate limit (we employ `governor` already), while
membership messages (uni streams) need to result in a DISCONNECT. On top of
that, a simple "ban list" of `PeerId`s may also prove effective.

alexjg:
I intentionally left out stream close events for exactly these reasons. This seems still workable to me as rate limits over open events would probably solve the problems we are observing.

The main reason to expose this interface however, was based on the idea that rate limiting might more naturally live within the `daemon` codebase, which doesn't have access to all these internals. The approach you suggest of implementing this stuff directly in `librad` and then later exposing it makes sense to me. Closing on that basis.

Issue #461, opened 2020-12-02
storage: bring back read-only access


kim:
Not currently supported in `next`, due to signer trait object and ensuring that
the keys don't diverge.

PR #682, opened 2021-05-27
RFC: Application Architecture

kim:
RFC #673 starts out by describing various functionality in terms of "capability"
traits". While a worthwhile exercise for mapping out "core" functionality and
defining surface APIs, it is difficult to see how the RFC can reach its stated
goal of "[allowing to] compose them easily, allowing upstream consumers to
mix-and-match them in any way they desire" without committing to a concrete
application architecture.

Specifically, it does not consider that different applications can have (very)
different lifecycles, yet will need certain ways to communicate with each other
(eg. to be notified of certain changes). It also doesn't talk about key
management and authentication, process management, and the distinction between
"online" and "offline" operations (recall that only one instance per device key
of a peer-to-peer stack is allowed to run at any given time). Somewhat worrying
is the implicit tendency to model the system as a single, monolithic server,
which is in diametrical opposition to the stated modularity goal.

This proposal aims to fill those gaps, and answer some of the more tricky
questions raised in discussions so far. It deliberately leverages platform
features, which naturally narrows down the scope of what the core team may
support. This is also a matter of focus: we cannot expect others to build on
library-level modules, if we haven't shown how to compose them in one particular
way.

Constraints liberate: this proposal does not aim to render #673 obsolete, but on
the contrary allow it to focus on what it started out with: defining core APIs
and module boundaries.

~[Rendered](https://github.com/radicle-dev/radicle-link/blob/kim/apparch/docs/rfc/0000-application-architecture.md)~

**Edit by Fintan**: 
The rendered link above pointed to the old markdown document. Here's the new rendered link:
[Rendered](https://github.com/radicle-dev/radicle-link/blob/kim/apparch/docs/rfc/0682-application-architecture.adoc)

### DIFF
--- a/docs/rfc/0682-application-architecture.adoc
+++ b/docs/rfc/0682-application-architecture.adoc
@@ -1,0 +1,470 @@
+= RFC: Application Architecture
+:author: @kim
+:revdate: 2021-05-24
+:revremark: accepted
+:toc:
+:toc-placement: preamble
+
+* Author: {author}
+* Date: {revdate}
+* Status: {revremark}
+* Community discussion: n/a
+* Tracking issue: n/a
+
+== Motivation
+
+Historically, development of applications on top of the libraries provided by
+`radicle-link` has been driven by the <<radicle-upstream,`radicle-upstream`>>
+application. This led to the following unpleasant situation:
+
+- Installation of components and process management is owned by a GUI
+  application, while operating on shared persistent state
+- Functionality is exposed over a TCP socket, which raises some security
+  concerns (see <<_browser_applications,Browser Applications>>)
+- Interoperability with other `radicle-link` applications is not defined,
+  respectively tied to the GUI-based process management
+- Key management is implemented as an afterthought
+- It is, more generally, unclear how a monolithic architecture can support
+  independent application development
+
+Going forward, we aim to diversify application development, and thus require a
+baseline architecture of how different `radicle-link` applications can interact
+with each other.
+
+== Overview
+
+Instead of a monolithic architecture, we admit the presence of distinct
+application concerns, lifecycles, and security constraints. Consequently, we lay
+out a service-oriented architecture, which builds on platform-specific
+primitives for process management, IPC, and credentials management. Service-
+(and thus: network-) boundaries shall be defined mainly according to the
+lifecycle of the application.
+
+Note that this matters only for a usage scenario where all installed
+`radicle-link` applications share the same user-scoped state. An alternative
+strategy would be to scope all state to the application, ie. any two
+applications would not share the same storage nor _device key_. Since this
+raises the question of how those different states are synchronised in order to
+provide a coherent view to the user, we haven't explored this approach further,
+but admit that some clever engineering on the storage layer could enable it in
+the future.
+
+We dismiss the approach of providing a local monolithic server which encompasses
+all required functionality for modularity and extensibility reasons.
+
+== Terminology and Conventions
+
+The key words "`MUST`", "`MUST NOT`", "`REQUIRED`", "`SHALL`", "`SHALL NOT`",
+"`SHOULD`", "`SHOULD NOT`", "`RECOMMENDED`", "`NOT RECOMMENDED`", "`MAY`", and
+"`OPTIONAL`" in this document are to be interpreted as described in <<RFC2119>>
+and <<RFC8174>> when, and only when, they appear in all capitals, as shown here.
+
+CBOR <<RFC8949>> datatype definitions are given using the notation devised in
+CDDL <<RFC8610>>. By convention, `struct`-like datatypes are encoded as CBOR
+maps, where the map key is the zero-based numeric index of the field in
+declaration order.
+
+== Prerequisites
+
+=== Platforms
+
+For this discussion, we divide platform targets into tiers:
+
+
+Tier 1::
++
+Fully supported platform, receives testing, deployment, and system
+configurations (may include packaging)
+
+  - Linux (`x86_64-unknown-linux-gnu`)
+  - macOS (`x86_64-apple-darwin`)
+
+Tier 2::
++
+Basic support ("it compiles")
+
+ - Windows
+ - Other processor architectures for tier 1 platforms
+
+Tier 3::
++
+Not supported, but potentially worth considering when assessing architecture
+decisions
+
+ - Mobile (iOS, Android)
+ - Wasm
+
+=== System Services
+
+We will rely on declarative service / process management to be provided by the
+platform, specifically "socket activation".
+
+On macOS, this is provided via <<launchd>>. Linux distributions considered as tier
+1 shall, however, be limited to those based on <<systemd>> for the time being.
+
+We also assume the presence of an agent capable of producing Ed25519 signatures
+utilising the SSH agent protocol <<miller>>.
+
+=== Signatures
+
+Each process which needs to produce signatures using the `radicle-link` _device
+key_, SHALL delegate this to an agent conforming to the SSH agent protocol
+<<miller, section 4.5>>.
+
+*Key access restrictions* thus depend on user configuration: the agent could
+be configured to prompt for a passphrase after a timeout elapses (or never),
+require presence of a hardware security token, or utilise platform
+authenticators such as TouchID&trade;. Note that this presents a challenge: some
+processes will reasonably require unattended access to the key (notably
+<<_peer_to_peer_node,peer-to-peer nodes>>), while interactive use may benefit
+from a user presence confirmation. This cannot be solved without restricting the
+user's access to the key material, eg. by requiring a second-factor token. As we
+can't assume widespread use of hardware tokens yet, we defer addressing the
+issue to a future proposal.
+
+We RECOMMEND to provide tooling for importing key material into the agent which
+sets a reasonable lifetime for the key, after which it must be re-loaded from
+disk (requiring a passphrase prompt for decryption).
+
+*Key generation* SHOULD, however, be performed using `radicle-link`-supplied
+tooling building on the <<ed25519-zebra>> library, which prevents certain weaker
+keys from being generated.
+
+*Signature verification* MUST be performed using the <<ed25519-zebra>> library,
+and thus MUST NOT be delegated to the agent.
+
+Note that (hypothetical) applications which require a large number of signatures
+to be generated (such as data migration tools) may find `ssh-agent` throughput
+insufficient, and require the raw key material to be supplied directly. This
+means that long-term secure key storage remains a concern of the `radicle-link`
+application suite.
+
+=== Process Orchestration
+
+All daemon processes SHALL be started on-demand using the platform socket
+activation protocol. Unless otherwise noted, daemons SHALL terminate after a
+configured period of idle time (in the order of 10s of minutes). The platform
+process manager MUST NOT restart socket-activated daemon processes if they exit
+with a success status (except by another request for the socket).
+
+All daemon processes SHALL run with user privileges (typically the logged-in
+user). Further confinement (e.g. SELinux) is beyond the scope of this document.
+It is RECOMMENDED to restrict modification of service definition files to
+require super-user privileges.
+
+=== IPC
+
+All communication with daemon processes SHALL occur over UNIX domain sockets in
+`SOCK_STREAM` mode. The socket files MUST be stored in a directory only
+accessible by the logged-in user (typically `$XDG_RUNTIME_DIR`), and have
+permissions `0700` by default. Per-service socket paths are considered "well
+known".
+
+RPC calls over those sockets use CBOR <<RFC8949>> for their payload encoding. As
+incremental decoders are not available on all platforms, CBOR-encoded messages
+shall be prepended by their length in bytes, encoded as a 32-bit unsigned
+integer in network byte order.
+
+RPC messages are wrapped in either a `request` or `response` envelope structure
+as defined below:
+
+[source,cddl]
+----
+request = {
+    request-headers,
+    ? payload: bstr,
+}
+
+response = ok / error
+
+ok = {
+    response-headers,
+    ? payload: bstr,
+}
+
+error = {
+    response-headers,
+    code: uint,
+    ? message: tstr,
+}
+
+request-headers = {
+    ua: client-id,
+    rq: request-id,
+    ? token: token,
+}
+
+response-headers = {
+    rq: request-id,
+}
+
+; Unambiguous, human-readable string identifying the client application. Mainly
+; for diagnostic purposes. Example: "radicle-link-cli/v1.2+deaf"
+client-id: tstr .size (4..16)
+
+; Request identifier, chosen by the client. The responder includes the
+; client-supplied value in the response, enabling request pipelining.
+;
+; Note that streaming / multi-valued responses may include the same id in
+; several response messages.
+request-id: bstr .size (4..16)
+
+; Placeholder for future one-time-token support.
+token: bstr
+----
+
+=== PubSub
+
+A number of instances exist in the system where a process (interactive or not)
+may want to react to an _event_ of some sort: the repository state has been
+updated by an explicit push or a fetch from the network, some defined object was
+created / deleted / modified, connectivity was lost or gained, etc.. The
+_sender_ of such events commonly does not need or want to be concerned with what
+recepients may or may not exist, and deal with buffering or re-delivery. In
+other words, this follows a publish-subscribe pattern.
+
+An obvious implementation choice for platforms which are tightly integrated with
+it would be <<dbus,D-Bus>>. This may, however, cause unreasonable build-time
+complexity on other platforms, while not achieving the same degree of
+integration. Since a wide variety of message brokers exist (but few of them
+intended for use in desktop environments), we leave the choice to a future
+proposal, and assume the following properties for the sake of this discussion:
+
+- Arbitrary binary payloads can be sent
+- Subscription is explicit, ie. we don't assume process activation upon matching
+  events
+- There is some basic facility to express interest in particular events (eg.
+  prefix-matching a topic name), which is evaluated by the broker
+- There is a way to apply buffering on a per-topic basis, in order to cater for
+  subscriber restarts
+
+== Architecture
+
+With the prerequisites in place, we lay out the following architecture:
+
+----
+                        +-----------+
+                        |  browser  |
+                        +--+------+-+
+                          /        \
+          +-------+  +---+----+  +--+-------+
+          |  CLI  |  | native |  | scripted |
+          +---+---+  +----+---+  +----+-----+
+              |           |           |
+              |     +-----+----+ +----+-----+
+              +-----|  libcli  |-|  CLIaaS  |
+                    +-----+----+ +-----+----+
+                          |            |
+        +-----------------+------------+--------+
+        |                 |                     |
++-------v--------+  +-----v----------+  +-------v--------+   +---------------+
+|  peer-to-peer  |  |    ssh-agent   |  |     pub-sub    |<--|      gitd     |
++----------------+  +----------------+  +----------------+   +---------------+
+                             ^                                       ^
+                             |                                       |
+                    +--------+-------+                               |
+                    |      git       |-------------------------------+
+                    +----------------+
+----
+
+The different components are descibed in more detail in the forthcoming
+sections.
+
+=== Peer-to-Peer Node
+
+Stateful processes which interact with the `radicle-link` peer-to-peer network
+come in two flavours: regular and seed nodes. Conceptually, the only difference
+between them is their lifecycle: seed nodes are "always-on", whereas regular
+nodes may shut down after a period of time in which no interactive operations
+are conducted, in order to save system resources and for security reasons.
+
+In practice, both node types are typically not deployed on the same kind of
+machine, and specifically seed nodes may want to limit interactive use to those
+needed for monitoring purposes. There is, however, no inherent reason to prevent
+both node types from being deployed on the same machine. While a regular node
+may be configured to behave like a seed node (e.g. an automatic tracking
+configuration), we RECOMMEND to treat both as separate services, each with their
+own peer id and persistent state. Whether this entails a single executable
+exposing the superset of configuration options applicable to both modes of
+operation, or two separate ones is an implementation choice.
+
+The IPC socket path for peer-to-peer nodes shall follow the following
+convention:
+
+ $XDG_RUNTIME_DIR/radicle/<srv>-<peer-id>.sock
+
+where `srv` is either "node" or "seed".
+
+The RPC API exposed follows the interface defined by ``librad``'s `net::Peer`.
+
+NOTE: This obviously excludes all `using_storage` operations, and is pending
+async-ification of the `replicate` function (which then shall be promoted to a
+`net::Peer` method).
+
+Events emitted by the node cannot be subscribed to directly over the RPC API,
+but only indirectly over the pub-sub bus. The node shall itself listen for
+pub-sub messages related to publishable updates of the repository state (eg.
+after a `git push`).
+
+=== git
+
+Interactions with (logical) git repositories managed by a peer-to-peer node
+currently relies on a git remote helper and client-side refs rewriting. This is
+mainly because of a lack of library support for rewriting refs (eg. to display
+nick names alongside peer ids) on the "server". This is somewhat inconvenient,
+as the remote helper needs to be placed in the user's PATH during installation,
+and the (client) git configuration needs to be rewritten whenever the remote
+state changes. Ideally, we should be able to use a built-in git transport to
+connect to a custom git daemon, which is started on-demand via socket
+activation.
+
+Pending some advancements in the first, we can achieve this through the
+`gitoxide` and `thrussh` crates, and using the `ssh` transport for "rad
+remotes".
+
+NOTE: The obvious drawback is that this requires to bind to a TCP socket, and
+thus breaks the desired isolation to the logged-in user. A workaround could be
+to supply the `ProxyCommand` option to ssh, which proxies the connection over a
+UNIX socket. While possible to achieve by modifying the user's git
+configuration, a more robust and flexible solution might be to provide a wrapper
+command (`rad push/pull`).
+
+The git daemon emits events to the pub-sub bus after successful pushes. Like the
+<<post-receive>> hook, this is one event per updated ref, for example:
+
+[source,cddl]
+----
+updated = (
+    oid-old: bstr .size 20,
+    oid-new: bstr .size 20,
+    namespace: bstr,
+    ref: tstr        ; relative to namespace
+)
+----
+
+The peer id the git daemon is assuming shall be encoded in the topic name.
+
+=== CLI
+
+The canonical `radicle-link` CLI shall follow the "subcommand" pattern familiar
+from git and other complex commandline applications. Like git, it shall be
+possible to extend the set of available subcommands by placing executables
+conforming to a naming convention (eg. `rad-<command>`) in the user's PATH. It
+SHALL NOT be possible to override builtin commands by this mechanism.
+
+Each subcommand MUST expose its functionality as a linkable library, and provide
+CBOR serialisation for its arguments and outputs.
+
+To enable <<_scripted_applications,scripted applications>>, subcommands may be
+callable over the <<ipc,IPC>> protocol ("CLIaaS"). The canonical CLI may provide
+a command to bind all available subcommands to a socket wholesale, or a
+configurable subset of them. Unprivileged subcommands (ie. commands which do not
+modify the configuration nor state) may also be exposed as a HTTP API (see
+<<_browser_applications,Browser Applications>>).
+
+The builtin subcommands shall include network clients for interacting with local
+p2p nodes and the pub-sub bus.
+
+NOTE: It does not make much sense to proxy networked subcommands when binding to
+a socket. Frontends will need to connect to the respective services directly.
+
+A comprehensive specification of the (initial) set of builtins is beyond the
+scope of this document.
+
+=== Native Applications
+
+Native applications are assumed to link directly against the same library
+modules as the <<_cli,CLI>> ("libcli").
+
+=== Scripted Applications
+
+Under scripted applications, we subsume:
+
+. <<electron,Electron>> applications
+. Applications which can not, or do not want to link against Rust libraries
+
+To facilitate rapid prototyping, but also to mitigate the risk of remote code
+execution (RCE) / cross-site scripting (XSS) attacks especially for electron
+applications, we RECOMMEND to develop native applications in a client-server
+style, where `radicle-link` functionality is provided as <<_cli,CLI>>
+subcommands callable over an <<_ipc,IPC>> socket. Recomposition of subcommands
+into custom IPC daemons is encouraged.
+
+For electron applications specifically, we strongly RECOMMEND to follow security
+best practices. Minimally, renderer processes SHALL NOT have access to the node
+environment, and proxy backend calls through the main process (renderer-main
+communication utilises Chrome IPC, which is harder to attack than a TCP
+connection to a backend process).
+
+=== Browser Applications
+
+At this point, we discourage browser applications which allow modification of
+the `radicle-link` state for the following reasons:
+
+- For most practical purposes, browser-backend communication relies on a TCP
+  socket. Even if bound only to the loopback interface, this poses a security
+  risk due to RCE / XSS attack vectors.
+- Meaningful authentication & authorization can only be achieved using
+  second-factor authentication, which we don't consider feasible at this point
+  (see also <<_signatures,Signatures>>).
+
+Even when unprivileged, browser applications SHOULD implement an authentication
+scheme using one-time / time-restricted access tokens.
+
+== Drawbacks
+
+The main drawback of any service-oriented architecture is that it increases
+complexity by having to consider inter-service dependencies. Since all
+components are to be deployed on a single host, some of this can be mitigated by
+proper package management. During development this can however cause friction,
+especially when multiple services need to be updated in lockstep.
+
+== Future Work
+
+The main omission of this RFC is to explore platform specifics for Windows
+targets. This is mainly because the author has stopped using Windows two decades
+ago, and was never very keen to understand the platform's idiosyncrasies. He
+conjectures, however, that equivalents to the system services proposed here
+exist, and would be grateful for pointers.
+
+Another question raised during discussions which led up to this document is if
+mobile platforms should be considered as first-class targets. The author's
+stance on the topic is that only a subset of the `radicle-link` protocol is
+applicable to resource-constrained environments, and thus a protocol revision
+would be prerequisite. A traditional client-server architecture, where the
+mobile device serves as a frontend to a remote service might be a first step,
+however.
+
+As alluded to throughout the document, security rests mainly on an uncompromised
+user space: both malware running under the user's privileges as well as
+root-level compromise can, simplified, gain access to the `SSH_AUTH_SOCK`, and
+thereby compromise the application. Note that this is not specific to the
+service-oriented architecture. The underlying difficulty is one of user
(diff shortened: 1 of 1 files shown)


docs/rfc/0000-application-architecture.md
```diff
@@ -0,0 +1,503 @@
...
+shall be prepended by their length in bytes, encoded as a 32-bit unsigned
+integer in network byte order.
+
+Our security model does not include an attacker which is able to intercept
+traffic on local UNIX sockets. We also do not consider the possibility of
+substituting a service executable with a malicious one listening on the "well
+known" socket for that service. Consequently, we do not require transport-layer
+security nor verification of the service's identity.
+
+We do, however, require that client applications requesting privileged
```
kim, line 184: Admittedly, this is somewhat paranoid: as long as the service socket and `SSH_AUTH_SOCK` have the same permissions, they are both in the same “security domain”, and this would not be needed. I am, however, sure that reasons will be found to relax the service socket permissions =]

docs/rfc/0000-application-architecture.md
```diff
@@ -0,0 +1,503 @@
...
+> out to complicate builds due to varying link-time requirements. As we only
+> intend to use D-Bus for pub-sub messaging ("signals"), we may instead consider
+> to provide a purpose-built in-memory pub-sub service.
+
+
+### Signatures
+
+Each process which needs to produce signatures using the `radicle-link` _device
+key_, SHALL delegate this to an agent conforming to the [SSH agent
+protocol][ssh-agent] (section 4.5.).
```
geigerzaehler, line 132: Could you expand on how this would work in practice? How would applications add the device key to the agent? How would they know which key is the device key? Do we want to make use of [key confirmation constraints](https://datatracker.ietf.org/doc/html/draft-miller-ssh-agent-02#section-4.2.6.2)?

One thing I’m concerned about is that if one service loads the key into the agent it is available to all services running as the user. This would make RPC request signing for process redundant from a security standpoint

docs/rfc/0000-application-architecture.md
```diff
@@ -0,0 +1,503 @@
...
+1. The backend must bind to a well-known TCP port, which is then accessible by
+   all other programs on the same host. Moreover, browsers can be tricked by
+   remote sites into issuing requests against the local host.
+2. Public-key based authentication APIs (such as [WebAuthn][webauthn]) are not
+   yet mature / flexible / portable enough to allow a browser frontend to prove
+   access to the (shared) private key to a backend process.
+
+For these reasons, we devise the following restrictions:
+
+* TCP-based backend services MUST NOT expose privileged operations.
```
geigerzaehler, line 450: Does this still hold if we require privileged operations to be signed? (Of course, this doesn’t enable us to build web-based privileged applications yet. It might provide a building block together with browser extensions or WebAuthn though.)

docs/rfc/0000-application-architecture.md
```diff
@@ -0,0 +1,503 @@
...
+conforming to a naming convention (eg. `rad-<command>`) in the user's PATH. It
+SHALL NOT be possible to override builtin commands by this mechanism.
+
+Each subcommand MUST expose its functionality as a linkable library, and provide
+CBOR serialisation for its arguments and outputs.
+
+To enable [browser applications](#browser-applications), subcommands may be
+callable over the [IPC](#ipc) protocol ("CLIaaS"). The canonical CLI may provide
+a command to bind all available subcommands to a socket wholesale, or a
+configurable subset of them. Unprivileged subcommands may also be exposed as a
```
geigerzaehler, line 417: Could you explain what you mean by “unprivileged subcommands”? I assume these are sucommands that don’t issue any unprivileged requests to any of the services or the storage directly?

geigerzaehler (commented):
I like the general architecture very much.

One major upside for Upstream is that we can extend the system more easily. Instead of having to add something to a monolithic service we can write our own commands that use Rust libraries.

docs/rfc/0000-application-architecture.md
```diff
@@ -0,0 +1,503 @@
...
+
+Since preferences of web developers regarding the granularity of remote calls
+vary, they may choose to implement their own IPC backend, possibly re-composing
+the library functionality of [CLI](#cli) subcommands.
+
+Browser applications SHOULD be packaged in a way such that their main backend
+conforms to [Process Orchestration](#process-orchestration), and include other
+services via the package manager's dependency management mechanism. If that is
+not desired (e.g. [AppImage] bundles), they MUST sandbox all service
+dependencies, and MUST NOT interact with services installed outside the sandbox
```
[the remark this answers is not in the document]
kim: AppImage’s philosophy is decidedly to bundle everything in one image, and _not_ provide any dependency management. Sure you can stick an AppImage into a distribution package, but I don’t think this would be accepted in most distros.

So, if that’s the case, then the AppImage cannot assume anything about the environment it is running in — and consequently should also not modify it in any way (eg by installing things outside of its sandbox). 

Of course one can say “install packages x, y, z using your system package manager before being able to use this”, but this seems to defeat the whole idea of self-containedness.

docs/rfc/0000-application-architecture.md
```diff
@@ -0,0 +1,503 @@
...
+the time being.
+
+We also assume the presence of an agent capable of producing Ed25519 signatures
+utilising the [SSH agent protocol][ssh-agent].
+
+On macOS, the distribution `ssh-agent` integrates with the system keychain.
+Other choices include the `ssh-agent` provided by the `openssh` suite, the
+`gpg-agent`, or (on Linux systems) `gnome-keyring`.
+
+Lastly, we consider [D-Bus][dbus] for publish-subscribe messaging. D-Bus is
```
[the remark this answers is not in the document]
kim: I am myself not entirely convinced of relying on D-Bus — it is rather arcane, and has been criticised for low performance. Perhaps we should do a prototype first, also to see just how painful it is to get going on macOS.

You’re right that D-Bus can, in principle, provide socket (well, service) activation as well. It is then, however, *required* for the client to authenticate, while enforcing socket file permissions may make this unnecessary.

docs/rfc/0000-application-architecture.md
```diff
@@ -0,0 +1,503 @@
...
+from git and other complex commandline applications. Like git, it shall be
+possible to extend the set of available subcommands by placing executables
+conforming to a naming convention (eg. `rad-<command>`) in the user's PATH. It
+SHALL NOT be possible to override builtin commands by this mechanism.
+
+Each subcommand MUST expose its functionality as a linkable library, and provide
+CBOR serialisation for its arguments and outputs.
+
+To enable [browser applications](#browser-applications), subcommands may be
+callable over the [IPC](#ipc) protocol ("CLIaaS"). The canonical CLI may provide
```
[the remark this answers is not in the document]
kim: The idea would be that it is not currently supported to have “plain” browser applications (except for things which don’t require access to the secret key). Of course, the electron `main` process could also just fork CLI commands, but calling them over a pipe is more performant, and probably more convenient that building argument lists and parsing the output. Also, if the app doesn’t require any extra functionality or command wrappers, it can be deployed without any native code.


Ultimately, it’s up to the application.

docs/rfc/0000-application-architecture.md
kim: We’d provide a small tool to substitute `ssh-keygen`, which prompts for a passphrase, stores the key somewhere, and then imports it into `ssh-agent` (the protocol supports that). The gpg agent requires more configuration (of course), but I guess that’s for advanced users anyway.

Requesting a signature from the agent requires the public key and the payload to sign as parameters — so the application needs to store the public key somewhere.

docs/rfc/0000-application-architecture.md
kim: No, if we had a way to sign “transactions” in the browser, this would work.

docs/rfc/0000-application-architecture.md
kim: In the email notification you mention key confirmation constraints: I couldn’t really get those to work using `thrussh`, so that would require some experimentation. Also note that other agents which happen to support the agent protocol may not respect those at all (eg gpg agent only accepts manual configuration).

Ideally, the configuration should be such that not _every_ signature request requires pin entry — otherwise one would enter it twice for every command of authentication is in place. Idk... “requires further research”

docs/rfc/0000-application-architecture.md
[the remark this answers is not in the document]
kim: Perhaps the main problem with the current architecture is that upstream manages its own p2p stack — there should only be one instance per key, so if there are other apps on the same machine, they either need to talk to upstream, or use their own key.

Offline operations are less problematic — but then, if your AppImage _assumes_ the networking service to be present, yet writes to the storage directly, I’d expect this to result in a mess to install and configure.

docs/rfc/0000-application-architecture.md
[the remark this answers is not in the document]
kim: It’s a user-space application

docs/rfc/0000-application-architecture.md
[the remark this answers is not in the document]
kim: Here’s another approach: say we use lock files to prevent multiple instances of the network stack. Being a good app bundle, the app bundles all executables it needs, implements its own process management etc. Being a good citizen in the radicle app ecosystem, its interfaces to the outside world are the `ssh-agent` and DBus (because it wants to know about pushes). 

This would mean DBus is a hard requirement. It would also mean that we can’t use DBus to activate services, because which one will respond? If we decide that DBus is not the right fit, because it’s too slow, not portable enough or whatever, we’re back to square one because that component must be made available to all radicle apps.

So idk, but self-contained app bundles just don’t seem to be a good fit if we’re aiming for a diverse ecosystem (unless they are truly self-contained that is).

docs/rfc/0000-application-architecture.md
```diff
@@ -0,0 +1,503 @@
...
+management is up to the application, and separate key material must be used).
+
+
+## Future Work
+
+The main omission of this RFC is to explore platform specifics for Windows
+targets. This is mainly because the author has stopped using Windows two decades
+ago, and was never very keen to understand the platform's idiosyncrasies. He
+conjectures, however, that equivalents to the system services proposed here
+exist, and would be grateful for pointers.
```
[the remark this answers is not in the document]
kim: The blessed way seems to be via PowerShell: https://docs.microsoft.com/en-us/windows-server/administration/openssh/openssh_keymanagement

The strange thing about windows is that there are so many different ways to avoid having to use their platform idiosyncrasies, and instead just use UNIX tools: Cygwin, MSYS2, MinGW, WSL2, ...

I can’t even tell what the differences are, nor what developers typically use / have installed, or if that matters at all. Maybe these days PowerShell is so powerful that it’s kind of the emacs of windows and one can do anything we’d want on top of it?

docs/rfc/0000-application-architecture.md
kim: I looked into WebAuthn for this a bit, btw, and it makes some ... interesting choices (well, not _that_ surprising given that it’s driven by Yubico to no small extent):

* A “platform authenticator” (TouchID, Windows Hello) is not considered first-class (it does not identify the user), instead, the flows are described such that one first authorises the device using a “roaming authenticator” (USB fob), and only subsequently one may use the platform authenticator. That obviously doesn’t make much sense in a serverless environment.
* Although the spec describes a scenario where a mobile phone is used as a second factor, this doesn’t seem to exist in the wild, let alone in some generic form.
* In general how WebAuthn works is by creating a fresh keypair for each site, which is attested by a hardware key. It does not seem possible to somehow store an existing key with the current browser APIs. I’m not sure what to do with this, given that we’re on the local host: somehow, the hardware key would need to be authorised by the link device key, such that a signature by a key signed by the hardware key is considered eligible. However, the backend still needs to get access to the actual device key!


While I find the idea entertaining that it should only be possible to use radicle link with a hardware token (great for sybil prevention), I get the impression that this is all a bit too convoluted for our purposes.

docs/rfc/0000-application-architecture.md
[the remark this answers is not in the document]
geigerzaehler: > Sure you can stick an AppImage into a distribution package, but I don’t think this would be accepted in most distros.

This was exactly what I was getting at. I wanted to know whether there are any reasons not to use AppImage even if we distribute it via package managers and don’t support it standalone. The reasoning behind this is that I wasn’t sure how easy it is to bundle electron apps without relying on AppImage. (Spoiler: it’s easy)

In general I agree that we should distribute applications so that they integrate naturally with other Radicle services. The best way to accomplish that is indeed package managers and service managers.

docs/rfc/0000-application-architecture.md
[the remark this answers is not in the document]
geigerzaehler: > You’re right that D-Bus can, in principle, provide socket (well, service) activation as well. It is then, however, _required_ for the client to authenticate, while enforcing socket file permissions may make this unnecessary.

As far as I know most D-Bus installations allow the `EXTERNAL` [authentication mechanism](https://dbus.freedesktop.org/doc/dbus-specification.html#auth-mechanisms) that does use the socket to authenticate the user.

> I am myself not entirely convinced of relying on D-Bus — it is rather arcane, and has been criticised for low performance.

Alternatively, we could drop D-Bus altogether and also use service-to-service pub-sub over RPC. My main concern is about splitting this up and . If we rely on D-Bus we will have to deal with it’s arcana and I’d rather go all-in or drop it alltogether.

docs/rfc/0000-application-architecture.md
geigerzaehler: Thanks. Might be a good idea to add this clarification to the document.

docs/rfc/0000-application-architecture.md
```diff
@@ -0,0 +1,503 @@
...
+Note that (hypothetical) applications which require a large number of signatures
+to be generated (such as data migration tools) may find `ssh-agent` throughput
+insufficient, and require the raw key material to be supplied directly. This
+means that long-term secure key storage remains a concern of the `radicle-link`
+application suite.
+
+
+### Process Orchestration
+
+All daemon processes SHALL be started on-demand using the platform socket
```
FintanH, line 155: I think I'm left a bit confused as to what the "inner" architecture is. Is there some kind of "main" process that can spawn sub-processes, i.e. "daemon processes", worded here? Or when you say "daemon processes" do you mean multiple applications spinning up daemons?

docs/rfc/0000-application-architecture.md
```diff
@@ -0,0 +1,503 @@
...
+
+### IPC
+
+All communication with daemon processes SHALL occur over UNIX domain sockets in
+`SOCK_STREAM` mode. The socket files MUST be stored in a directory only
+accessible by the logged-in user (typically `$XDG_RUNTIME_DIR`), and have
+permissions `0700` by default. Per-service socket paths are considered "well
+known".
+
+RPC calls over those sockets use [CBOR] for their payload encoding. As
```
FintanH, line 173: Am I right in thinking that this RFC is not specifying what we're going to use as an RPC?

docs/rfc/0000-application-architecture.md
```diff
@@ -0,0 +1,503 @@
...
+                                   |                |                                      |                |
+                                   +----------------+                                      +----------------+
+```
+
+The boxes in the middle form our "service layer", while the outer boxes
+represent interactive applications. The different components are descibed in
+more detail in the forthcoming sections.
+
+
+### Storage
```
FintanH, line 307: Am I right in thinking that implementing #461 means that a seed node won't need access to a key?

docs/rfc/0000-application-architecture.md
```diff
@@ -0,0 +1,503 @@
...
+All daemon processes SHALL run with user privileges (typically the logged-in
+user). Further confinement (e.g. SELinux) is beyond the scope of this document.
+
+
+### IPC
+
+All communication with daemon processes SHALL occur over UNIX domain sockets in
+`SOCK_STREAM` mode. The socket files MUST be stored in a directory only
+accessible by the logged-in user (typically `$XDG_RUNTIME_DIR`), and have
+permissions `0700` by default. Per-service socket paths are considered "well
```
FintanH, line 170: Can you give an example of what a service might be?

docs/rfc/0000-application-architecture.md
```diff
@@ -0,0 +1,503 @@
...
+
+The receiver of an RPC with a non-null `signature` header MUST verify the
+`signature` against its own public key. If signature verification fails, it
+responds with a `forbidden` error, regardless of whether the RPC actually
+requires authentication as per the application semantics.
+
+
+## Architecture
+
+With the prerequisites in place, we lay out the following architecture:
```
FintanH, line 254: The `PubSub` component isn't connected to the `peer-to-peer` component. Does that mean it's not going to be used for events as part of the p2p stack?

docs/rfc/0000-application-architecture.md
kim: > Is there some kind of "main" process that can spawn sub-processes

`systemd`, `launchd`

docs/rfc/0000-application-architecture.md
kim: No, it is pretty explicit: you send `<u32><CBOR payload>`

docs/rfc/0000-application-architecture.md
kim: No, a peer-to-peer networked program requires access to the key for transport-layer security and refs signing.

docs/rfc/0000-application-architecture.md
kim: "Peer-to-peer Node", "CLIaaS"

docs/rfc/0000-application-architecture.md
kim: Right... ASCII diagrams on lenient 80 columns are hard

docs/rfc/0000-application-architecture.md
kim: No, it is pretty explicit: you send `<u32><CBOR payload>`

kim: And get back `<u32><CBOR payload>`.

kim: What `CBOR payload` is is given in IPC. What `payload` of `CBOR payload` is, may be defined in #673, or a subsequent document.

docs/rfc/0000-application-architecture.md
```diff
@@ -0,0 +1,506 @@
...
++--------+--------+    +-----------------+    +--------+--------+    +--------+--------+   +-------+--------+
+         |                                             |         \            |           /        |
+         |                                 +-----------+          +----------------------+         |
+         |                                 |                                  |           \        |
+         |                         +-------+--------+                         |            +-------+--------+
+         |                         |                |                         |            |                |
+         +-------------------------+      CLI       +-------------------------+            |      git       |
+                                   |                |                                      |                |
+                                   +----------------+                                      +----------------+
+```
```
yordis: It would be super helpful to have a brief description of each box and what they suppose to do. Trying to understand each component and the reason for isolation.

docs/rfc/0000-application-architecture.md
FintanH: Right, didn't put 2 and 2 together. So that's the message format and the transport is the socket.

docs/rfc/0000-application-architecture.md
FintanH: So is this RFC punting on how a seed might handle that? Or would it be something like setting the `ssh-agent` to only ask once?

docs/rfc/0000-application-architecture.md
kim: The components are described in the paragraphs following this masterpiece -- can I make this clearer in some way?

docs/rfc/0000-application-architecture.md
kim: I scrapped the whole idea mostly, just leaving an optional field for request tokens in the IPC protocol. I'm pretty much convinced that we would require an actual second-factor. As a stop-gap, we can expire the key and then prompt for the passphrase every so often.

docs/rfc/0000-application-architecture.md
kim: 9540500a167b0d76fe26735882cc0ea9de0258c8

docs/rfc/0000-application-architecture.md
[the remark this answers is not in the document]
kim: I have a distinction between "native apps" (including electron), and "browser apps". Hope that makes it clearer.

docs/rfc/0000-application-architecture.md
yordis: I would suggest

- Using the exact same name in the headlines, without renaming or enhancing the text.
- Have a simple list before the diagram where people could read the name (pointing to the bigger content that follow) and a really short description
- Include all the component names some where (even if the description is short)

Take my feedback with a grain of salt, this is my perspective as somebody that has less than 2 hours of reading stuff (so newbie perspective)

docs/rfc/0000-application-architecture.md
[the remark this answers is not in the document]
kim: > the EXTERNAL authentication mechanism 

Oh my, that looks like a pain in the neck.

> Alternatively, we could drop D-Bus altogether

Ya, maybe that's better. After all, we only need a small ringbuffer (possibly scoped by topic), and not necessarily any of the advanced functionality. There are two reasons why D-Bus could be attractive, but _possibly_ this only applies to platforms which actually integrate with it (ie. not the proprietary ones):

1. We may want processes not owned by the current session to react to events (eg. post-fetch hooks a la Fleek)
2. We may want a way to dispatch notifications which result in some user interaction (a passphrase prompt, f.ex.): say the user has `upstream` running, it may be preferable to focus its window and draw the dialog there.  Say `upstream` is **not** currently running (or not installed at all), then can some other program handle this? I _believe_ that this is a common pattern for D-Bus-enabled applications.

Arguably, both can also be solved separately (and platform-specific), so I tend to agree that we should just roll our own pub-sub daemon.

docs/rfc/0000-application-architecture.md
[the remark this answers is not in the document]
kim: I dropped this. Perhaps we should have a separate proposal for packaging guidelines, maybe in `radicle-decisions` instead?

docs/rfc/0000-application-architecture.md
kim: Unless told explicitly, the `ssh-agent` doesn't ask for anything -- `ssh-add` may, if the key on-disk is encrypted. The agent _protocol_ (well, the single call we actually need) is simple enough so a frontend to some cloud KMS could be put together quickly. By saying that we support the protocol, we don't have to deal with any clown provider JSON abomination tho.

docs/rfc/0000-application-architecture.md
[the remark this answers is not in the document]
geigerzaehler: Sounds good to me :+1:

kim:
Team Call Notes:

1. It is useful to think of a peer-to-peer node as a separate process (daemon). Only one per device key should be running at any point in time.
2. It is not entirely clear just how much CLI (-framework) the team is willing to own
   1. If it doesn't, then who does?
   2. If it provides _some_ commands, then how do those fit into a larger framework, if at all?
   3. If anything not strictly protocol related is up to someone else, we end up with:

      1. Functionality which requires connectivity. This is not recomposable -- it's just what the protocol exposes.
      2. Plain git
      3. The `Storage` interface
        
          The storage can be generalised to: assumptions about certain refs under the `refs/rad` namespace, and datastructures those peel to. As these are assumptions by the _protocol_, one could argue that it is not even a public interface: either the protocol needs to base some behaviour on the presence or absence of data at those locations -- and therefore requires a specification -- or it doesn't. Thus, plain git provides extensibility for arbitrary applications.

    4. Assuming 1., 2.iii.a and 2.iii.b, we _still_ need to solve most of the issues addressed in the proposal:
    
         * How do we communicate with the daemon process?
         * How is the daemon process managed in the first place?
         * How is access to keys managed, and can we add stronger forms of authz later?
         * How do we dispatch _events_, where the sender might not know the recipients up front, and doesn't even care?
          
           Arguably, with only two components (daemon + git), we could simplify this to not require a standalone service, and simply provide a streaming endpoint for the daemon. The problem of _application level_ notifications doesn't go away, though, but is someone else's.
3. The author did not mention this in the document, but one solution for authz could be [macaroons](https://storage.googleapis.com/pub-tools-public-publication-data/pdf/41892.pdf)

    The issue is that macaroons rely on a shared secret: say we have a tool which loads the passphrase-encrypted key from disk and issues a macaroon, which is passed to an application which wishes to sign something with that key. The application can assert the validity of the macaroon, but it doesn't have access to the key itself (if it had, the mechanism is redundant -- just prompt for the passphrase twice). The same problem exists with [JWT](https://datatracker.ietf.org/doc/html/rfc7519) or [CWT](https://datatracker.ietf.org/doc/html/rfc8392): those could carry the private key as an encrypted payload, but the recipient needs access to the decryption key (respectively the key that decrypts the decryption key). 

    Securing localhost ist just... mindbogglingly recursive, it seems.

cloudhead:
Interesting! I have a few questions:

1. If the idea is that the daemon is "short-lived", ie. you mention it shutting down after a few minutes, does it only wake up when the local user runs a command, or does it also wake up when the external TCP socket has activity? I would hope that there is a way for applications to receive events as soon as possible, eg. a remote ref was updated, a comment was posted on an issues, etc., without requiring user action. I guess this could also be solved by configuration, eg. `daemon start --keep-alive`.

2. I'm unclear on how the CLIaaS works, ie. who is supposed to be spawning these CLI processes? Is it the daemon, or is this the application's responsibility, eg. through its own backend?

3. The PubSub system could be specified for one platform at least, as an example. Currently it's hard to imagine how that would work. Or perhaps having an idea of what a high-level API would look like, and which service/component would be serving this API.

4. When thinking about the general architecture, if I understand correctly, all "capabilities" are to be implemented as command-line tools that can input/output CBOR. Since this requires some kind of backend service to run (to be able to spawn these commands), I wonder if at that point it's not easier to directly invoke library functions? Basically, I'm trying to understand in which case(s) spawning a CLI makes more sense than calling into a library.

kim:
> If the idea is that the daemon is "short-lived", ie. you mention it
> shutting down after a few minutes, does it only wake up when the local
> user runs a command, or does it also wake up when the external TCP
> socket has activity?

In a lot of situations it is highly undesirable to have some kind of server
connected to the internet running in the background: mobile connection, VPN,
firewall, general paranoia,...

The idea is that by connecting to the "control" (UNIX) socket, the daemon is
started via socket activation. As long as there is a connection it will keep
running, when there are none, it will exit after a while. This way, the
lifecycle is tied to interactive use.

> guess this could also be solved by configuration, eg. daemon start --keep-alive

I think users should interact with the executable, that's what process
management is for. One could emulate "keep alive" by making a command which
keeps a connection to the control socket.

> I'm unclear on how the CLIaaS works, ie. who is supposed to be
> spawning these CLI processes?

Same thing: as soon as something connects to the IPC socket, the command server
gets spawned.

> The PubSub system could be specified for one platform at least, as an
> example. Currently it's hard to imagine how that would work. Or perhaps
> having an idea of what a high-level API would look like, and which
> service/component would be serving this API.

I'm not sure how to answer that. There are a bunch of events emitted by the
`net::Peer` already, which in the case of a single p2p daemon are of interest to
more than one recipient. Similarly, when you push through the git CLI.

> "capabilities"

Not sure this is good terminology anymore: what is called a capability in #673
is not divisible -- it's just the set of functions you can call on the protocol
stack.

It's more _additional_ functionality which is not protocol specific (ie.  the
protocol doesn't need to know about it).

> I wonder if at that point it's not easier to directly invoke library
> functions?

Yes, that's why those subcommands should be their own library modules. If you
don't want to link (eg. because you are using a SCRIPTING LANGUAGE), you don't
have to. For example, you can deploy an Electron app without any native code
backend, if instead you can instruct the package manager to make sure the CLI
server is installed.

FintanH:
> It is not entirely clear just how much CLI (-framework) the team is willing to own

Thanks for laying out the questions. @xla and I had a chat and we landed on the following thoughts. 

The team should own and maintain a certain subset of CLI commands. Since the goal is to dogfood our own tasty treats we should build the CLIs that we will want to use. Inspired by tiered systems, we outlined what we will want to maintain as first-class citizens of CLI-topia, followed by what the radicle-link team will have opinionated CLIs on, and finally things that we will not consider as part of our responsibilities.

### First-class

The following components are owned and maintained by the radicle-link team and are part of the core experience.

* Projects
* Persons
* Tracking
* Replication
* Daemon control
  * Things like start, stop, restart of the daemon
* Daemon stats
  * Getting p2p activity like connectivity statistics
* Key management
  * Generation of new keys for "profiles"

### Opinionated-class

The following are components that will have opinionated versions of but will not commit to any SLAs around the maintenance of them. We expect that the community will like to use our versions of the components but could also form their own versions of the components.

* Pretty much all future collab object models
  * Issues
  * Patches
  * Wikis
  * RFCs
  * whatever we can dream of
* radicle-surf/radicle-source

### No-class

We will not be providing any HTTP/TCP implementations -- in contrast to what we first thought we would do.

### Open Questions

The first open question is around the `rad-git` CLI. In this RFC you made a note:
> The obvious drawback is that this requires to bind to a TCP socket, and thus breaks the desired isolation to the logged-in user. A workaround could be to supply the ProxyCommand option to ssh, which proxies the connection over a UNIX socket. While possible to achieve by modifying the user's git configuration, a more robust and flexible solution might be to provide a wrapper command (rad push/pull).

Is the surface area simply `push/pull` or do we expect more git commands to be needed/added? If so, we could likely commit to that being part of the First-class set, if the proxy over the UNIX socket is too much of a PITA.

The second grey area is around key management. Despite chatting about it on the team call, it's still a bit unclear to both of us what kind of problems we need to solve here. 

Something we're not sure about is the validation of online requests to the peer. For example, say we ask to replicate an Urn we don't know about, will that need some verification? Is this what the [IPC](https://github.com/radicle-dev/radicle-link/blob/kim/apparch/docs/rfc/0000-application-architecture.md#ipc) section is attempting to cover?

We also don't have bright ideas of how to handle the "unlocking" of a key for use in CLI actions, e.g. creating a project. Clearly, we need to provide primitives that work in both a CLI world and a CLIaaS world.

So you're right, key management does seem to be the more interesting problem given the number of open questions about it :)

### Conclusion

We see the owning of core CLI components as a positive thing for the team as it will allow us to dogfood more easily and will be _synergistic_ with the protocol development. It will also allow us to think about how the ecosystem may be extended while also guarding important namespaces that we care about.

All that being said, we have formed two immediate goals:
1. Draft up an RFC that discusses the daemon peer and its API. This will include the expectations of what functionality we will need to provide, e.g. asking and replicating a project.
2. Draft up an RFC that discusses the CLI plan and infrastructure which will include the points outlined above.

If I missed anything that you think should be included @xla, please feel free to fill them in :)

As always, looking forward to feedback on this for improvment

kim:
> The team should own and maintain a certain subset of CLI commands.

I just don't see how this yields the coherent experience that is so highly
valued, without a framework those commands adhere to.

> The first open question is around the rad-git CLI.
[...]
> Is the surface area simply push/pull or do we expect more git commands to be
> needed/added?

The surface area is the entirety of the git suite of commands. The interesting
part here is the "server" end: it should be able to rewrite refs when fetching,
eg. to display user names (instead of the client-side rewriting), or do
Gerrit-style rewrites on push (`for/review/master`). It is also about notifying
other processes of modifications to the repo.

> The second grey area is around key management.
[...]
> Something we're not sure about is the validation of online requests to the
> peer. For example, say we ask to replicate an Urn we don't know about, will
> that need some verification?

The p2p process needs access to the key for that. Whether the client initiating
this request needs to be authenticated depends on the security model: whoever
has access to `RAD_DAEMON_SOCK` also has access to `SSH_AUTH_SOCK`. So what's
the point of authenticating again? Because there is no way to assert that this
command was initiated _interactively_ -- it could be a background process, a
malicious node module, a DNS rebinding attack on your awesome browser app.

Iow, it is desirable to have a proof of user presence for both security and
"experience" reasons (the experience to have 25GiB worth of git history
downloaded to your hdd without you having approved of that is certainly not a
quality one).


> * Draft up an RFC that discusses the daemon peer and its API. This
> will include the expectations of what functionality we will need to
> provide, e.g. asking and replicating a project.
> * Draft up an RFC that discusses the CLI plan and infrastructure which
> will include the points outlined above.

Huh? That's exactly what the two open RFCs #673 and #682 discuss.

FintanH:
> without a framework those commands adhere to.

Sorry, I'm not sure what you mean by "framework" in this context. Could you expand on this?

> The surface area is the entirety of the git suite of commands. The interesting
part here is the "server" end

Just to clarify, the "server" end being the monorepo storage in our case?

> Because there is no way to assert that this command was initiated _interactively_

Ah, I think part of _my_ confusion was there was an attempt to solve this by signing a request payload with the key at some point, but it was dropped.

> Huh? That's exactly what the two open RFCs #673 and #682 discuss.

Sorry, 1. being that #673 get reworked/rewritten to cover the daemon specific functionality, whereas 2. builds on #682 to better outline what the CLI-tiers that we covered (after this discussion is resolved).
Alternatively, what were you expecting?

Apologies if I'm misunderstanding some points you're trying to get across.

kim:
> Sorry, I'm not sure what you mean by "framework" in this context.

If there is just an assortment of binaries which happen to be maintained in this
repo, then how do they fit with _other_ binaries which want to be part of the
radicle suite of commands? For example, you say you want to own "tracking".
Someone else is assembling other sorts of commands, but wants that
functionality, too. Do they need to assemble the argument list and `execve`? How
do they pass "global" options (verbosity being the classic)? What happens after
the tracking state was mutated, is it the responsibility of the "tracking"
command _here_ to notify other interested parties? Or is that up to someone
else as well? (Nb. tracking is an offline operation)

> Just to clarify, the "server" end being the monorepo storage in our
> case?

The server end is what is implemented as a remote helper currently.

kim:
this RFC -- which I'm in agreement withWe seem to be talking past each other — does your comment not say “this is too much stuff, we’re gonna do some commands to control a p2p daemon, and that’s it”?

alexjg:
I've written a little bit about other projects with a monolithic architecture in order to make the ideas outlined in this RFC clearer to me, it might be helpful to others: https://gist.github.com/alexjg/e2b63dc103d0a5a895b2a4588a38e2af

FintanH:
> We seem to be talking past each other

Ya, I think so too :sweat_smile: 

> does your comment not say “this is too much stuff, we’re gonna do some commands to control a p2p daemon, and that’s it”?

That wasn't my intent. My (and xla's) attempt was to build on top of this RFC filling in the gaps based on the architecture proposed. Those gaps leaning more towards implementation details, e.g. what we foresee the daemon peer being able to do (i.e. sprucing up #673), laying out the CLI components radicle-link should own. Maybe this attempt was a mistake due to me misunderstanding something _or_ miscommunicating something? :) Maybe a less async discussion could happen (since we all know async is hard ;)) so that we can get on the same page again.

kim:
@alexjg That's a nice writeup of some of the implicits which led me to propose a different approach.

One thing actually stood out from your points about IPFS: you have separate bullet points for DHT, bitswap, and graph syncing (which I'm not sure I understand what it is exactly). There is a possible separation of networked operations, which I've always imagined for `link`:

1. Keep a daemon running for some period of time
2. Connect to a list of well-known nodes (seeds, coworker machine, standard git server), bulk sync, disconnect

Perhaps something to keep in mind.

@FintanH Ok, well then. I guess we agree that talking about just the daemon (called peer-to-peer node here) things in #673 makes sense.

I intentionally tagged @cloudhead here, because he had expressed interest in driving a larger "official CLI" kind of experience. Perhaps we could dial him in for synchronously discussing some tactical details (beware of cancellation =]).

FintanH (dismissed):

cloudhead (approved):
All concerns are addressed from my side :ok_hand:

geigerzaehler (approved):
:ok_hand:

PR #687, opened 2021-06-01
wham

kim:
I have no idea what I'm doing


kim:
Aha.. so the issue is that blocking tasks, namely those which use the transport,
do not make progress any more when the non-blocking ones get dropped. That is,
we need to keep everything running until the blocking tasks complete.

Please stand by while I just quickly write an exectuor inside the executor..

PR #688, opened 2021-06-01
daemon: don’t panic when discovery stream ends

geigerzaehler:
When the sender for the seed nodes is dropped we stop emitting values from the discovery stream instead of panicking.

We’re [seeing these panics](https://buildkite.com/monadic/radicle-upstream/builds/9127#cc84739f-dd04-40e4-88e7-10367875b581/6-169) regularly in Upstream CI when the node is reloaded.

### DIFF
--- a/daemon/src/config.rs
+++ b/daemon/src/config.rs
@@ -115,7 +115,14 @@ impl discovery::Discovery for StreamDiscovery {
                 for seed in seeds {
                     yield seed.into();
                 }
-                self.seeds_receiver.changed().await.unwrap();
+
+                match self.seeds_receiver.changed().await {
+                    Ok(_) => {},
+                    Err(_) => {
+                        log::warn!("Peer discovery stream dropped");
+                        break;
+                    }
+                }
             }
         };
 


daemon/src/config.rs
```diff
@@ -115,7 +115,12 @@ impl discovery::Discovery for StreamDiscovery {
                 for seed in seeds {
                     yield seed.into();
                 }
-                self.seeds_receiver.changed().await.unwrap();
+
+                match self.seeds_receiver.changed().await {
+                    Ok(_) => {},
+                    Err(_) => { break }
```
alexjg, line 121: Might be worth adding a log line here so we know the discovery is going away.

geigerzaehler: Added the log line in https://github.com/radicle-dev/radicle-link/pull/688/commits/d1558357feb177b76ff06b883d545ea4f41708f7

FintanH (approved):

alexjg (approved):

xla (approved):
🛳 🍢 🛍 🌓

geigerzaehler:
Triple approval :muscle:

PR #689, opened 2021-06-01
daemon: Announce user updates

CodeSandwich:
This fixes an issue with users not seeing each others' updates, e.g. of Ethereum claims

### DIFF
--- a/daemon/src/peer/announcement.rs
+++ b/daemon/src/peer/announcement.rs
@@ -10,7 +10,7 @@ use std::collections::HashSet;
 use librad::{
     git::Urn,
     git_ext::{Oid, RefLike},
-    identities::urn::ParseError,
+    identities::{urn::ParseError, SomeIdentity},
     net::peer::Peer,
     signer::Signer,
 };
@@ -75,31 +75,30 @@ async fn build<S>(peer: &Peer<S>) -> Result<Updates, Error>
 where
     S: Clone + Signer,
 {
-    let mut list: Updates = HashSet::new();
-
-    match state::list_projects(peer).await {
+    let identities = match state::list_identities(peer).await {
         // TODO(xla): We need to avoid the case where there is no owner yet for the peer api, there
         // should be machinery to kick off these routines only if our app state is ready for it.
-        Err(state::Error::Storage(librad::git::storage::Error::Config(_err))) => Ok(list),
-        Err(err) => Err(err.into()),
-        Ok(projects) => {
-            for project in &projects {
-                if let Some(refs) = state::list_owner_project_refs(peer, project.urn()).await? {
-                    for ((one_level, oid), category) in refs.iter_categorised() {
-                        list.insert((
-                            Urn {
-                                path: Some(RefLike::from(category).join(one_level.clone())),
-                                ..project.urn()
-                            },
-                            *oid,
-                        ));
-                    }
-                }
-            }
-
-            Ok(list)
-        },
+        Err(state::Error::Storage(librad::git::storage::Error::Config(_))) => Vec::new(),
+        result => result?,
+    };
+    let mut updates: Updates = HashSet::new();
+    for identity in identities {
+        let urn = match identity {
+            SomeIdentity::Person(person) => person.urn(),
+            SomeIdentity::Project(project) => project.urn(),
+            _ => continue,
+        };
+        let refs = match state::load_refs(peer, urn.clone()).await? {
+            Some(refs) => refs,
+            None => continue,
+        };
+        for ((one_level, oid), category) in refs.iter_categorised() {
+            let path = RefLike::from(category).join(one_level.clone());
+            let urn = urn.clone().with_path(path);
+            updates.insert((urn, *oid));
+        }
     }
+    Ok(updates)
 }
 
 /// Computes the list of announcements based on the difference of the `new` and
--- a/daemon/src/state.rs
+++ b/daemon/src/state.rs
@@ -261,7 +261,7 @@ where
 /// # Errors
 ///
 /// * if opening the storage fails
-pub async fn list_owner_project_refs<S>(peer: &Peer<S>, urn: Urn) -> Result<Option<Refs>, Error>
+pub async fn load_refs<S>(peer: &Peer<S>, urn: Urn) -> Result<Option<Refs>, Error>
 where
     S: Clone + Signer,
 {
@@ -926,6 +926,24 @@ where
             signer: peer.signer().clone(),
         }),
     }
+}
+
+/// Returns the list of [`SomeIdentity`]s for the local peer.
+///
+/// # Errors
+///
+///   * Retrieving the project entities from the store fails.
+pub async fn list_identities<S>(peer: &Peer<S>) -> Result<Vec<SomeIdentity>, Error>
+where
+    S: Clone + Signer,
+{
+    peer.using_storage(move |store| {
+        let identities = identities::any::list(store)?
+            .filter_map(Result::ok)
+            .collect::<Vec<_>>();
+        Ok(identities)
+    })
+    .await?
 }
 
 /// Determine the [`peer::Role`] for a given [`Project`] and [`PeerId`].
--- a/test/src/daemon/common.rs
+++ b/test/src/daemon/common.rs
@@ -7,11 +7,15 @@ use std::{future::Future, net::SocketAddr, path::PathBuf, time::Duration};
 
 use anyhow::Context as _;
 use futures::{future, StreamExt as _};
+use lazy_static::lazy_static;
+use serde::{Deserialize, Serialize};
 use tokio::{sync::broadcast, time::timeout};
+use url::Url;
 
 use librad::{
     git::{identities::local::LocalIdentity, Urn},
     git_ext::OneLevel,
+    identities::payload::HasNamespace,
     keys::SecretKey,
     peer::PeerId,
     reflike,
@@ -262,5 +266,20 @@ pub fn shia_le_pathbuf(path: PathBuf) -> project::Create {
         },
         description: "do".to_string(),
         default_branch: OneLevel::from(reflike!("it")),
+    }
+}
+
+#[derive(Deserialize, Serialize)]
+pub struct TestExt(pub String);
+
+lazy_static! {
+    static ref NAMESPACE: Url = "https://radicle.xyz/test"
+        .parse()
+        .expect("Static URL malformed");
+}
+
+impl HasNamespace for TestExt {
+    fn namespace() -> &'static Url {
+        &NAMESPACE
     }
 }
--- a/test/src/test/integration/daemon/gossip.rs
+++ b/test/src/test/integration/daemon/gossip.rs
@@ -12,9 +12,17 @@ use tokio::time::timeout;
 use radicle_daemon::{peer::run_config, seed::Seed, state, RunConfig};
 
 use crate::{
-    daemon::common::{assert_cloned, radicle_project, requested, shia_le_pathbuf, Harness},
+    daemon::common::{
+        assert_cloned,
+        radicle_project,
+        requested,
+        shia_le_pathbuf,
+        Harness,
+        TestExt,
+    },
     logging,
 };
+use librad::identities::payload::{Person, PersonPayload};
 
 #[test]
 fn can_observe_announcement_from_connected_peer() -> Result<(), anyhow::Error> {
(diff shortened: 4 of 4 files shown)


daemon/src/state.rs
```diff
@@ -151,10 +153,14 @@ where
...
     Error: From<P::Error>,
 {
-    let urn = default_owner(peer).await?.ok_or(Error::MissingOwner)?.urn();
+    let owner = default_owner(peer).await?.ok_or(Error::MissingOwner)?;
+    let urn = owner.urn();
     let payload = payload.try_into()?;
     peer.using_storage(move |store| person::update(store, &urn, None, payload, None))
         .await??;
+
+    gossip::announce(peer, &owner.urn(), None);
```
alexjg, line 162: I think you could add the revision of the document in this gossip announcement using the result of the `peer::using_storage` call (specifically the `content_id` field on the `Person` returned by `person::update`). This allows peers that receive the message to check if they have already fetched this revision before fetching from the peer they receive the gossip from which is useful to avoid situations where a gossip message gets delayed and another peer has replicated in the meantime causing the peer that receives the gossip message to initiate an unneeded fetch.

FintanH (changes requested):
I would think that it might be better to change [this](https://github.com/radicle-dev/radicle-link/blob/master/daemon/src/peer/announcement.rs#L80) to iterate over all namespaces, which can be done using [`any::list`](https://github.com/radicle-dev/radicle-link/blob/master/librad/src/git/identities/any.rs#L49-L51).

xla (changes requested):
To build on @FintanH points. The current solution will not have the desired effect when the local peer is offline. Extending the announcement subroutine is the more general way to solve this.

daemon/src/state.rs
```diff
@@ -938,6 +939,24 @@ where
...
+/// # Errors
+///
+///   * Retrieving the project entities from the store fails.
+pub async fn list_identities<S>(peer: &Peer<S>) -> Result<Vec<SomeIdentity>, Error>
+where
+    S: Clone + Signer,
+{
+    peer.using_storage(move |store| {
+        let projects = identities::any::list(store)?
+            .filter_map(Result::ok)
```
FintanH, line 953: I'm wary of the error cases being hidden here. Something like this will lead to questions like, "How come this Urn was never announced?", due to it being malformed or an underlying bug. I would say we either outright `Error` here or we add some logging. I think I'm leaning more towards the former.

CodeSandwich:
@FintanH I've added [this]( https://github.com/radicle-dev/radicle-link/pull/689/commits/d11669af5bd40da6244a90d8ecc977607bcf622c). Now it processes every URN received from `any::list`. It pushes both projects and persons through `list_owner_project_refs`, but all it seems to be doing is receiving the OIDs for a URN. Is this what you've suggested?

I'm not sure, but it looks like now the node will track all persons and projects it can lay eyes on. Is this necessary? After the addition of announcements of person updates my 2 instances of upstream on 2 machines started seeing their person updates. What's the reason to broaden the listening spectrum?

daemon/src/state.rs
CodeSandwich: It's a clone of `list_projects` and `list_users`, they too skip these errors. Should I fix them too or is it a special case?

CodeSandwich:
@xla
> Extending the announcement subroutine is the more general way to solve this.

I don't follow how should it be extended?

daemon/src/state.rs
FintanH: Ah fair, it's been a while since I looked at this code :innocent: I would say it's fine then as this is likely to be revisited regardless due to the upcoming RFCs =]

daemon/src/state.rs
CodeSandwich: That's a great improvement! Fixed

FintanH:
> It pushes both projects and persons through list_owner_project_refs, but all it seems to be doing is receiving the OIDs for a URN. Is this what you've suggested?

Ya, `list_owner_project_refs` is a bit of a misnomer really. It lists the `signed_refs` of any identity. So as you said it's listing the lastest `Oid` for a `Urn`. So what's happening is that when the annoucnement store notices a new change it will make the announcement and keep track of it in the store.

> I'm not sure, but it looks like now the node will track all persons and projects it can lay eyes on. Is this necessary?

That's a question for the `radicle-seed` implementation. But note that it will track persons anyway if you create/replicate projects because your `Person` information is needed as part of `rad/self` and `rad/ids/*`.

FintanH:
> Ya, list_owner_project_refs is a bit of a misnomer really.

Could we rename this to `load_refs` instead, while we're here? :)

CodeSandwich:
Renamed :)

daemon/src/state.rs
```diff
@@ -508,19 +515,24 @@ where
...
-        Ok::<_, Error>(local::load(store, malkovich.urn())?)
-    })
-    .await??
-    .ok_or(Error::IdentityCreationFailed)
+            Ok::<_, Error>(local::load(store, malkovich.urn())?)
+        })
+        .await??
+        .ok_or(Error::IdentityCreationFailed)?;
+
+    gossip::announce(peer, &identity.urn(), None);
```
FintanH, line 533: I don't think this needed, since the `announcement` loop will pick it up :)

daemon/src/state.rs
```diff
@@ -151,10 +153,15 @@ where
...
-    let urn = default_owner(peer).await?.ok_or(Error::MissingOwner)?.urn();
+    let owner = default_owner(peer).await?.ok_or(Error::MissingOwner)?;
+    let urn = owner.urn();
     let payload = payload.try_into()?;
-    peer.using_storage(move |store| person::update(store, &urn, None, payload, None))
+    let updated_person = peer
+        .using_storage(move |store| person::update(store, &urn, None, payload, None))
         .await??;
+
+    gossip::announce(peer, &owner.urn(), Some(updated_person.content_id));
```
FintanH, line 163: Likewise here =)

daemon/src/state.rs
```diff
@@ -134,6 +134,8 @@ where
 
     set_default_owner(peer, owner.clone()).await?;
 
+    gossip::announce(peer, &person.urn(), None);
```
FintanH, line 137: And here :P

FintanH (commented):
Sorry, one last thing. The point I was making was that the `announce` calls don't need to be made because the announcement loop will handle that.

I see that [`init_project` has one too](https://github.com/radicle-dev/radicle-link/pull/689/files#diff-4ca5d857b3b745b05c3c179a96e45e3fc74f072d8bd4c9b2b1ad178f076d491fR499), so maybe you were copying that? I'm not actually sure why that needs to be there. Could be an artifact of "UX". Do you know @xla?

daemon/src/state.rs
CodeSandwich: Addition of these 3 `announce`s is the reason why this PR exists in the first place :laughing: Without them the nodes wouldn't exchange information about person updates.

FintanH: Wuuuuut?! Even with the suggestion of changing the announcements code?

xla:
> I see that [`init_project` has one too](https://github.com/radicle-dev/radicle-link/pull/689/files#diff-4ca5d857b3b745b05c3c179a96e45e3fc74f072d8bd4c9b2b1ad178f076d491fR499), so maybe you were copying that? I'm not actually sure why that needs to be there. Could be an artifact of "UX". Do you know @xla?

That's an artifact from when the announcement wasn't running as frequently, smoke and mirrors. The explicit announce call can go.

daemon/src/state.rs
CodeSandwich: This PR doesn't change the announcements code (probably yet). It adds 3 `announce`s, which are fixing the root issue and expands `build`, which IIUC should make the node listen for more announcements than before. I'm not even sure if we need to modify `build` in the first place, the issue is clearly about sending, not receiving announcements, the nodes were communicating fine with unmodified `build`.

FintanH: But it does change the `announcements`: https://github.com/radicle-dev/radicle-link/pull/689/commits/d11669af5bd40da6244a90d8ecc977607bcf622c

> node listen for more announcements than before

Other way around, it expands the announcements _made_ from just projects to projects & persons. This was suggested so we could remove adding `announce` to each `init`/`update` action, separating them from the annoucnement logic.

CodeSandwich: Oh, I didn't notice that about `build` :facepalm:

I've created an experimental branch with `build` changes from this PR, but no explicit `announce` additions. Unfortunately it doesn't work well, upstream based on it apparently doesn't announce its person updates, they aren't visible on the other instance.

FintanH: > Oh, I didn't notice that about build 

Ahaha no worries :)

> Unfortunately it doesn't work well, upstream based on it apparently doesn't announce its person updates, they aren't visible on the other instance.

Huh, could you possibly run with logs enable, i.e. `RUST_LOG=librad::net=trace`, and see if you can spot announcements for `Person` or if they don't show up at all. Maybe sprinkle some logs for the announcements as well.

CodeSandwich: Hmm, I can't spot announcements in logs in any version, with or without manual `announce`s. The worst part is that person changes don't seem to be propagated anymore, the other upstream node doesn't notice the updates :grimacing:

FintanH: Is there a specific Upstream branch you're working on that I can run and inspect?

CodeSandwich: Yes, thank you!

I've backported changes from this PR to a commit used by upstream:
- With changes to `build`: [igor/announce-2](https://github.com/radicle-dev/radicle-link/tree/igor/announce-2)
- With changes to `build` and additional `announce`s: [igor/announce-3](https://github.com/radicle-dev/radicle-link/tree/igor/announce-3)

These backports are used in upstream:
- With changes to `build`: [igor/org-list-no-announce](https://github.com/radicle-dev/radicle-upstream/tree/igor/org-list-no-announce)
- With changes to `build` and additional `announce`s [igor/org-list-with-announce](https://github.com/radicle-dev/radicle-upstream/tree/igor/org-list-with-announce)

I've been trying to provoke person update announcements by linking the wallet to different addresses. The update to the document is made BEFORE the Ethereum transaction is sent. If you need any help, please reach out, it's not that trivial to get all the pieces together.

CodeSandwich: Any news, @FintanH?

FintanH: Before I go down a rabbit hole, could you make an additional test similar to https://github.com/radicle-dev/radicle-link/blob/master/test/src/test/integration/daemon/gossip.rs#L19-L65 so that we test closer to the source?

CodeSandwich: I've done some rebasing, there were some changes to the base upstream branch. Now it uses a newer version of link, which includes `can_observe_announcement_from_connected_peer`. It passes.

The link branches used in upstream:
- With changes to `build`: [announce-backport-no-explicit](https://github.com/radicle-dev/radicle-link/tree/igor/announce-backport-no-explicit)
- With changes to `build` and additional `announce`s: [igor/announce-backport](https://github.com/radicle-dev/radicle-link/tree/igor/announce-backport)

They are used in upstream:
- With changes to `build`: [igor/org-list-no-announce](https://github.com/radicle-dev/radicle-upstream/tree/igor/org-list-no-announce)
- With changes to `build` and additional `announce`s [igor/org-list-announce](https://github.com/radicle-dev/radicle-upstream/tree/igor/org-list-announce)

FintanH: > I've done some rebasing, there were some changes to the base upstream branch. Now it uses a newer version of link, which includes can_observe_announcement_from_connected_peer. It passes.

But can we add a test (or add to that test) a case where we have a `Person` update and witness the announcement? I believe that test only checks for a `Project`'s announcement.

CodeSandwich: I've added the test. It fails before `igor/announce-backport-no-explicit` and passes since it. Explicit `announce`s don't make a difference.

FintanH: Ok, so that's a good sign that we at least know the announcement loop is working correctly. So we need to understand why this isn't happening in the Upstream environment, one oddity being that the explicit `announce` calls worked. 

On the same branch, were you able to see projects being announced and replicated?

xla: Wanted to highlight radicle-dev/radicle-upstream#1956 as it indicates broader issues with gossip in the context of Upstream.

CodeSandwich:
Can we merge this PR, @FintanH? It's obviously fixing an issue, the test shows that, but it'll be much easier to work and look for the bugs after the merge. The constant rebasing and backporting is introducing massive complexity and uncertainty, which obscures the real problems and eats away hours of work.

FintanH:
> It's obviously fixing some issue, the test shows that, but it'll be much easier to work and look for the bugs after the merge.

Sorry, I don't understand your reasoning here. The test works without the explicit `announce` calls, so that means the bug you're witnessing is down to something else and the announcement loop is emitting the gossip correctly.

Understandable that it may be frustrating to rebase, and I'm happy for us to get this into a mergeable state. But I'd also like to understand what's your plan to find the bug then?

CodeSandwich:
> The test works without the explicit announce calls

You're right, the `announce`s aren't needed, I've dropped them. But changes to `build` are fixing the test and IMO they are worth merging for that reason alone. That's probably our bug fixed or at least it's a step in the right direction. It's hard to tell if it fixes the issue completely, there may be additional steps, but it'll be easier to find out and keep track after the merge.

FintanH:
Ya, the improvement of `build` should be merged! =] Sorry, maybe I've been communicating poorly here -- the explicit `announce` calls were what I wanted to prevent being introduced and I assumed that they were the "fix" to the upstream issue. I'll take one last pass and should be good to go

FintanH (approved):
Thanks for the improvement and test :heart:

CodeSandwich:
@xla How about you? This PR is blocked unless you approve it.

daemon/src/peer/announcement.rs
```diff
@@ -75,31 +75,30 @@ async fn build<S>(peer: &Peer<S>) -> Result<Updates, Error>
...
-                            },
-                            *oid,
-                        ));
-                    }
-                }
-            }
-
-            Ok(list)
-        },
+        Err(state::Error::Storage(_)) => Vec::new(),
```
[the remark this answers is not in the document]
CodeSandwich: It's a less elaborate version of this removed line:
```rust
// TODO(xla): We need to avoid the case where there is no owner yet for the peer api, there
// should be machinery to kick off these routines only if our app state is ready for it.
 Err(state::Error::Storage(librad::git::storage::Error::Config(_err))) => Ok(list),
```
I have no idea what does it mean, so I've kept both the logic and the comment.

daemon/src/state.rs
```diff
@@ -928,6 +928,24 @@ where
...
+///
+/// # Errors
+///
+///   * Retrieving the project entities from the store fails.
+pub async fn list_identities<S>(peer: &Peer<S>) -> Result<Vec<SomeIdentity>, Error>
+where
+    S: Clone + Signer,
+{
+    peer.using_storage(move |store| {
+        let projects = identities::any::list(store)?
```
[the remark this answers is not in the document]
CodeSandwich: It should. Fixed.

daemon/src/peer/announcement.rs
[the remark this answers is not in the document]
xla: These are not equivalent, the original ignores exactly one possible storage error while the rest is surfaced. With the new version all storage errors are masked.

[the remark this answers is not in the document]
CodeSandwich: Good catch, I haven't noticed that it's an enum variant! Fixed.

xla:
CI is not running all the way through to the tests because of an issue in a dependency. Can you reproduce locally?

CodeSandwich:
@xla It's compiling fine locally. All the CI runs are failing for the same reason and it's the same in all currently open PRs.

xla:
@CodeSandwich Could you merge/rebase master to have the tests run properlier on CI.

FintanH (approved):

xla (approved):
💂 🚝 🌴 🔗

PR #690, opened 2021-06-03
librad: remove `err` logging from instrument macro

kim:
The `err` parameter for the `tracing::instrument` macro logs at error
level unconditionally when the `Result`-returning function returns, err,
`Err`. This was once useful for debugging, but nowadays adds noise: it
is often hard to tell where the log line comes from, it's not always an
error (ie. it might be handled), or multiple log lines are emitted for
the same error condition.

Thus, remove it everywhere.


FintanH (approved):

xla (approved):
📌 🐋 ⬆️ 🏝

PR #692, opened 2021-06-03
io: graceful shutdown

kim:
Latest installment of the cancellation saga: the shutdown sequence for
an endpoint needs to be carefully grafted in order to not stall blocking
tasks (which call back into the async context), and it is advisable to
call `wait_idle` on the endpoint. In turn, our aggressive cancellation
policy needs to be loosened a bit to not immediately abort everything,
but drain outstanding tasks. Iow, if tests hand on shutdown, it is now a
programmer error.

Arguably, the return type of `accept` is now pretty weird (one future
for shutdown, and another one to actually drive the accept loop). Not
sure if there is a better formulation, as we are also constrained by
opaque types.

Note that the `menage` tests are still failing, which appears to depend
on whether the force-flag-honouring libgit2 is picked up or not.

### DIFF
--- a/daemon/src/peer.rs
+++ b/daemon/src/peer.rs
@@ -196,10 +196,12 @@ where
                         addrs_tx
                             .send(bound.listen_addrs())
                             .expect("subroutines is gone");
-                        if let Err(e) = net::protocol::accept(bound, disco.clone().discover()).await
-                        {
+                        let (shutdown, run) =
+                            net::protocol::accept(bound, disco.clone().discover());
+                        if let Err(e) = run.await {
                             log::error!("accept error: {}", e);
                         }
+                        shutdown.await
                     },
                     Err(e) => {
                         log::error!("bound error: {}", e);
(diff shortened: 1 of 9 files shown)


xla:
> Arguably, the return type of `accept` is now pretty weird (one future
> for shutdown, and another one to actually drive the accept loop). Not
> sure if there is a better formulation, as we are also constrained by
> opaque types.

From the docs I get that the `shutdown` shouldn't be polled before `accept` is complete, with the api be hardened by returning the shutdown future in accept result to ensure this invariant?

kim:
hardenedNo because the caller wants to cancel the accept loop before it returns (which might be never), **and** the cancellation is a future itself (and there is no async drop). Idk, we could use a tagged union instead of a tulle, but due to opaque types this would really just be about giving the two futures a name.

xla (approved):
🐩 🔅 ☝️ ♈️

alexjg (approved):

PR #693, opened 2021-06-03
Rate limits

kim:
Apply some rate limiting to membership + gossip ingress traffic. Quota defaults are fictional.


xla (dismissed):
utACK

🏢 📠 ⛴ 🦂

alexjg (dismissed):

FintanH (commented):
I'm a bit confused running through this code. I can see where we check the limiter to ignore bad actors, but I can't see where the limits are increased and where a bad actor is eventually removed from the limiter (perhaps because they accidentally sent too many requests). Is this happening in the background on the `governor`?

kim:
https://en.wikipedia.org/wiki/Generic_cell_rate_algorithm

kim:
The semantics is that you'll be disconnected immediately if you send to many membership messages (which is fair, because you're not well-behaving). For gossip, it drops messages, which is not ideal because we're not looking at the payload. It would probably be better to cache `origin || payload`, which is inevitable a `Stale` result when received more than once. But then I'm not sure how to make this cache bounded...

FintanH:
> The semantics is that you'll be disconnected immediately if you send to many membership messages (which is fair, because you're not well-behaving)

And then you can re-join if there's the capacity for you again (ala the GCRA)?

> which is not ideal because we're not looking at the payload

Are we guarding against stale messages being sent over and over or just too much gossip from one peer in general? If it's the former then could we do a `git_has` pre-emptively as well as check the limit?

kim:
> And then you can re-join

Depends what the other end does when it gets disconnected. It could also immediately re-connect and flood and get limited and get disconnected.

> Are we guarding against stale messages being sent over and over or just too much gossip from one peer in general?

I don't know, because I haven't looked at what those folks are sending. 

> do a git_has pre-emptively

We do that. It's expensive.

FintanH (approved):

FintanH:
> I don't know, because I haven't looked at what those folks are sending.

Sorry, I meant if the _goal_ was to stop general gossip flooding or just stale messages.

Thanks for the explanations and links :)

kim:
> I meant if the goal was to stop general gossip flooding or just stale messages

Well, it doesn't matter if you think about it: if you receive the same message again, it's stale (ie. you don't forward it). If you receive a lot of _different_ (but valid) messages, then you need to somehow compensate for the load. Unless we add PoW to a message, the best we can do is to avoid I/O by discarding already-seen messages earlier.

librad/src/net/peer/storage.rs
```diff
@@ -215,6 +288,15 @@ impl broadcast::LocalStorage<SocketAddr> for Storage {
...
                 Err(e) => match e {
                     Error::KnownObject(_) => PutResult::Stale,
+                    Error::RateLimited { remote_peer, urn } => {
+                        tracing::warn!(
+                            "skipped fetch of {} from {} due to rate limiting",
+                            remote_peer,
+                            urn
+                        );
+                        // Someone else may still be interested
+                        PutResult::Uninteresting
```
xla, line 298: Isn't a `PutResult::Stale` more meaningful for this branch? Alternatively bubble up the RateLimited as variant of `PutResult`?

FintanH: I think `Stale` means the gossip won't be forwarded, whereas `Uninteresting` will be.

So, correct me if I'm wrong, the semantics of this is that this peer is pummeling me so I'm going to rate-limit them, but some other peer may be able to reply to their gossip (they may not be being pummelled =])

kim: Yes that’s right, although if all the pummels hit the limit at the same time, the message will bounce indefinitely. Stale is not correct, either, unless one argues that chatty peers should be punished by not forwarding their logorrhoea. Perhaps a delay would work, but that consumes resources on the wrong end.

librad/src/git/storage/watch.rs
```diff
@@ -0,0 +1,98 @@
...
+    pub path: PathBuf,
+    pub kind: EventKind,
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+#[non_exhaustive]
+pub enum EventKind {
+    Create,
+    Remove,
+    Update,
```
FintanH, line 48: Is it intentional that `Update` is never used?

kim:
Pending some history hygiene, this is now ready for another round.

Summary of changes:

- The xor filter holding the URNs present locally is shared, such that both interrogation and peer storage have access to it
- The max size of the filter is bumped to 100k elements or ~315KiB

   It is rather unlikely that we can support 100k namespaces in a single monorepo, so this should be safe for a while =]

- Peer storage will only hit the disk if a URN is in the filter (modulo the fp rate of the filter)
- The shared filter is rebuilt when namespaces change on disk. 
 
   This is implemented by watching reflogs, which we create for the refs managed by librad (and only those). There is obviously a delay between namespace creation and rebuilding the cache, during which the peer will claim to not know about it. That may cause a decrease in QoS in the shotgun model, but is acceptable once #506 lands.

- Rate limits are applied as follows:
   * Retries when the local storage returns errors are limited aggressively
   * Too many consecutive `Want`s are ignored (again, #506)
   * Fetch attempts in response to gossip are throttled to 1/min (fetches are expensive, cannot proceed concurrently, and under fair use updates are infrequent)
   * Too many membership messages cause the remote peer to be disconnected (the default value is potentially too lax)

FintanH (approved):

xla (approved):
🚖 🍩 👧 ⚡️

librad/src/git/storage/pool.rs
```diff
@@ -63,41 +64,65 @@ impl From<Object<Storage, Error>> for PooledRef {
...
+    pub fn no() -> Self {
+        Self(Arc::new(RwLock::new(false)))
+    }
+}
+
 #[derive(Clone)]
 pub struct Config<S> {
     paths: Paths,
     signer: S,
+    init: Initialised,
```
FintanH, line 80: Can you elaborate a bit on what's going on with `Initialised`, please? From looking at the code, it only seems to track the `bool` value, but we call `mk_storage` regardless, so I'm a bit confused as to what magic it might be doing.

kim: It's sharing a mutex between different pool instances, to make sure only one storage instance initialises the repo.

FintanH: Ah, I see it now. So initially it will be false and one thread will have the exclusive lock initialising the repo via `Storage::with_fetchers`. After that `Storage::with_fetchers` will just open the repo since it exists and all threads will go through the first branch of the `if`.

kim:
FYI: the tests suddenly hanging again seems to have to do with some unfortunate constellation of cosmic radiation or something. I can reproduce on `master`, so it is not introduced by this series.

xla (approved):
Pushed to prod, no peers were harmed in the process.

💷 📠 🏂 📃

kim:
> Pushed to prod, no peers were harmed in the process.

I doubt that. There is a failing test `test::integration::daemon::replication::track_peer` which apparently relies on issuing a `Want` having the side-effect of cloning a project. That obviously doesn't work, because the tracking peer will not recognise the `Have`.

kim:
Or well, no, scrap that. But the test fails nevertheless.

kim:
The test passes when waiting for a replicate event _twice_. I have no clue how that can be, perhaps something introduced in #689?

The remaining test failures are very fun: apparently `libgit2` doesn't clear its ref cache fast enough.

PR #694, opened 2021-06-07
First pass at collaborative objects

alexjg:
This is a first pass at an implementation of the colaborative objects RFC. It does not yet include schema validation. It provides a CRUD API for collaborative objects which are synced by the existing replication mechanisms. Here are the main things this PR does:

- Introduce a `librad::collaborative_objects::CollaborativeObjects` struct which is a CRUD interface to collaborative objects stored in the radicle monorepo
- hoist some code out of `identities::sign` into a top level module so we can reuse the commit signing logic for collaborative objects

There are currently no tests, I'm working on that now. However, I've been test driving this with a simple CLI for issues which you can find [here](https://github.com/alexjg/collaborative-issues)

Here are a few points which have occurred to me as I wrote this

- The `CollaborativeObjects` struct - which is the entry point for this functionality - has a direct reference to a `git2::Repository` and many parts of the collaborative objects implementation talk directly to this repository. This may not be desirable.
- Retrieving an object involves traversing the graph of changes to that object. This is possibly problematic for a single large object but it's definitely a problem when listing entire subsets of collaborative objects (e.g all issues). We will need to introduce a caching layer somewhere and then we will need to make decisions about when that cache is created and updated.
- Currently all objects are replicated as part of the standard replication mechanisms. We have outlined in the RFC a way in which we might avoid doing this by passing parameters to the URN we're replicating to specify that whether we do or don't want to fetch collaborative objects. This means we would need to expose an API for this to applications and decide when we do and don't want to replicate collaborative objects in the application.


kim:
> Merged #694 into master.

Why, and how, could this happen?

PR #696, opened 2021-06-08
RFC: Peer-to-peer node

xla:
[Rendered](https://github.com/radicle-dev/radicle-link/blob/xla/rfc-p2p-node/docs/rfc/0696-p2p-node.adoc)

### DIFF
--- a/docs/rfc/0696-p2p-node.adoc
+++ b/docs/rfc/0696-p2p-node.adoc
@@ -1,0 +1,300 @@
+= RFC: Peer-to-peer Node
+:author: @xla
+:revdate: 2021-09-02
+:revremark: accepted
+:toc:
+:toc-placement: preamble
+
+* Author: {author}
+* Date: {revdate}
+* Status: {revremark}
+* Community discussion: n/a
+* Tracking Issue: https://github.com/radicle-dev/radicle-link/issues/722
+
+== Overview
+
+This RFC expands on the peer-to-peer node outlined in <<rfc-0682, RFC 0682>> to
+propose a new stand-alone daemon with focus on driving the p2p stack, exposing a
+minimal API and defer supervision to more integrated and mature system services.
+Furthermore it will prove that all daemons are made equal and only distinct with
+regards to the configuration given (e.g. tracking behaviour) and lifetime of the
+process (i.e. socket-activated or long-running).
+
+It also strives to recommend APIs and mechanisms -- in the form of non-core
+protocols -- which remove the need for online requests made by clients to a
+running node.
+
+== Terminology and Conventions
+
+The key words "`MUST`", "`MUST NOT`", "`REQUIRED`", "`SHALL`", "`SHALL NOT`",
+"`SHOULD`", "`SHOULD NOT`", "`RECOMMENDED`", "`NOT RECOMMENDED`", "`MAY`", and
+"`OPTIONAL`" in this document are to be interpreted as described in <<RFC2119>>
+and <<RFC8174>> when, and only when, they appear in all capitals, as shown here.
+
+CBOR <<RFC8949>> datatype definitions are given using the notation devised in
+CDDL <<RFC8610>>. By convention, `struct`-like datatypes are encoded as CBOR
+maps, where the map key is the zero-based numeric index of the field in
+declaration order.
+
+== Protocols
+
+The main purpose of the node is to operate the core protocol of link currently
+implemented in `librad::net::peer`, which currently consists of holding onto the
+futures du jour and dropping them at opportune moments. Furthermore all
+information in form of synchrounous calls and events obtainable internally via
+the sufficiently space-engineered tincans SHALL be made available.
+
+Besides the core protocol other functionality will be encapsulated in protocols
+which will be driven by a stream of well-defined inputs. The well-defined
+outputs of these protocols are used for effect management.
+
+=== Subroutines
+
+This section expands on the auxiliary routines driven by the node on top of the
+core protocol. Where each SHOULD be allowed to be disabled through configuration.
+A subroutine acts as a reactor driving a protocol by feeding inputs - usually
+results of effects (i/o, timers, etc.) - to it and capturing the outputs to
+schedule new effects (i/o, timers, etc.).
+
+==== Public API
+
+For other processes to be able to talk to the node, it SHALL keep an API
+subroutine running, which offers introspection into its state. See the dedicated
+section for the surface the node will expose.
+
+==== Announcements
+
+Until superseded by <<pr-653, graftier ways>>, currently the main means of
+dissemination is through gossiping new interesting `Have`s. This is achieved by
+maintainig a persisted view on the list of identities and their current refs.
+During periodic runs a delta is created which serves as the basis of the
+announcements. In turn that list is made known to actively connected peers via
+the core protocols gossip.
+
+The delta mechanism works by recording a snapshot of the state of the monorepo,
+specifically the refs, and on the next iteration diffing against the current
+state. Relevant refs from that diff can be used to announce or perform other
+periodic tasks.
+
+==== Replication Requests
+
+The core protocol implicitly tracks projects after replication, and explicitly
+with the `track` operation for remotes in the context of a `Urn`. To enable the
+same approach as for announcements where offline changes to the storage can be
+picked up at any point the node is running and online, this subroutine will
+maintain a persisted view and periodically (potentially provoked by a request
+over the <<API>>) build the delta for new tracked `Urn`s and remotes in their
+context.
+
+Note: A blocking issue for that is
+https://github.com/radicle-dev/radicle-link/issues/141[#141]
+
+==== Tracking
+
+Given the ability to configure a set of urn, peers, both or anything observed
+from connected peers, the node will automatically track and replicate. This
+strategy is currently in use in the seed node and SHALL be preserved.
+
+== API
+
+=== IPC
+
+All communication with daemon processes SHALL occur over UNIX domain sockets in
+`SOCK_STREAM` mode. The socket files MUST be stored in a directory only
+accessible by the logged-in user (typically `$XDG_RUNTIME_DIR`), and have
+permissions `0700` by default. Per-service socket paths are considered "well
+known".
+
+RPC calls over those sockets use <<cbor, CBOR>> for their payload encoding. As
+incremental decoders are not available on all platforms, CBOR-encoded messages
+shall be prepended by their length in bytes, encoded as a 32-bit unsigned
+integer in network byte order.
+
+RPC messages are wrapped in either a `request` or `response` envelope structure
+as defined below:
+
+_Verbatim copy from <<rfc-0682, RFC 0682>> to contextualise the payloads below_
+
+[source,cddl]
+----
+request = [
+    request-headers,
+    ? payload: bstr,
+]
+response = ok / error
+ok = [
+    response-headers,
+    ? payload: bstr,
+]
+error = [
+    response-headers,
+    code: uint,
+    ? message: tstr,
+]
+request-headers = {
+    ua: client-id,
+    rq: request-id,
+    ? token: token,
+}
+response-headers = {
+    rq: request-id,
+}
+; Unambiguous, human-readable string identifying the client application. Mainly
+; for diagnostic purposes. Example: "radicle-link-cli/v1.2+deaf"
+client-id: tstr .size (4..16)
+; Request identifier, choosen by the client. The responder includes the
+; client-supplied value in the response, enabling request pipelining.
+;
+; Note that streaming / multi-valued responses may include the same id in
+; several response messages.
+request-id: bstr .size (4..16)
+
+; Placeholder for future one-time-token support.
+token: bstr
+----
+
+
+=== Request/Response payloads
+
+There is a supported set of requests a daemon will answer to which are as follows:
+
+[source,cddl]
+----
+request = get-connected-peers / get-membership-info / get-stats
+get-connected-peers = [0]
+get-membership-info = [1]
+get-stats = [2]
+
+response = connected-peers / membership-info / stats
+connected-peers = [* peer-id]
+membership-info = {
+  active: [* peer-id],
+  passive: [* peer-id],
+}
+stats = {
+  "connections-total": uint,
+  "membership-active": uint,
+  "membership-passive": uint,
+  * tstr => any
+}
+
+connect-info = [
+  peer_id,
+  [address],
+]
+
+address = address-ipv4 / address-ipv6
+address-ipv4 = [
+  ip4,
+  port,
+]
+address-ipv6 = [
+  ip6,
+  port,
+]
+
+ip4 = bstr .size 4
+ip6 = bstr .size 16
+
+; Canonical representation of a peer.
+peer-id: bstr
+
+; Network port.
+port: uint
+----
+
+All types representing requests and responses and their serialisation logic MUST
+be exposed as linkable libraries. It is RECOMMENDED to also expose the
+functionality to communicate with the node via IPC as a library.
+
+== Operations
+
+=== Supervision
+
+Process supervision SHOULD be deferred to established system level service
+managers i.e. `<<systemd>>` and `<<launchd>>` for Linux and macOS respectively.
+To support both long-running as well as ad-hoc usage the daemon implementation
+SHALL be equipped with the ability to detect and read the information from its
+environment necessary to determine if it has been activated via socket. When
+binding to a socket it SHALL use the file descriptors provided by the init
+process. If none are provided it SHALL assume long-running operation and SHALL
+bind to the well-known path on a UNIX domain socket in mode `SOCK_STREAM` under
+Linux:
+
+  $XDG_RUNTIME_DIR/radicle/<srv>-<peer-id>.sock
+
+macOS:
+
+  $TMPDIR/radicle/<srv>-<peer-id>.sock
+
+Both service managers offer support to fullfil the legacy `inetd` interface.
+Which is deemed insufficient for concerns over security, lack of support for
+UNIX domain sockets and the design focusing on a process per connection.
+
+==== systemd
+
+Socket activation under systemd is passed on via: 
+
+* `LISTEN_PID` - MUST be equal to the PID of the daemon.
+* `LISTEN_FDS` - Number of received file descriptors, starting at 3.
+* `LISTEN_NAMES` - Contains colon-separated list of names corresponding to the
+  `FileDescriptorName` option in the service file.
+
+==== launchd
+
+* `LAUNCH_DAEMON_SOCKET_NAME` - Name of the socket according to the `.plist`
+  configuration file.
+
+The name passed to the process MUST be used to check-in with launchd as
+documented in `launch(3)` which in essence involves obtaining the FDs via
+`launch_activate_socket` expecting a name.
+
+=== Configuration
+
+Common service configuration files SHALL be provided alongside the source code
+of the node binary. To support the semi-dynamic nature of one process per
+profile, facilities to manage services with both systemd and launchd SHALL be
+provided through the CLI and automated together with the profile lifecycle
+management.
+
+The binary SHALL expose all knobs necessary to fine-tune the internal configs of
+the core protocol, i.e. `membership`, `protocol`, `storage`. Additionally,
+any switches and configuration that subroutines require. The configuration surface
+SHALL be exposed as command line arguments, until further evidence is brought
+forward which makes a strong case for external config files.
+
+== Key Access
+
+Access to key material SHALL be done through the facilities provided by
+`<<radicle-keystore>>`. Except for debug/development purpose this SHOULD be
+limited to the use of the `ssh-agent`.
+
+The author assumes that the `rad` CLI provides functionality to manage keys on a
+per profile basis including adding them to a running ssh-agent.
+
+== Future Work
+
+Originally this document included a section outlining PubSub solutions. As it
+affects too many other parts of the overall architecture, specifying it will be
+deferred to a follow-up RFC.
+
+Developers! Developers! Developers! - or how nobody knows what to do with
+Windows. While solutions like WSL are present, it's unclear at this point how/if
+a native solution could look like.
+
+
+[bibliography]
+== References
+
+* [[[cbor]]] https://datatracker.ietf.org/doc/html/rfc8949
+* [[[cddl]]] https://datatracker.ietf.org/doc/html/rfc8610
+* [[[launchd]]] https://en.wikipedia.org/wiki/Launchd
+* [[[radicle-keystore]]] https://github.com/radicle-dev/radicle-keystore/
+* [[[systemd]]] https://systemd.io/
+* [[[pr-653]]] https://github.com/radicle-dev/radicle-link/pull/653
+* [[[rk-17]]] https://github.com/radicle-dev/radicle-keystore/pull/17
+* [[[rfc-0682]]] https://github.com/radicle-dev/radicle-link/blob/master/docs/rfc/0682-application-architecture.adoc
+* [[[RFC2219]]] https://datatracker.ietf.org/doc/html/rfc2119
+* [[[RFC8174]]] https://datatracker.ietf.org/doc/html/rfc8174
+* [[[RFC8610]]] https://datatracker.ietf.org/doc/html/rfc8610
+* [[[RFC8949]]] https://datatracker.ietf.org/doc/html/rfc8949


FintanH:
Shall I close https://github.com/radicle-dev/radicle-link/pull/673 in favour of this?

xla:
> Shall I close #673 in favour of this?

Sounds good!

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,318 @@
...
+This RFC expands on the peer-to-peer node outlined in RFC 682 to propose a new
+stand-alone daemon with focus on driving the p2p stack, exposing a minimal API
+and defer supervision to more integrated and mature system services. Furthermore
+it tries to prove that all daemons are made equal and only distinct with regards
+to the configuration given (e.g. tracking behaviour) and lifetime of the process
+(i.e. socket-activated or long-running).
+
+It also strives to recommend APIs and mechanisms - in the form of non-core
+protocols - which remove the need for online requests made by clients to a
+running node.
```
kim, line 39: I think this would perhaps do as `Motivation` :slightly_smiling_face:

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,318 @@
...
+"`SHOULD`", "`SHOULD NOT`", "`RECOMMENDED`", "`NOT RECOMMENDED`", "`MAY`", and
+"`OPTIONAL`" in this document are to be interpreted as described in <<RFC2119>>
+and <<RFC8174>> when, and only when, they appear in all capitals, as shown here.
+
+CBOR <<RFC8949>> datatype definitions are given using the notation devised in
+CDDL <<RFC8610>>. By convention, `struct`-like datatypes are encoded as CBOR
+maps, where the map key is the zero-based numeric index of the field in
+declaration order.
+
+== Protocols
```
kim, line 53: It might be helpful for future reference to introduce the terminology `Protocol` and `Subroutine` properly here. A `Subroutine` is kind of an event loop which drives a `Protocol`, or?

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,318 @@
...
+
+In order to guarantee initial connectivity the protocol needs to be given a set
+of bootstrap nodes, while the main means of discovery is part of the membership
+algorithms. Beyond the initial list passed at start-up as source for discovery
+a streaming variant should be - on opt-in basis - made available. Consult the
+<<API>> section for the internals of how that can be achieved by another process.
+
+==== Announcements
+
+Until superseded by graftier ways, currently the main means of dissemination is
```
kim, line 82: Hyperlink #653 ?

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,318 @@
...
+
+RPC calls over those sockets use <<cbor, CBOR>> for their payload encoding. As
+incremental decoders are not available on all platforms, CBOR-encoded messages
+shall be prepended by their length in bytes, encoded as a 32-bit unsigned
+integer in network byte order.
+
+RPC messages are wrapped in either a `request` or `response` envelope structure
+as defined below:
+
+[source,cddl]
```
kim, line 126: This is unchanged from #682, right? Maybe make a note here.

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,318 @@
...
+request-discover = {
+  type: "discover",
+  info: connect-info,
+}
+request-memebership = {
+  type: "membership",
+}
+request-stats = {
+  type: "stats",
+}
```
kim, line 181:
```suggestion
request = discover / membership-info / stats
discover = [0, connect-info]
membership-info = [1]
stats = [2]
```

This is more compact, while retaining extensibility. Note that you can map this to an `enum` in Rust.

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,318 @@
...
+response-memebership = {
+  active: [peer-id],
+  passive: [peer-id],
+}
+response-stats = {
+  connections-total: uint,
+  connected-peers: [connect-info],
+  memebership-active: uint,
+  memebership-passive: uint,
+}
```
kim, line 193:
```suggestion
response = membership-info / stats
membership-info = {
  active: [* peer-id],
  passive: [* peer-id],
}
stats = {
  "connections-total" ^ => uint,
  "connected-peers" ^ => [* connect-info],
  "membership-active" ^ => uint,
  "membership-passive" ^ => uint,
  * tstr => any
}
```

I'm specifically suggesting to use string keys for the `stats` response, as it's not too important to make this particularly compact, but important to be able to change things later by adding or removing fields. The "cut" (`^ =>`) in combination with `* tstr => any` says that we're free to add more fields to the map (of any type), but the specified keys must be of the given type. Eg. if there was a value '"connections-total": "NaN"', `* tstr => any` is not allowed to match.

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,318 @@
...
+Both service managers offer support to fullfil the legacy `inetd` interface.
+Which is deemed insufficient for concenrs over security, lack of support for
+UNIX domain sockets and the design focusing on a process per connection.
+
+==== systemd
+
+Socket activation under systemd is passed on via: 
+
+* `LISTEN_PID` - MUST be equal to the PID of the daemon.
+* `LISTEN_FDS` - Number of received file descriptors, starting at 3 .
```
FintanH, line 257:
```suggestion
* `LISTEN_FDS` - Number of received file descriptors, starting at 3.
```

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,318 @@
...
+
+
+=== Request/Response payloads
+
+There is a supported set of requests a daemon will answer to which are as follows:
+
+[source,cddl]
+----
+request = discover / memebership / stats
+request-discover = {
```
kim, line 172: There's no response defined for `discover` -- what is it for?

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,318 @@
...
+binding to a socket it SHALL use the file descriptors provided by the init
+process. If none are provided it SHALL assume long-running operation and SHALL
+bind to the well-known path on a UNIX domain socket in mode `SOCK_STREAM` under
+Linux:
+
+  $XDG_RUNTIME_DIR/radicle/<srv>-<peer-id>.sock
+
+macOS:
+
+  TODO(xla): Define well-known path for darwin systems.
```
kim, line 242: I think it's actually `/tmp` (or rather, `/private/tmp`)

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,318 @@
...
+profile, facalities are required to manage service definitions along the
+lifetime of a profile and when updated need to inform the process management
+system. Such facalities need to be considered wherever profiles are modified,
+i.e. CLI.
+
+The binary SHALL expose all knobs necessary to fine-tune the internal configs of
+the core protocol, i.e. `membership`, `protocol`, `storage`. Additionally
+switches and configuration that subroutines require. As this surface is sizable
+even without subroutines parameters, there SHALL be persistent file based
+configuration in a sane format.
```
FintanH, line 283: I presume the "sane format" will be TOML based on yesterday's conversation?

FintanH (commented):
Coming together! =]

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,318 @@
...
+bind to the well-known path on a UNIX domain socket in mode `SOCK_STREAM` under
+Linux:
+
+  $XDG_RUNTIME_DIR/radicle/<srv>-<peer-id>.sock
+
+macOS:
+
+  TODO(xla): Define well-known path for darwin systems.
+
+If sufficiently focused library implementations are present in the ecosystem to
```
kim, line 244: Afaik, that'd be [`listenfd`](https://crates.io/crates/listenfd). It is unfortunately cheating on macOS and Win, so proper `launchd` registration would have to be added. The cheat for Win may actually be useful.

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,318 @@
...
+
+The name passed to the process MUST be used to check-in with launchd as
+documented in `launch(3)` which in essence involves obtaining the FDs via
+`launch_activate_socket` expecting a name.
+
+=== Configuration
+
+Common service configuration files SHALL be provided alongside the source code
+of the node binary. To support the semi-dynamic nature of one process per
+profile, facalities are required to manage service definitions along the
```
kim, line 274: Did you mean _fecalities_ :sweat_smile:

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,318 @@
...
+lifetime of a profile and when updated need to inform the process management
+system. Such facalities need to be considered wherever profiles are modified,
+i.e. CLI.
+
+The binary SHALL expose all knobs necessary to fine-tune the internal configs of
+the core protocol, i.e. `membership`, `protocol`, `storage`. Additionally
+switches and configuration that subroutines require. As this surface is sizable
```
kim, line 281: I'd like to see this justified / expanded on a bit more. Config files tend to add more complexity overall, eg. precedence, hot or cold reloading, surfacing syntax errors,...

`systemd` has drop-ins, which power users may want to familiarise themselves with if they're not already (`systemctl edit`). Of course, `launchd` doesn't really have an equivalent, perhaps worth looking at what `homebrew` does there.

docs/rfc/0696-p2p-node.adoc
kim: On second pass, I'm not too enthused about handing out IP addresses in the stats. This should be removed. Stats is about numbers.

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,318 @@
...
+context.
+
+Note: A blocking issue for that is
+https://github.com/radicle-dev/radicle-link/issues/141[#141]
+
+==== Tracking
+
+Given the ability to configure a set of urn, peers, both or anything observed
+from connected peers, the node will automatically track and replicate. This
+strategy is currently in use in the seed node and SHALL be preserved.
```
kim, line 106: Does this mean that the `seed` code is going to be obsoleted?

docs/rfc/0696-p2p-node.adoc
alexjg: As a counterpoint, I have found this information extremely useful for debugging purposes. It's useful to be able to understand problems with connectivity when you can see what the actual network addresses you are connecting (or not connecting) to are. I think this is important for the peer to peer node because we're no longer in a situation where it's easy to ask users to just invoke Upstream (for example) with some additional logging. They would need to change the logging for the peer to peer node (which they may not even be aware of) and then restart the node. This API means the application can directly provide useful diagnostics to the user.

kim: Then go design a debugging facility for authorized personell. We do not conciously expose IP addresses as part of a public interface. There is nothing to discuss about this, unless you are okay with transferring 4% of your yearly revenue to the project legal treasury under GDPR.

docs/rfc/0696-p2p-node.adoc
xla: :hankey: :hankey: :hankey:

docs/rfc/0696-p2p-node.adoc
xla: Was eyeeing Mr. Ronachers libs as well for that purpose but actively refrained from making a recommendation in this document and leave it up to the implementation phase to address it. If you see a need to make this explicit before this document can be accepted I can expand with existing options that seem suitable.

docs/rfc/0696-p2p-node.adoc
xla: Almost all of it will be, the one thing subsumed by the all-might new daemon is the `Tracking` protocol described in the paragraph above - which in turn will be opt-in and has to be explicitly enabled. The rest of the seed code is pretty obsolete with the new architecture and was largely serving the misguided use-case of serving as an API to power pixels showing peer ids and big project numbers. That part is now covered by https://github.com/radicle-dev/radicle-http-api and therefore will not interfere with the operation or requirements of the seed node itself.

docs/rfc/0696-p2p-node.adoc
xla: My angle here was to expose the public introspection methods of `net::peer::Peer` for stats that is `librad::net::protocol::event::downstream::Stats`. This struct has the `conencted_peers` field which is a map of peer_id to `Vec<SocketAddr>`. Given that, would it therefore make sense to remove that information from that struct in the first place. So that the invariant that IP information are not exposed publicly will hold in the future and the generations to follow won't get tempted by the power at their fingertips?

docs/rfc/0696-p2p-node.adoc
xla: Good catch, that part was a bit hastily added when I realised that `Discovery` is an integral part of the protocol orchestra. The intention for `discover` is to make a peer address known to a running daemon, similar to how `StreamDiscovery` works which can react to changes of the configured seed nodes in upstream at runtime. As the underlying operation is streaming in nature I'd assume the response would be a noop of thoughts, unless there is something else that is more indicative wrt the success of the request.

https://github.com/radicle-dev/radicle-link/blob/138813dba6d2cf377865ba163049982cb4c03c1c/daemon/src/config.rs#L93-L132

xla: Frankly, I don't like the terminology of `discover` much, but do think there is the need to manipulate the set of configured bootstrap nodes. If that is an agreeable statement that means it needs more scrutiny if these changes are persistent, if so where and how.

docs/rfc/0696-p2p-node.adoc
xla: Are you referring to the entire section, or just the last paragraph?

docs/rfc/0696-p2p-node.adoc
FintanH: > I don't like the terminology of discover much

Ya, this might get confusing as content discovery becomes a thing. Maybe it should be `peer-discovery`? Or a completely different term :)

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,318 @@
...
+; Note that streaming / multi-valued responses may include the same id in
+; several response messages.
+request-id: bstr .size (4..16)
+
+; Placeholder for future one-time-token support.
+token: bstr
+----
+
+
+=== Request/Response payloads
```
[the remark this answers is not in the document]
xla: > Should the `Requests` action also have a `request`? Or put another way, why doesn't it make the cut?

Took me a couple of reads to unpack these sentences. ^^

It didn't make the cut so far, because the main purpose of the API was to introspect. Which is not to say that using it for other means, i.e. more interactive use-cases is unattainable. What would the `request` request - yo dawg we need to find a better name - do tho?

> I'm also wondering if it might be worth having a request for announcements. I think we might have discussed this offline -- I think we landed on the daemon running announcements when it would come online, however, I think it might be useful to have an explicit mechanism to announce. It could be useful for testing, for example.

Similar question as above what would the request/response look like here, and more importantly what's the interplay with the announcement protocol? Given that it's tailored to only announce deltas, would the request trigger announcements of everything?

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,316 @@
...
+`launch_activate_socket` expecting a name.
+
+=== Configuration
+
+Common service configuration files SHALL be provided alongside the source code
+of the node binary. To support the semi-dynamic nature of one process per
+profile, facilities are required to manage service definitions along the
+lifetime of a profile and when updated need to inform the process management
+system. Such facilities need to be considered wherever profiles are modified,
+i.e. CLI.
```
xla, line 273: @FintanH The author assumes this is significant for the work happening in #698 as it is tight to the profile management. Maybe you can chip in if this should be expanded on in a follow-up RFC which assumes presence of a cli and a daemon or if you see other ways of how this could work.

docs/rfc/0696-p2p-node.adoc
kim: > This struct has the conencted_peers field

Wait what? That's a bug then.


I know this is hard to grasp for the code-is-law crowd, but the truth is that law is a Wittgensteinian business. It is actually very simple: collecting IP addresses in secondary storage is not necessary for providing the service of running a radicle link node. It _is_ necessary, however, to ensure the quality of the service under below-optimum network conditions. We cannot expect the average operator of a radicle link node (be it a publicly advertised node or a "private" laptop node) to be aware of any data protection obligations they acquire implicitly by doing so. 

Therefore, the **default** must be data frugality. Only _when you know what you're doing_, and you're doing it _conciously_, you should be able to access data where explicit consent is somewhat difficult to acquire.

**tl;dr** Stats are a thing we'd provide for people to put on a dashboard. It should consist of numbers. Administrative introspection can be a separate channel, which should preferably be disabled by default, and can be enabled when needed.

xla: > Wait what? That's a bug then.

https://github.com/radicle-dev/radicle-link/blob/138813dba6d2cf377865ba163049982cb4c03c1c/librad/src/net/protocol/event.rs#L56-L63

docs/rfc/0696-p2p-node.adoc
kim: This should go away entirely. The idea of "discovery" is ill-defined, cf #629. The use case of query provider of piece-of-data, acquire piece-of-data can and should be hidden behind the `net::Peer` interface.

docs/rfc/0696-p2p-node.adoc
alexjg: This is something I built when I first started to support the network diagnostics panel, I wasn't aware of the stance on exposing public IP addresses.

docs/rfc/0696-p2p-node.adoc
kim: Then perhaps remove the sentence altogether -- what we need probably doesn't exist pret-a-porter, but if it did, it is obvious that we would try to collaborate.

docs/rfc/0696-p2p-node.adoc
kim: Ok. Request for inclusion of `rm -r ./seed` in the tracking issue of this RFC.

docs/rfc/0696-p2p-node.adoc
kim: I'm suggesting to remove `Motivation`, but keep `Overview`. #682 already used sufficiently dismissive language to motivate the move ^)

docs/rfc/0696-p2p-node.adoc
kim: I found your name in `git blame` already! **/shakesfist** Seems like I missed that change in code review.

docs/rfc/0696-p2p-node.adoc
xla: #722

docs/rfc/0696-p2p-node.adoc
FintanH: It is of the opinion of the CLI author that there may be two routes we can go:
* Profile setting can only be switched if there is no peer-to-peer daemon running
* Or profile setting triggers a reload of the peer-to-peer daemon if it is currently running. Possibly through the PubSub component.

Perhaps there are other options, but these two are the ones that come to mind right now.

alexjg: I guess an alternative option is to have a way to spin up a separate peer-to-peer daemon for each profile. Maybe have the well known socket name of the peer to peer daemon include some indication of the profile ID.

FintanH: Mmmm that's a good point. `PeerId`s and `Profile`s have a 1:1 relationship, but then the whole idea of "active profile" becomes a bit murky.

alexjg: Can we make "active profile" an application concern? E.g have it be an environment variable for the CLI, a file in `<XDG_HOME>/radicle-upstream` for Upstream and so on.

FintanH: Very possible -- the changes here https://github.com/radicle-dev/radicle-link/pull/720 are all in terms of `RadHome` not relying on the active profile. There is a `PROFILE_ID` env var as well, which is currently only considered in `load`, but I've also outlined that [the profile should be a global parameter for subcommand](https://github.com/radicle-dev/radicle-link/pull/698/files#diff-a944b0f221c56c1f789fb8ba1ad84ba9c859dcf7a09c68430b9224769eb538f4R106).

docs/rfc/0696-p2p-node.adoc
xla: To summarise no footgun should be handed out which interferes with the membership protocol as it has potential of incurring churn and undesired behaviour wrt. management of active/passive sets. In the short-term we accept a static set of boostrap nodes as discovery mechanism and mid-term expect the discovery interface to change to be pull based when the network stack has the need for a fresh node to connect to. Does that roughly sum up the state of affairs?

If that is the scenario, I'm happy to drop the discovery protocol from the document - one less thing to worry about - and capture follow up work around discovery in issue(s).

docs/rfc/0696-p2p-node.adoc
alexjg: :flushed:

docs/rfc/0696-p2p-node.adoc
xla: Some clarifications as I think we missed my original point in the ensued discussion. The assumption is that there is a 1:1 relation ship between profile and daemon. Truthfully 1:2 as there could be a node and a seed for any given monorepo. Which means that any system service manager must be configured in a way to run these. If at the same time or not is out of the control of those managers as the corresponding sockets can be activated at any point with no order guarantee - distributed systems in a box, ya know.

With all that in mind, when a profile is created a corresponding configuration for the service managers must be created (detailing how and where to run it) and fed to said managers can have them ready for operation. How the interplay is coming about between the templates of such service configuration, the rendering to an actual file and loading into the service manager is where I'd like to steer the conversation towards. And as I said above that can also happen in a follow-up and we assume for now that such configuration templates exist.

docs/rfc/0696-p2p-node.adoc
xla: With a heavy heart shall I bid goodbye to my beloved prose.

docs/rfc/0696-p2p-node.adoc
xla: 6964dee

docs/rfc/0696-p2p-node.adoc
FintanH: > Truthfully 1:2 as there could be a node and a seed for any given monorepo.

That's not recommended :sweat_smile:  https://github.com/radicle-dev/radicle-link/blob/master/docs/rfc/0682-application-architecture.adoc#peer-to-peer-node
> we RECOMMEND to treat both as separate services, each with their own peer id and persistent state

> With all that in mind, when a profile is created a corresponding configuration (detailing how and where to run it) for the service managers must be created and fed to said managers so it can have them ready for operation.

Ah so for example, we might have `~/.config/radicle-link/42-69-3/protocol/membership.toml`, generated from some template? Is that what you're trying to discuss?

xla: > That's not recommended sweat_smile https://github.com/radicle-dev/radicle-link/blob/master/docs/rfc/0682-application-architecture.adoc#peer-to-peer-node

Thanks for the reminder!

> Ah so for example, we might have `~/.config/radicle-link/42-69-3/protocol/membership.toml`, generated from some template? Is that what you're trying to discuss?

Not really, more like `~/.config/systemd/user/linkd-<peer-id>.socket` and the corresponding invocations to `systemctl reload-daemon` or equivalents on other systems.

FintanH: Hmmm so what does this have to do with profiles? :thinking: Sorry if I'm slow with the uptake here :bow:

xla: The existence of a daemon is bound to the existence of a profile, as they operate over the state associated with the profile/key/peer-id. So only at the point when the profile is created can the service manager be informed that there needs to be a service which manages that daemon. And the parameterisation of the service depends on the outcome of the profile creation and the expectation is that after the creation there is a socket in a well-known location that reacts to IPC calls. Of course symmetrically when a profile is deleted the service manager should be informed of the need to remove the service, but that is not important to the point. Does that help illustrate the issue here? Otherwise I can try to draw something up to contextualise what I'm trying to get at.

alexjg: Taking the systemd version of this as an example there would need to be a unit somewhere describing the socket activated service for each profile. This unit would be created after the profile key and storage are created so that these parameters can be instantiated in the unit. One implication of this is that either you have to `sudo` to create a profile, or you use `systemd --user` etc.

xla: My example above assumes a user unit as it fits the model quite neatly.

kim: Yes yes user unit / Launch*Agent*. Note that it is still a good idea to revoke your own write permissions for security reasons, requiring sudo for modifications

docs/rfc/0696-p2p-node.adoc
xla: 7674dcf

xla: 7674dcf

docs/rfc/0696-p2p-node.adoc
xla: Reworked the requests, to match the `net::Peer` API where there is a call to get connected peers which returns just the list of peer ids in 57e6c29 - also added an item in #722 to remove the offensive field from `Stats` entirely. Speak up if anyone of you see further need to capture the outcome of this discussion.

@alexjg One thing that is apparent from the exchange in this thread is that we should early on think about introspection for _authorized personnel_ and I was wondering if you have - from the experience building that for the current daemon - a list of requirements, constraints, prior art that could inform that.

docs/rfc/0696-p2p-node.adoc
xla: Dropped the discovery protocol in 14c14a4 and created a tracking issue #723 to capture the follow-up work. Anything else we like to see captured for this?

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,318 @@
...
+==== Announcements
+
+Until superseded by graftier ways, currently the main means of dissemination is
+through gossiping new interesting `Have`s. This is achieved by maintainig a
+persisted view on the list of identities and their current refs. During
+periodic runs a delta is created which serves as the basis of the announcements.
+In turn that list is made known to actively connected peers via the core
+protocols gossip.
+
+==== Requests
```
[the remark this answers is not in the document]
FintanH: Perhaps we could call this `Replication` since its main concern is attempting to replicating an `Urn`? Or `Replication Requests` if we want to distinguish it from bare replication.

docs/rfc/0696-p2p-node.adoc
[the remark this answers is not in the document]
FintanH: > What would the request request - yo dawg we need to find a better name - do tho?

So I think this is related to the discussion here https://github.com/radicle-dev/radicle-link/pull/698/files#r660397972. In general, I think it would be useful to have explicit mechanisms for announcements and replication. At least for debugging in a test net scenario it will be immensely helpful. I'm envisioning a scenario where I'm running a test net and announcements aren't working properly and, for example, the announcement protocol isn't going to announce again because it's based on deltas. If I have a mechanism to announce a SHA manually for a Urn then I have more ample opportunity to correlate that explicit call with the logs/tracing.

Now, I realise that the downside is that this opens the door for developers (like you and me in our naivete) to possibly abuse these mechanisms and flood announcements in a shotgun fashion. But at least we could test the rate-limiting then ;)

What do yourself and others think?

docs/rfc/0696-p2p-node.adoc
alexjg: Yeah there's a number of thoughts I have on this. My plan was to put it into an RFC once the CLI and daemon RFCs are merged (so that we're working from a somewhat stable base). My instinct is that we should not have too many in progress RFCs which relate to each other. Sound sensible or should we just write the RFC?

FintanH: > Sound sensible or should we just write the RFC?

Sounds sensible to me, at the same time I'd read a draft :) But I'm sure you'd also like to concentrate on the COBs too

xla: Closing this thread and follow-up work is tracked in #722

docs/rfc/0696-p2p-node.adoc
xla: Closing this thread as #723 captures the conversation wrt. discovery changes.

docs/rfc/0696-p2p-node.adoc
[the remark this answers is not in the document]
xla: @FintanH Would it be sensible to defer tooling like the one you are asking for when we look at the APIs for *authorised personnel* (similar to how we settled to do introspection on top of stabilised p2p node and cli)?

docs/rfc/0696-p2p-node.adoc
xla: @FintanH Gentle ping to see if there is a common understanding of the interplay between profile management and daemon unit files. :dog:

docs/rfc/0696-p2p-node.adoc
xla: Would you consider 8dfb717 sufficient?

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,290 @@
...
+
+; Canonical representation of a peer.
+peer-id: bstr
+
+; Network port.
+port: uint
+----
+
+All types representing requests and responses and their serialisation logic MUST
+be exposed as linkable libraries. It is RECOMMENDED to also expose the
```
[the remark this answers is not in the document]
xla: That's on the money, would you prefer this to be more explicit in the RFC text?

docs/rfc/0696-p2p-node.adoc
xla: Amended in 0158b57.

docs/rfc/0696-p2p-node.adoc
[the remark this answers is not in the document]
xla: Disambiguated in 6233e75.

docs/rfc/0696-p2p-node.adoc
xla: Addressed in 41b6454

As `brew services` is just a hipster facade for manage launch jobs (agents for the logged in user) they leverag `.plist`s in well-known locations, which in turn is just XML with strings determining the location of the executable and the args passed to it. So it is reasonable to expect power users to alter these directly if/when desired.

docs/rfc/0696-p2p-node.adoc
xla: Settling on any specific language is not needed since 41b6454.

docs/rfc/0696-p2p-node.adoc
xla: @kim Does this address your concern?

docs/rfc/0696-p2p-node.adoc
xla: This is now capture as well in #722

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,290 @@
...
+For other processes to be able to talk to the node, it SHALL keep an API
+subroutine running, which offers introspection into its state. See the dedicated
+section for the surface the node will expose.
+
+==== Announcements
+
+Until superseded by <<pr-653, graftier ways>>, currently the main means of
+dissemination is through gossiping new interesting `Have`s. This is achieved by
+maintainig a persisted view on the list of identities and their current refs.
+During periodic runs a delta is created which serves as the basis of the
```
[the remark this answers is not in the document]
xla: Is 4184bf8 sufficient?

docs/rfc/0696-p2p-node.adoc
FintanH: Just re-read this thread. Is the summary that unit files should be stored in a profile? And should be part of the `rad profile` setup logic? :)

xla: > Just re-read this thread. Is the summary that unit files should be stored in a profile? And should be part of the `rad profile` setup logic? :)

Where exactly they should be stored we can suss out in the RFC mentioned in #722, but generally speaking yes this is the takeaway.

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,302 @@
...
+Besides the core protocol other functionality will be encapsulated in protocols
+which will be driven by a stream of well-defined inputs. The well-defined
+outputs of these protocols are used for effect management.
+
+=== Subroutines
+
+This section expands on the auxilliary routines driven by the node on top of the
+core protocol. Where each SHOULD allow to be disabled through configuration. A
+subroutine acts as reactor driving the protocols by feeding inputs - usually
+results of effects - into them and informed by the outputs schedules new
```
FintanH, line 56: This sentence could use some reworking. I'm not sure what you're trying to say about scheduling new effects :)

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,302 @@
...
+
+Until superseded by <<pr-653, graftier ways>>, currently the main means of
+dissemination is through gossiping new interesting `Have`s. This is achieved by
+maintainig a persisted view on the list of identities and their current refs.
+During periodic runs a delta is created which serves as the basis of the
+announcements. In turn that list is made known to actively connected peers via
+the core protocols gossip.
+
+The delta mechanism works by recording a snapshot of the state of the monorepo,
+specifically the refs and on the next iteration diffing against the current
```
FintanH, line 75:
```suggestion
specifically the refs, and on the next iteration diffing against the current
```

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,302 @@
...
+same approach as for announcements where offline changes to the storage can be
+picked up at any point the node is running and online, this subroutine will
+maintain a persisted view and periodically (potentially provoked by a request
+over the <<API>>) build the delta for new tracked `Urn`s and remotes in their
+context.
+
+Note: A blocking issue for that is
+https://github.com/radicle-dev/radicle-link/issues/141[#141]
+
+==== Tracking
```
FintanH, line 92: Will this API be dynamic or will it only be specified at start up?

docs/rfc/0696-p2p-node.adoc
FintanH: Happy for you to mark this as resolved from my side at least :)

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,302 @@
...
+Common service configuration files SHALL be provided alongside the source code
+of the node binary. To support the semi-dynamic nature of one process per
+profile, facilities are required to manage service definitions along the
+lifetime of a profile and when updated need to inform the process management
+system. Such facilities need to be considered wherever profiles are modified,
+i.e. CLI.
+
+The binary SHALL expose all knobs necessary to fine-tune the internal configs of
+the core protocol, i.e. `membership`, `protocol`, `storage`. Additionally
+switches and configuration that subroutines require. The configuration surface
```
FintanH, line 263:
```suggestion
any switches and configuration that subroutines require. The configuration surface
```

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,302 @@
...
+
+Common service configuration files SHALL be provided alongside the source code
+of the node binary. To support the semi-dynamic nature of one process per
+profile, facilities are required to manage service definitions along the
+lifetime of a profile and when updated need to inform the process management
+system. Such facilities need to be considered wherever profiles are modified,
+i.e. CLI.
+
+The binary SHALL expose all knobs necessary to fine-tune the internal configs of
+the core protocol, i.e. `membership`, `protocol`, `storage`. Additionally
```
FintanH, line 262:
```suggestion
the core protocol, i.e. `membership`, `protocol`, `storage`. Additionally,
```

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,302 @@
...
+switches and configuration that subroutines require. The configuration surface
+SHALL be exposed as command line arguments, until further evidence is brought
+forward which makes a strong case for external config files.
+
+== Key Access
+
+Access to key material SHALL be done through the facilities provided by
+`<<radicle-keystore>>`. Except for debug/development purpose this SHOULD be
+limited to the use of the `ssh-agent` expected to land in
```
FintanH, line 271: The eagle has landed

FintanH (dismissed):
This looks good to me in general. I have some minor suggestions to grammar and language to help reading it :) And I think one or two questions :eyes:

docs/rfc/0696-p2p-node.adoc
kim: Is our common understanding that by facilities, we mean that our own tooling is able to perform the platform-specific (re-)configuration? As in, wrappers for systemd incantations.

Maybe we can say this more explicitly then :) It reads a little handwavy.

docs/rfc/0696-p2p-node.adoc
xla: So far I thought of it as purely static configuration at start-up, yet use-cases in other parts of the ecosystem could benefit from a dynamic API for it. It opens up a bit can worm wrt. to persistence of such changes.

xla: Added a point to #722, to keep the scope tight for this RFC, is that acceptable for now or do you want it to be addressed in here?

FintanH: Works for me :+1:

docs/rfc/0696-p2p-node.adoc
xla: 76975d1

docs/rfc/0696-p2p-node.adoc
xla: 884b5e5

docs/rfc/0696-p2p-node.adoc
xla: abb9f67

docs/rfc/0696-p2p-node.adoc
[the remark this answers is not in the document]
alexjg: That should do it.

docs/rfc/0696-p2p-node.adoc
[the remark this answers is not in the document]
alexjg: I think this is clear enough. I was just checking my understanding was correct.

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,300 @@
...
+
+This RFC expands on the peer-to-peer node outlined in <<rfc-0682, RFC 0682>> to
+propose a new stand-alone daemon with focus on driving the p2p stack, exposing a
+minimal API and defer supervision to more integrated and mature system services.
+Furthermore it will prove that all daemons are made equal and only distinct with
+regards to the configuration given (e.g. tracking behaviour) and lifetime of the
+process (i.e. socket-activated or long-running).
+
+It also strives to recommend APIs and mechanisms -- in the form of non-core
+protocols - which remove the need for online requests made by clients to a
```
FintanH, line 24:
```suggestion
protocols -- which remove the need for online requests made by clients to a
```

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,300 @@
...
+information in form of synchrounous calls and events obtainable internally via
+the sufficiently space-engineered tincans SHALL be made available.
+
+Besides the core protocol other functionality will be encapsulated in protocols
+which will be driven by a stream of well-defined inputs. The well-defined
+outputs of these protocols are used for effect management.
+
+=== Subroutines
+
+This section expands on the auxilliary routines driven by the node on top of the
```
FintanH, line 53:
```suggestion
This section expands on the auxiliary routines driven by the node on top of the
```

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,300 @@
...
+core protocol. Where each SHOULD be allow to be disabled through configuration.
```
FintanH, line 54:
```suggestion
core protocol. Where each SHOULD be allowed to be disabled through configuration.
```

docs/rfc/0696-p2p-node.adoc
```diff
@@ -0,0 +1,300 @@
...
+A subroutine acts as reactor driving a protocol by feeding inputs - usually
```
FintanH, line 55:
```suggestion
A subroutine acts as a reactor driving a protocol by feeding inputs - usually
```

docs/rfc/0696-p2p-node.adoc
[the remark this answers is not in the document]
FintanH:
```suggestion
During periodic runs, a delta is created which serves as the basis of the
```

FintanH (dismissed):
A few more suggestions for :gb: Happy otherwise :blush:

Issue #141, opened 2020-05-12
Track project while disconnected


kim:
Currently, `git::repo::clone` requires a `PeerId`, and an established connection to that peer. It should also be possible to schedule a known identity hash to be cloned later, for example while offline.

Practically, this would create an empty repository (or namespace), which would technically be invalid, but the fetch/clone machinery should be able to detect and handle that case. It may also entail defining a startup-sequence for the networking stack, which would emit `want` and `have` gossip messages as appropriate (queued in secondary storage, or read from the repository state).

FintanH:
I started thinking about the initial part of this about creating the empty state.

If we're offline and only have access to the storage and we're trying to track `<project_id>` from `<remote>` we could do the following:
* Create `refs/namespaces/<project_id>/refs/rad/self`
* Create `refs/namesapces/<project_id>/refs/remotes/<remote>/rad/id`

The above means that:
a) We have linked our `rad/self` to the project's `rad/self`
b) Won't have a `rad/id` ourselves for the project
c) The `rad/id` of the remote is empty

The idea being that we uniquely identify this case, and when we're online we can query the `RadUrn` + `PeerId` and replace the remote's `rad/id` with the actual info, and also finish the setup on our own local references.

In a way, it seems a bit hacky to have this "special case". But I'm not sure how to create the references otherwise without some `Oid`. 

As I'm typing this I'm wondering if maybe `Oid::zero` would work? :thinking:

FintanH:
Implemented in https://github.com/radicle-dev/radicle-link/commit/cc7d348ff79f338c8136903393fedd7f098665d6

PR #698, opened 2021-06-10
RFC: RadiCLI

FintanH:
Proposal of implementing the `rad` binary as well as the core set of
subcommands that will be provided.

[Rendered](https://github.com/radicle-dev/radicle-link/blob/finto/rfcli/docs/rfc/0698-cli-infrastructure.adoc)


docs/rfc/0000-cli-infrastructure.adoc
```diff
@@ -0,0 +1,254 @@
...
+defined:
+
+* `create` — create a `Project` identity and either create a
+  respective working copy, or link it to an existing working copy.
+* `get` — attempt to retrieve a `Project` identity from the store based on
+  a provided URN.
+* `list` — list all `Project` identities in the store.
+* `update` — update a `Project` identity with a new payload.
+* `merge` — merge a peer's version of the `Project` identity with your own
+* `track` — track a peer for the `Project` identity and add the peer
```
FintanH, line 213: I'm going back and forth on having "tracking" be part of `Project`/`Person` and having a separate component that does this. 

I'm also questioning the working copy being only associated with `Project` after the brief discussion:

> fintohaps:
> Had a thought today: if a `Person` identity was "checked out" as a working copy you could store something like a static webpage in `heads` and then that could be something like your personal blog :)

> kim:
> The difference between Person and Project is only in the delegation rules, and whether some metadata makes sense or not. Otherwise, a namespace is a git repository, and you can put anything you want in there.

FintanH, line 213: Some other things to be added are:
* Listing peers
* Listing delegates
* Get `rad/self`

FintanH: There might be more that I'm forgetting =]

docs/rfc/0000-cli-infrastructure.adoc
```diff
@@ -0,0 +1,254 @@
...
+
+Cons of *2.*:
+
+* Makes it easier to override one of the core commands.
+* Many binary must be packaged and distributed.
+
+The author recommends *2.*. The modularity provides a clean separation
+of concerns and avoids a monolothic approach. We assume good faith
+of overrides not happening [TODO: is there some way we can just
+disallow this programatically?]. For packaging and distribution of the
```
kim, line 165: Instead, you may detail how development can be made more convenient. Eg. make it so development builds always produce binaries with a `-dev` suffix, and provide `systemd`/`launchd` definitions so they either call `cargo build` or reference the `-dev` binaries.

docs/rfc/0000-cli-infrastructure.adoc
```diff
@@ -0,0 +1,254 @@
...
+libraries. However, this leaves the question of where the binary might
+live. So far, it has been an unspoken formalism that all binaries live
+in `radicle-bins` where we can provide a `Cargo.lock` file. All
+functionality will live in `radicle-link`, including the `main`
+function, and the entry in `radicle-bins` will be a thin wrapper
+around this function.
+
+=== Distribution
+
+We are noting here that -- at the time of writing -- we cannot publish
```
kim, line 182: Is this relevant? We are relying on the system package manager either way.

docs/rfc/0000-cli-infrastructure.adoc
```diff
@@ -0,0 +1,254 @@
...
+
+== Core
+
+With all the details in place, we define a non-exhaustive list of the
+core commands that we wish to include for the first iteration. More
+components MAY be added as the result of future RFCs. We SHALL keep a
+list of the core commands as part of the `radicle-link` documentation
+— as part of the `README`.
+
+=== `rad-project`
```
kim, line 200: I don't think project/person as separate subcommands is a smart idea, because that will just repeat most of the commonalities. It also prevents extensibility without having to change the CLI for the extension specifically. 

A more maintainable approach would be to either disambiguate with a `--type` argument, or add another layer of sub-command-ness (eg. `rad-identities project [create|...]`). Or else, move CRUD-style verbs to the top-level, like `rad create project [opts]`.

docs/rfc/0000-cli-infrastructure.adoc
```diff
@@ -0,0 +1,254 @@
...
+* `get`— attempt to retrieve a `Person` identity from the store based
+  on a provided URN.
+* `list`— list all `Person` identities in the store.
+* `update`— update a `Person` identity with a new payload.
+* `merge`— merge a peer's version of the `Person` identity with your own.
+* `track`— track a peer for the `Person` identity.
+
+=== `rad-replicate`
+
+* `ask` — ask the network for a given URN and replicate it to local storage.
```
kim, line 231: Those are two things.

docs/rfc/0000-cli-infrastructure.adoc
```diff
@@ -0,0 +1,254 @@
...
+replication of a URN from the network.
+
+This leaves us in a predicament of thinking about how multiple Radicle
+applictions may interact with each other. For example, if Upstream is
+running and I create a project via the CLI, then a good user
+experience is that the Upstream is updated with the new project
+(without refreshing). In RFC-682, it is alluded that a PubSub
+component could solve this, however, that requires the daemon to be
+online. The author prefers not to specify the PubSub component
+here. To bridge the gap, we recommend that a subcommand can add a
```
kim, line 109: This is not necessary, and arguably not something the user can or should know. The daemon will start _automatically_ as soon as a command requires it, via socket activation.

docs/rfc/0000-cli-infrastructure.adoc
```diff
@@ -0,0 +1,254 @@
...
+(nor in further RFCs).
+
+== Global Parameters
+
+Many subcommands will likely want to share a set of options. These
+options, dubbed as "global parameters" here, will be a defined set --
+exported from `radicle-link` for re-use -- that the base `rad` command
+will always receive and pass onto the subcommand being called.
+
+Thus, all subcommands SHALL receive these global parameters as part of
```
kim, line 81: It would be good to detail _how exactly_ they receive those parameters. Builtin commands are obviously just function calls, but how is this handled for `execve`'d binaries? Does the `rad` wrapper parse (and validate) those options? Does it then just output them verbatim to the subcommand? Does it prefix them with `--global-` or something? Is order significant? Or does it use some other format like `CBOR` or `JSON`?

docs/rfc/0000-cli-infrastructure.adoc
```diff
@@ -0,0 +1,254 @@
...
+
+To allow the ecosystem of commands to be extended by ourselves and
+others in the community it will be necessary to extend the `rad`
+command with further subcommands. The details of this are outlined in
+RFC-682#CLI, however, we will give a quick overview here for the sake
+of the reader.
+
+Taking inspiration from tools such as `git` and `cargo`, we propose
+that the `rad` binary can be extended by placing another
+binary on the users `$PATH` (or the Windows equivalent). This
```
kim, line 69: `%PATH%`

docs/rfc/0000-cli-infrastructure.adoc
```diff
@@ -0,0 +1,254 @@
...
+initially, unless we build the binaries and distribute them from a
+well-known source.
+
+== Core
+
+With all the details in place, we define a non-exhaustive list of the
+core commands that we wish to include for the first iteration. More
+components MAY be added as the result of future RFCs. We SHALL keep a
+list of the core commands as part of the `radicle-link` documentation
+— as part of the `README`.
```
kim, line 198:
```suggestion

```

docs/rfc/0000-cli-infrastructure.adoc
```diff
@@ -0,0 +1,254 @@
...
+
+=== `rad-project`
+
+This command will provide management of the `Project` identities and
+their respective working copies. The following actions will be
+defined:
+
+* `create` — create a `Project` identity and either create a
+  respective working copy, or link it to an existing working copy.
+* `get` — attempt to retrieve a `Project` identity from the store based on
```
kim, line 208: What is the output encoding of this and similar commands? Should the user be able to choose (eg. `--output=[json,cbor,java.io.Serializable]`)?

docs/rfc/0000-cli-infrastructure.adoc
```diff
@@ -0,0 +1,254 @@
...
+  a provided URN.
+* `list` — list all `Project` identities in the store.
+* `update` — update a `Project` identity with a new payload.
+* `merge` — merge a peer's version of the `Project` identity with your own
```
kim, line 212: This should receive some more thought (apart from `merge` not yet footgun-free): it is possibly preferable to force the user to _review_ what is being accepted, unless a `--force` flag or something is given.

kim (commented):
I am not sure I can approve this RFC without a proper gollum reference

docs/rfc/0000-cli-infrastructure.adoc
```diff
@@ -0,0 +1,254 @@
...
+  modular commands would mirror this structure well.
+* Subcommands can be versioned independently of each other.
+* We can dogfood our subcommand functionality
+
+Cons of *2.*:
+
+* Makes it easier to override one of the core commands.
+* Many binary must be packaged and distributed.
+
+The author recommends *2.*. The modularity provides a clean separation
```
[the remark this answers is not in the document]
FintanH: The author was on the fence, so happy to recommend 2 =]

docs/rfc/0000-cli-infrastructure.adoc
FintanH: Ah so, there would be the `rad` binary, but then there would also be a `rad-key-dev` binary, for example?

docs/rfc/0000-cli-infrastructure.adoc
FintanH: Good point, I had that written down in one of my notes from way back, so thanks for reminding me ^_^ I think it should be like your example says: `--output=` and have a series of options. You forgot `dhall` :P

docs/rfc/0000-cli-infrastructure.adoc
```diff
@@ -0,0 +1,237 @@
...
+RFC-682 set the stage for defining the architecture of applications
+built on top of `<<radicle-link>>`. This RFC wishes to build on top of
+this specification and fill in the gaps around the CLI component and
+its organisation. As dogfooding our own solutions is on our long-term
+roadmap, this team wants to provide tooling that they will maintain
+and use themselves.
+
+== Overview
+
+This proposal will outline an binary called `rad`, what it will
```
xla, line 25:
```suggestion
This proposal will outline a binary called `rad`, what it will
```

docs/rfc/0000-cli-infrastructure.adoc
```diff
@@ -0,0 +1,237 @@
...
+exported from `radicle-link` for re-use -- that the base `rad` command
+will always receive and pass onto the subcommand being called.
+
+Thus, all subcommands SHALL receive these global parameters as part of
+their interface. We list these parameters here, but it is NOT an
+exhaustive list:
+
+* `verbose`
+* `quiet`
+* `profile := active (default) | <profile-id>`
```
xla, line 87: How does this relate to location of the monorepo (i.e. `RAD_HOME`)? Surfacing the relationship would be interesting in the context of this RFC.

docs/rfc/0000-cli-infrastructure.adoc
```diff
@@ -0,0 +1,237 @@
...
+In RFC-682#Process-Orchestration, it was decided that the peer-to-peer
+damon would not be a long-running process. The terminology we will use
+here is that the daemon is *online* if it is running and connected to
+the network. Otherwise, if it is not running then it is *offline*.
+
+This disctinction becomes important when thinking about the
+functionality of `rad` subcommands. The majority of subcommands that
+interact with the storage can be done while the daemon is offline. The
+only action the author can currently think of — that is online — is
+replication of a URN from the network.
```
xla, line 100: Is that due to how the current "waiting room" is implemented as an online API call to issue a "request" for an URN? In the spirit of the p2p node converged over the state of the storage is there a potential way forward where requesting/replicating a URN from the network could be done through pure manipulation of the storage in contrast to rely on a running node process?

docs/rfc/0000-cli-infrastructure.adoc
```diff
@@ -0,0 +1,237 @@
...
+
+This leaves us in a predicament of thinking about how multiple Radicle
+applictions may interact with each other. For example, if Upstream is
+running and I create a project via the CLI, then a good user
+experience is that the Upstream is updated with the new project
+(without refreshing). In RFC-682, it is alluded that a PubSub
+component could solve this, however, that requires the daemon to be
+online. The author prefers not to specify the PubSub component
```
xla, line 108: Could be I miss an angle here, but the PubSub as described in 682 is dependent on an "online" p2p node. In contrast it specifically recommends a substrate which would be a system service. Or is your meaning here that the node needs to be running to observe the message from the PubSub?

docs/rfc/0000-cli-infrastructure.adoc
```diff
@@ -0,0 +1,237 @@
...
+`ssh-agent` using the supplied constraints, e.g. ask for password.
+. `radicle-link#461` to have a read-only storage.
+
+== Library/Command Organisation
+
+As proposed here, `radicle-link` will provide a core set of
+functionality. Something this document wishes to standardise is the
+organisation of these components. The author sees two options:
+. One binary that defines the core subcommands
+. Many binary that take advantage of the subcommand mechanism
```
xla, line 138: Dunno if GH tripping up or if the formatting generally doesn't produce a proper list here.

xla, line 138:
```suggestion
. Many binaries that take advantage of the sub-command mechanism
```

docs/rfc/0000-cli-infrastructure.adoc
```diff
@@ -0,0 +1,237 @@
...
+
+=== `radicle-bins`
+
+Per RFC-682#CLI, all subcommands will be provided as
+libraries. However, this leaves the question of where the binary might
+live. So far, it has been an unspoken formalism that all binaries live
+in `radicle-bins` where we can provide a `Cargo.lock` file. All
+functionality will live in `radicle-link`, including the `main`
+function, and the entry in `radicle-bins` will be a thin wrapper
+around this function.
```
xla, line 174: It would be nice to scrutinise this more and challenge if this played out well for the seed.

docs/rfc/0000-cli-infrastructure.adoc
xla: Would it be possible to track a URN and have the p2p stack take care of replicating it from the network?

xla (commented):
> It's a dangerous business, Finto, going out your door. You step onto the road, and if you don't keep your feet, there's no knowing where you might be swept off to.

docs/rfc/0000-cli-infrastructure.adoc
FintanH: > Would it be possible to track a URN and have the p2p stack take care of replicating it from the network?

There's an issue for that actually =] https://github.com/radicle-dev/radicle-link/issues/141

docs/rfc/0000-cli-infrastructure.adoc
FintanH: Thanks for the spot :) I'm going to stick with "subcommand" and "subcommands" because I think it's pretty well recognised

FintanH: Ya, I need a new line to activate the bulleting :)

docs/rfc/0000-cli-infrastructure.adoc
FintanH: Ya, I was thinking about this. I think there's no need to surface `RAD_HOME` since:
a) it can be set on the command-line, i.e. `RAD_HOME=/home/happy/foo rad key create`
b) the general case should be that `XDG_DATA_HOME` is used, and `RAD_HOME` is used in special cases. Contrasting that with the profiles, which can also be set via `RAD_PROFILE`, these will potentially be created and set more frequently

docs/rfc/0000-cli-infrastructure.adoc
FintanH: Well, fundamentally the point is that you need to actually have the daemon running so that it can replicate from another peer (waiting room or not).

As pointed out in the discussion https://github.com/radicle-dev/radicle-link/pull/698/files#r649924840 there are potentially two steps to the process. We do an offline track of a new URN, and when we're online it finds providers and replicates if it finds any.

So I guess the question is: Are there any online commands or is it all offline and the daemon knows what to do when it comes online via the "queded tasks" that we made while offline?

docs/rfc/0000-cli-infrastructure.adoc
FintanH: > In contrast it specifically recommends a substrate which would be a system service.

I'm not sure what you mean here. Could you maybe rephrase it?

> Or is your meaning here that the node needs to be running to observe the message from the PubSub?

Ya, exactly. In the example, if I created a project via the CLI and I also have Upstream running (without the daemon running), then there's no way for Upstream to know that an event happened (the project being created).

docs/rfc/0000-cli-infrastructure.adoc
FintanH: > It also prevents extensibility without having to change the CLI for the extension specifically.

Can you elaborate on what you mean here?

The rest makes sense, but I'm still mulling over what the consequences of one option over the other might be :) I think I'll need to sketch them out to see which makes more sense

docs/rfc/0000-cli-infrastructure.adoc
FintanH: I (hopefully) addressed this in https://github.com/radicle-dev/radicle-link/pull/698/commits/8a3d264dfbd3fd44cef3b27f29c3d7b07536a73c

docs/rfc/0000-cli-infrastructure.adoc
kim: I mean that extension payloads are basically opaque JSON blobs. One may also want to pipe an entire payload as JSON, and have the CLI infer the type. Or maybe some new identity kind was introduced, but the local CLI installation doesn’t know about it yet (disregard for a moment that the replication logic needs to know about it — maybe that namespace was obtained by some other means). We could still make it so the CLI is able to display _something_.

docs/rfc/0000-cli-infrastructure.adoc
kim: Contributions welcome. Essentially, we could just treat a remote without a `url` as a clone intent. Requires some work to decouple from libgit2

docs/rfc/0000-cli-infrastructure.adoc
kim: If upstream is running, it should connect to the pubsubbub, onto which the CLI places a message after creating the project. No daemon involved

docs/rfc/0000-cli-infrastructure.adoc
kim: So if I’m not careful, I might accidentally expect a parameter which is already global 🤔 

Follow up question: is `println` the only contract for program output?

docs/rfc/0000-cli-infrastructure.adoc
kim: Thinking about it, maybe the RFC is not the right place to talk about this (the same thing could also be achieved using `nix` and / or `direnv`). Maybe just assume proper packaging.

docs/rfc/0000-cli-infrastructure.adoc
kim: `sexp`

docs/rfc/0000-cli-infrastructure.adoc
kim: Perhaps. But you may still want to `start` / `stop` / `Status` it =>

docs/rfc/0000-cli-infrastructure.adoc
FintanH: I thought we hadn't landed on whether pubsubflub was a standalone service or if it was only up along with the daemon? If it's always available then I guess this section is pointless :upside_down_face: But it might be useful to have a section on events being emitted, what do yis think?

docs/rfc/0000-cli-infrastructure.adoc
FintanH: > But you may still want to start / stop / Status it =>

Right, I was holding off on any specification for daemon control commands, but I think this has convinced me that we want to define them here :) 

It also feels like it should be a requirement that all commands work offline and the daemon can perform the tasks when it's online. The following diagram must commute:
image
(image created using [quiver](https://q.uiver.app/))

docs/rfc/0000-cli-infrastructure.adoc
FintanH: Ah that's clear, thanks for the insight!

docs/rfc/0000-cli-infrastructure.adoc
FintanH: > So if I’m not careful, I might accidentally expect a parameter that is already global

RTFM :grimacing: 

Ya, I was thinking what's the point of a global parameter if someone is just going to ignore it and use their own meaning of it. But given some more thought, it would be a Bad Thing:tm: if their behaviour was overridden by the globals. However, if `rad` prefixed them then it would need to pass both parameters, i.e. `--global-profile=abc` and `--profile=abc`, just in case there was a non-global version of the parameter defined. One would hope that wouldn't mess up a parser :crossed_fingers:

> Follow up question: is println the only contract for program output?

It hadn't crossed my mind. The other contract of the output would be the format. Is there anything else that I should think about here?

kim: > The other contract of the output would be the format.

Yeah that's the thing. It did cross my mind that @cloudhead (_just_ as an example) might fancy some ASCII spinners, emoji admonitions or whatever. Now, where would you override those? And how do you adhere to the style in an external subcommand?

Maybe this kind of thing is just not feasibly possible.

FintanH: Right, I think we should only be committing to plumbing commands here for now. I would say external commands can add plumbing on top of that and it's up to them how they mangle `stdout` :smile:

cloudhead: Hmm, what if globals were passed as env vars to the sub-command, after being parsed?

docs/rfc/0000-cli-infrastructure.adoc
kim: I think we weren't sure what the right tool is (D-Bus being rather baroque). But I'm fairly sure that we want some sort of standalone thing, because it can become confusing quickly to see what actually happened. For example:

Say you unplugged your WiFi cable, so whenever the daemon comes up, it won't be able to connect to the network. Now say you have Upstream running, but you do something using another program (eg. push some changes using `git`). What should happen?

One option is that the daemon keeps running (because Upstream is connected to it), gets notified by `git` that something happened, and then sends `updated-refs ...` to Upstream, followed by `not-connected`. The other option is that the daemon exits when it can't connect (a few times). It will miss the notification from `git`, and Upstream will just wait in vain for some kind of event which tells it that something was replicated to the network.

Both seem pretty terrible (as are most "event driven" systems) -- you never know if and when the next message arrives, and "no message" can carry information. We know, however, that we need to record the modification of the repo _either way_, such that the daemon can pick it up once it gains connectivity.

The missing piece here -- and this is not the first time I'm saying this -- is that we can't have "offline first" and "it behaves like GitHub" at the same time. In other words, the act of synchronising the local state with the network must be controllable by the user (and is genuinely an "online" operation, btw). 

So in the example: 

If Upstream is running, receives a notification that the repo state has changed, and decides to pretend "it behaves like GitHub", it is _Upstream's_ responsibility to call a `sync` RPC on the daemon, and synchronously wait for the result. That `sync` RPC obviously needs to be idempotent, and it should probably be possible to retrieve information about when and what the last sync did.

If Upstream is _not_ running, things are slightly more complicated, because we can't have `git push` to also atomically `sync` (the refs must have been written already, and there's no atomic rollback). We can do this non-atomically, though, and just funnel back the result to `git` through the sideband. A `rad git push` wrapper would allow to make it optional (by either passing `--sync` or not).

So in summary, there are zero assumptions one can have about the availability of other processes in the service mesh. Therefore, events are always ephemeral, and it is up to the receiver to act upon them, not the sender. Actions which return a result must be synchronous RPCs. 

A consequence of this is that there must be ways to synchronously inspect state / history in case events were missed. The tricky thing about this is to beware of side-effects: say you want to get some info about the most recent `sync`. If you activate the daemon for that, it will have the side-effect of also attempting to connect to the network -- which may be pointless, because of the WiFi cable situation. In that sense, journaling event logs to disk as @alexjg  suggested might be worth considering.

docs/rfc/0000-cli-infrastructure.adoc
```diff
@@ -0,0 +1,264 @@
...
+* `ask` — ask the network for a given URN and replicate it to local storage.
+
+=== `rad-key`
+
+This command will provide the management of keys and the `Profile`
+construct in `librad`. The following actions will be defined:
+
+* `create`— generate a new key and a new `Profile`, which implies a
+  separate storage.
+* `profile get`— get the current `Profile`'s identifier
```
cloudhead, line 227: Are these sub-sub-commands? Or is this like, `rad profile-get`? I wonder if for plumbing commands we should keep it simple with only one level of sub-commands, like git, eg. `git branch -D` to delete a branch, not `git branch delete`

FintanH: What's the advantage of one over the other?

docs/rfc/0000-cli-infrastructure.adoc
alexjg: Yeah this is increasingly looking to me like we need a log and not a pubsub system. With a log we don't need to make any assumptions about what systems are running when producing or consuming events, we don't need to abstract over any platform capabilities, and we can more easily implement transactional semantics where we need to.

The downside is that consumers become more complex, but not enormously so I think, we don't need to build Kafka. Consumers just need to store an  offset and handle log file rotations when reading from the log files.

kim: That is phrased _slightly_ on the optimistic side 😅

alexjg: Ha, I suppose it is. I'm not suggesting that it's a trivial endeavour to build a log, just that a) it's not the same challenge as building a distributed log and b) I think that architecturally it concentrates the complexity in one place, rather than having the complexity of handling ephemeral messages and disappaering services spread around the various components of the system.

kim: Yeah I’m with you. My concern is just that reliable message buses typically lead to YOLOKAFKA syndrome, meaning that everyone is relying on this component to never lose any data.

I’d favour to first try and _reduce_ the amount of bookkeeping required. For example, once #653 lands, it would not strictly be necessary to remember what went out previously. You can just force a sync with your bootstrap nodes, and let git figure out if any work needs to be performed.

Perhaps we should start with pubsub = `ff01::1`?

docs/rfc/0000-cli-infrastructure.adoc
kim: May, maybe it would still be good to define some kind of machine-readable output format at some point, which can be turned on with `—emoji-bufs` or something.

docs/rfc/0000-cli-infrastructure.adoc
FintanH: Thanks for the elaboration and discussion! I think I misinterpreted a previous conversation and was under the assumption the PubSub component was inherently connected to a running daemon, but as you've pointed out that won't be the case.

So in terms of this RFC, it seems like the assumption that will be made is that there will be a PubSub component. If a CLI subcommand deems an event as interesting to track it SHALL publish this event to the PubSub, following the guidelines of the future PubSub RFC. With a caveat being that "interesting" doesn't fall into YOLOKAFKA syndrome -- I'm not entirely sure how we can prevent that but I suppose since it's the consumer's responsibility to sync then they are also responsible for ignoring message. Does that make sense?

docs/rfc/0000-cli-infrastructure.adoc
cloudhead: I guess the git-style namespacing has these advantages:

* It's clear what's a sub-command vs just a parameter to a command, so it may lead to less ambiguities, eg. `git push origin master` is `git-push origin master` and not `git-push-origin master` or `git-push-origin-master`, especially for commands that accept positional arguments.
* You don't need to handle cases like `rad profile --flag get` vs. `rad profile get --flag`

So basically it's simpler from an implementation perspective to only have 1 level of sub-commands. The advantages of having multiple levels are:

* For CLIs with hundreds of sub-commands and a clear hierarchy, eg. `gcloud`, it may be easier to namespace everything
* It "reads" better maybe or has the promise of being more user-friendly, though I'd say that is subjective

FintanH: I've leaned towards the hierarchy of commands. I feel that subcommands are easier on the cognitive load rather than trying to remember what shorts were given for a command. I was also thinking that tab completion would work better for subcommands, which isn't possible with flags (or at least not that I know of).

docs/rfc/0698-cli-infrastructure.adoc
```diff
@@ -0,0 +1,299 @@
...
+their interface. We list these parameters here, but it is NOT an
+exhaustive list:
+
+* `global-verbose`
+* `global-quiet`
+* `global-profile := active (default) | <profile-id>`
+
+== Online & Offline
+
+In <<arch-orch, RFC-682#Process-Orchestration>>, it was decided that the peer-to-peer
```
xla, line 110: The takeaway is that it can't be expected to be a long-running process and assumptions based on it won't hold true. Pointing this out to highlight that some malkovichs will change the way the system service manager is running the daemon.

docs/rfc/0698-cli-infrastructure.adoc
```diff
@@ -0,0 +1,299 @@
...
+damon would not be a long-running process. The terminology we will use
+here is that the daemon is *online* if it is running and connected to
+the network. Otherwise, if it is not running then it is *offline*.
+
+This disctinction becomes important when thinking about the
+functionality of `rad` subcommands. The majority of subcommands that
+interact with the storage can be done while the daemon is offline. The
+only action the author can currently think of — that is online — is
+replication of a URN from the network.
```
xla, line 119: I think this was highlighted in another thread - which seems to be lost now in GHs diff view :shrug: - #114 would allow for all operations being "offline-first" and could be alleviate as a blocking task to realise this grand CLI/daemon vision.

docs/rfc/0698-cli-infrastructure.adoc
```diff
@@ -0,0 +1,299 @@
...
+* Subcommands can be versioned independently of each other.
+* We can dogfood our subcommand functionality
+
+Cons of *2.*:
+
+* Makes it easier to override one of the core commands.
+* Many binaries must be packaged and distributed.
+
+The author recommends *1.*. As mentioned, we have a hard requirement
+of not allowing the core commands to be overrided. In *2.* this is
```
xla, line 180:
```suggestion
of not allowing the core commands to be overridden. In *2.* this is
```

docs/rfc/0698-cli-infrastructure.adoc
```diff
@@ -0,0 +1,299 @@
...
+* `rad-refs [ self | signed | delegates | delegate ]` -- retrieve the
+  blobs that live under `rad/<refname>`
+* `refs [ heads | notes | tags | category ]` -- list the reference names
+  under the respective category, where `category` is a user supplied category.
+
+=== `rad-replicate`
+
+* `ask` -- tell the `daemon` that you wish to replicate the
+  given URN. Pending <<rl-141, radicle-link#141>>.
+* `replicate` -- if the peer/address are known for the given URN,
```
xla, line 232: If `ask` exist as proposed, what are the scenarios we want `replicate`?

xla: Furthermore with the proposed command structure will this result in `rad replicate replicate`?

image

docs/rfc/0698-cli-infrastructure.adoc
```diff
@@ -0,0 +1,299 @@
...
+  passed in. *Note* that this should refuse to do so if the daemon is online.
+* `list`— list the available `Profile` identifiers
+* `ssh-add`— add the current `Profile`'s key to the `ssh-agent`
+
+=== `rad-daemon`
+
+This command provides control over and introspection into the peer-to-peer daemon.
+
+* `start` -- request to start the peer-to-peer daemon.
+* `stop` -- request to stop the peer-to-peer daemon.
```
xla, line 254: Assuming we are deferring lifecycle control to system service managers (systemd, launchd, windows11d), is it required to expose these commands as well. It might make it harder to define the behaviour if two systems try to interfere with each other.

docs/rfc/0698-cli-infrastructure.adoc
```diff
@@ -0,0 +1,299 @@
...
+* `status` -- request the status of the peer-to-peer daemon.
```
xla, line 255: What exactly does this return? Given that the main purpose of the daemon is the operation of the `net::Peer`, it should be assumed that the info that can be requested is closer to what can be obtained through the tincan, there is an outline of that [here](https://github.com/radicle-dev/radicle-link/blob/xla/rfc-p2p-node/docs/rfc/0696-p2p-node.adoc#requestresponse-payloads).

xla (commented):
:dancers:

docs/rfc/0698-cli-infrastructure.adoc
FintanH: Nonetheless the replication _itself_ requires you need to be connected to the network. The `ask` functionality outlined below will need #141. Perhaps this section can be revised. I first wrote it when I thought the PubSub was tied to the daemon being online, which turned out to be a misunderstanding on my part :)

xla: A revision of the section would be helpful, as it seems we are converging on a model - given the prerequisites are in place - where there is no interactive protocol in place to achieve tasks like "asking"/tracking.

docs/rfc/0698-cli-infrastructure.adoc
FintanH: x-post: https://github.com/radicle-dev/radicle-link/pull/696/files#r662841432

FintanH: > Furthermore with the proposed command structure will this result in rad replicate replicate?

`rad replicate it`? :upside_down_face:

docs/rfc/0698-cli-infrastructure.adoc
FintanH: Ya, that's fair. Your RFC has helped clarify things a bit. So these two commands can be left out and `systemctl`/`launchd`/`rundosrun` will manage the life cycle.

docs/rfc/0698-cli-infrastructure.adoc
FintanH: Ya, so this should likely be split out into the separate things it can retrieve as outlined in #696, i.e. `peers`, `membership`, and `stats`.

docs/rfc/0698-cli-infrastructure.adoc
```diff
@@ -0,0 +1,295 @@
...
+
+=== `rad-replicate`
+
+* `ask` -- tell the `daemon` that you wish to replicate the
+  given URN. Pending <<rl-141, radicle-link#141>>.
+* `from` -- if the peer/address are known for the given URN,
+  attempt to replicate the identity. This requires the activation of
+  the peer-to-peer daemon.
+
+=== `rad-profile`
```
FintanH, line 232: Asking the profile's `PeerId` and `Paths` could be useful as well. Instead of navigating to your config to see your `PeerId` and remembering your `Paths` all the time :)

docs/rfc/0698-cli-infrastructure.adoc
xla: `rad yo dawg`

docs/rfc/0698-cli-infrastructure.adoc
```diff
@@ -0,0 +1,295 @@
...
+metadata for that project. As per <<arch-cli, RFC-682#CLI>>:
+
+[quote]
+-------
+
+Each subcommand MUST expose its functionality as a linkable library,
+and provide CBOR serialisation for its arguments and outputs.
+-------
+
+The output should, however, not be limited to CBOR. We will provide an
```
FintanH, line 268: Note to self: CBOR actually is mentioned, need to add the link back https://datatracker.ietf.org/doc/html/rfc8949

docs/rfc/0698-cli-infrastructure.adoc
```diff
@@ -0,0 +1,295 @@
...
+option, `--output`, that will give the caller the option to output
+into another format. The initial options will be:
+
+. `json`
+. `cbor`
+
+with the default being `cbor` due the to above requirement. More
```
kim, line 275: Hm wait, that’s the output of running the command on the terminal, right? Perhaps JSON is the better default.

docs/rfc/0698-cli-infrastructure.adoc
```diff
@@ -0,0 +1,295 @@
...
+  modular commands would mirror this structure well.
+* Subcommands can be versioned independently of each other.
+* We can dogfood our subcommand functionality
+
+Cons of *2.*:
+
+* Makes it easier to override one of the core commands.
+* Many binaries must be packaged and distributed.
+
+The author recommends *1.*. As mentioned, we have a hard requirement
```
kim, line 175: There is also a hybrid approach, exemplified by `git`: the “builtin” executables are deployed to `GIT_EXEC_PATH`, which is by default not a user-writable directory. Security-wise, this is only marginally better than `PATH`, but perhaps if this path could only be overridden by a `—exec-path` argument it would be acceptable.

I still don’t, for the set of commands described here, see a compelling reason why release builds should be split into many binaries, but I do see how this could be handy during development.

I would say the simple thing is a single binary, and user commands a la cargo.

docs/rfc/0698-cli-infrastructure.adoc
```diff
@@ -0,0 +1,295 @@
...
+around this function.
+
+== Core
+
+[quote]
+------
+We swears to serve the master of the precious. We will swear on
+the…on the precious.
+------
+-- Gollum, The Two Towers
```
kim, line 196: 👌

kim (commented):
This looks coherent to me. 

One last thing to consider is documentation — it is probably not the case that we need a new RFC for every new toggle, but it is also probably not the case that `—help` is sufficient (what to expect as output schemata, f.ex.). There are various attempts at generated docs in other languages, a bit of research into how this could be solved wouldn’t hurt.

docs/rfc/0698-cli-infrastructure.adoc
FintanH: Ya, good point. Wasn't thinking in the right context =]

docs/rfc/0698-cli-infrastructure.adoc
FintanH: Ya, I'm happy with what we landed on. Leaving the prose here for documentation of at least considering it.

docs/rfc/0698-cli-infrastructure.adoc
```diff
@@ -0,0 +1,297 @@
...
+
+Many subcommands will likely want to share a set of options. These
+options, dubbed as "global parameters" here, will be a defined set --
+exported from `radicle-link` for re-use -- that the base `rad` command
+will always receive and pass onto the subcommand being called.
+
+Other subcommands that are written in the target language (currently
+Rust) will be able to import the `struct` for reuse. If the subcommand
+is in another target language it SHALL add these parameters to its
+parameter list prefixed with `global-`. The `rad` command will parse
```
alexjg, line 92: Does this imply that global parameters are always passed to the `rad` binary before the subcommand? Presumably this also means that they are never passed by environment variable?

docs/rfc/0698-cli-infrastructure.adoc
```diff
@@ -0,0 +1,297 @@
...
+interesting event. For example, if Upstream is
+running and I create a project via the CLI, then a good user
+experience is that the Upstream is updated with the new project
+(without refreshing). In this case, the event is the creation of the
+project.
+
+In <<arch-pubsub, RFC-682#PubSub>>, a PubSub component was proposed --
+with a detailed specification being left to a future RFC. We will
+assume the existence of this component here. The CLI commands in this
+RFC MAY publish interesting events to the PubSub component. This will
```
alexjg, line 116: We're not defining the events here presumably. Are we planning on a future RFC listing all known events, or should there be a command which lists all the events a particular `rad` can produce? Put another way, if I'm a developer looking to listen to interesting events, how should I find out about them? What stability guarantees do I have?

docs/rfc/0698-cli-infrastructure.adoc
```diff
@@ -0,0 +1,297 @@
...
+=== `rad-profile`
+
+This command will provide the management of keys and the `Profile`
+construct in `librad`. The following actions will be defined:
+
+* `create`— generate a new `Profile`-- setting up a new key and any
+  related infrastructure for the profile, e.g. storage.
+* `get`— get the current `Profile`'s identifier.
+* `set`— set the current `Profile` to the identifier.
+  passed in. *Note* that this should refuse to do so if the daemon is online.
```
alexjg, line 241: As above, do we need to refuse anything if the daemon is online? Would this not just start connecting to a different peer to peer node?

docs/rfc/0698-cli-infrastructure.adoc
FintanH: The `rad` binary will always be run first and shells out to the subcommand, so ya `rad` would first parse those parameters passing them on.

> Presumably this also means that they are never passed by environment variable?

I wouldn't say so :)

docs/rfc/0698-cli-infrastructure.adoc
```diff
@@ -0,0 +1,297 @@
...
+
+To achieve the goals of having a read-only storage and using
+`<<ssh-agent>>` for signing, we propose the following should be
+implemented:
+
+. `ssh-agent` compatability with `<<radicle-keystore>>`, which has been
+started in `<<rk-17, radicle-keystore#17>>`.
+. A subcommand `rad-profile-ssh-add` for adding one's secret key to the
+`ssh-agent` using the supplied constraints, e.g. ask for password.
+. Implement `<<rl-461, radicle-link#461>>` to have a read-only storage.
```
geigerzaehler, line 142: This issue is closed :thinking: https://github.com/radicle-dev/radicle-link/issues/461

docs/rfc/0698-cli-infrastructure.adoc
```diff
@@ -0,0 +1,297 @@
...
+  graph for a given identity.
+* `rad-refs [ self | signed | delegates | delegate ]` -- retrieve the
+  blobs that live under `rad/<refname>`
+* `refs [ heads | notes | tags | category ]` -- list the reference names.
+  under the respective category, where `category` is a user supplied category.
+
+=== `rad-replicate`
+
+* `ask` -- tell the `daemon` that you wish to replicate the
+  given URN. Pending <<rl-141, radicle-link#141>>.
```
geigerzaehler, line 227: To implement a list of identities we want to replicate but haven’t found yet we would need a corresponding command. Something like `rad replicate status`.

docs/rfc/0698-cli-infrastructure.adoc
FintanH: Good questions, I think at the least the events could be documented as part of the specific CLI library. But for stability guarantees, I think it'd be worth discussing in the future PubSub RFC, what do you reckon?

docs/rfc/0698-cli-infrastructure.adoc
```diff
@@ -0,0 +1,297 @@
...
+  the peer-to-peer daemon.
+
+=== `rad-profile`
+
+This command will provide the management of keys and the `Profile`
+construct in `librad`. The following actions will be defined:
+
+* `create`— generate a new `Profile`-- setting up a new key and any
+  related infrastructure for the profile, e.g. storage.
+* `get`— get the current `Profile`'s identifier.
```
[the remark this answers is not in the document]
FintanH: Ya, so for the currently implemented `Profile` there's a file located in the config directory that's called `active_profile`. The file simply contains a `ProfileId` that points to the default to use. So if no `ProfileId` is specified, the default is used.

docs/rfc/0698-cli-infrastructure.adoc
FintanH: Ya, on second thought it might be fine. It essentially switches the value and then any action after that would be in the scope of the newly set profile, so a daemon task could access the specific socket for that profile/peer id.

docs/rfc/0698-cli-infrastructure.adoc
FintanH: It wasn't at the time of writing :) I'll remove it now.

docs/rfc/0698-cli-infrastructure.adoc
FintanH: Great point :+1:

geigerzaehler (approved):
I believe this should cover the needs of upstream

docs/rfc/0698-cli-infrastructure.adoc
```diff
@@ -0,0 +1,298 @@
...
+components MAY be added as the result of future RFCs. We SHALL keep a
+list of the core commands as part of the `radicle-link` documentation.
+
+=== `rad-identities`
+
+The `rad identities` subcommand provides the management of identity
+documents and their metadata. It is further split into subcommands
+which we sketch out below:
+
+* `project [ create | get | list | update | checkout | review ]` --
```
FintanH, line 208: Note to self: the include file management is needed as well, for managing remotes.

FintanH: Actually, I'm just outright missing `track` and `untrack` :facepalm:

FintanH: Nevermind, I _do_ have tracking which will have the side-effect of adding or removing from the include files

docs/rfc/0698-cli-infrastructure.adoc
FintanH: To me, it's much of a muchness. Either we reserve the keywords and they're always passed via the `rad` executable, or we only reserve the `global-` prefixed versions and allow other executables to use the names non-prefixed. The latter seems more courteous =]

FintanH:
I believe the RFC is in a healthy state and would like to have the last call on review from @radicle-dev/cococo =]

docs/rfc/0698-cli-infrastructure.adoc
alexjg: I think that makes sense. Does the "SHALL pass them varbatim to the subcommand" line cover the non-prefixed case?

Also: will the same processing apply to environment variables?

docs/rfc/0698-cli-infrastructure.adoc
alexjg: Yeah I think this probably makes more sense in the PubSub RFC.

docs/rfc/0698-cli-infrastructure.adoc
FintanH: > I think that makes sense. Does the "SHALL pass them varbatim to the subcommand" line cover the non-prefixed case?

I'm not sure I understand the question, could you elaborate some more :)

> Also: will the same processing apply to environment variables?

Good question, the only two env vars we have so far are `RAD_HOME` and `RAD_PROFILE`. I feel like since they're to do with an environment (read global environment), I'm more in the boat of "buyer beware" when attempting to re-use the names :sweat_smile:

alexjg: I'm reading "pass them verbatim" as passing the arguments both with and without the `global-` prefix. That's how I'm reading the sentence "allow other executables to use the names non-prefixed". Maybe I'm misunderstanding?

My personal opinion is that it's worth specifying environment variable handling because it's very common to define environment variables to match your CLI options (see for example https://docs.rs/structopt/0.3.22/structopt/#environment-variable-fallback). This is especially useful in container environments (or any 12-factor app really) where it's much easier to define arguments in environment variables in you `Dockerfile` or `deployment.yaml` or whatever other hellish config format you have.

FintanH: > I'm reading "pass them verbatim" as passing the arguments both with and without the global- prefix. That's how I'm reading the sentence "allow other executables to use the names non-prefixed". Maybe I'm misunderstanding?

So my intended meaning is that `rad` will do the following steps, using `--profile` as an example:
1. Parse `--profile abcd`
2. Create a parameter `--global-profile abcd`
3. Pass the result of 2. to `rad-my-subcommand`

But now that I'm thinking about this a bit deeper, does it even make sense? :joy: Because then you wouldn't be able to specify a second `--profile` parameter. So to make things a bit more clear and easy, we just have three global parameters prefixed by `link-`. These are reserved, and it's clear they come from `radicle-link`. 

What kind of env var handling do you think we would need to do? :thinking: Just prefix with `LINK-`? So maybe we need to propose `LINK_HOME` and `LINK_PROFILE` instead? :sweat_smile:

alexjg: One option would be to spawn the subcommand with the `RAD_HOME` and `RAD_PROFILE` environment variables set to whatever was resolved in the core based on the environment variables and command line arguments. Then we specify that subcommands MUST load those values from the environment if they want to access them at all. The remaining point to decide on is how to handle those arguments being passed to the subcommand, my preference would be to specify that global arguments are reserved, this avoids confusion for both subcommand authors and CLI users.

geigerzaehler (dismissed):

docs/rfc/0698-cli-infrastructure.adoc
kim: This thread illuminates nicely why the `git` suite is a set of executables for _every bit_ of functionality a subcommand may require. I suppose this kind of extensibility is just inherently very limited, and we should instead provide it via an embedded interpreter if and when it becomes a thing.

docs/rfc/0698-cli-infrastructure.adoc
```diff
@@ -0,0 +1,355 @@
...
+
+. `json`
+. `cbor`
+
+with the default being `json` since the CLIs are user-centric
+applications and human-readable footnote:disclaimer[For some
+definition of human-readble] output would be desirable. More output
+types MAY be proposed in future RFCs.
+
+== Documentation
```
kim, line 281: Thanks for taking the time to research this. I get that the outcome is that there is no tooling for such a thing in Rust. If it is necessary to standardise on the option parsing library (because that is part of the subcommand library surface), then I'd be in favour of `clap`.

Otherwise, it seems like documentation equals hand curated `man` pages, like in the olden days.

FintanH: There's no such tooling, but there's some groundwork at least for `clap`. I'm also in favour of it.

> Otherwise, it seems like documentation equals hand curated man pages, like in the olden days.
Also got you covered :upside_down_face: https://github.com/rust-cli/man

docs/rfc/0698-cli-infrastructure.adoc
FintanH: @alexjg I'm not sure I follow the distinction of how environment variables are usually used and what you're proposing. Maybe we could chat about it in person :)

@kim I'm just not sure what you mean :sweat_smile: maybe you could elaborate with an example from the `git` suite?

docs/rfc/0698-cli-infrastructure.adoc
kim: Hm, TIL a clap “app” can also be generated from YAML, so maybe a unison program which outputs dhall which outputs yaml could be a solution 🤔

docs/rfc/0698-cli-infrastructure.adoc
kim: What I mean is that while it is more or less trivial to just propagate some kind of stringly k/v pair to a sub process, it is not so trivial to pass an actual environment — for example the signer which follows from the choice of `RAD_PROFILE`.

In `git`, commands often exec other commands — everything referred to as “plumbing” (eg. `rev-parse`, `read-tree`, `update-ref`) . So if you implement some custom subcommand without being able to link in “plumbing” functionality, you just exec one of those. In the example, this would be `rad-sign` (which in turn delegates to an ssh-agent and) which takes some payload on stdin and outputs a signature on stdout.

The point being: what we do here is analogous to plumbing, so perhaps making this extensible via extern sub commands is not that important anyway.

That being said, I think you have almost convinced me of “global” settings being overridable by env variables because I had this flashback of how badly Haskell `stack` handles this: I recall that many options where not at all obvious as being global, but the parser wouldn’t recognise them if they came _after_ the subcommand argument. So essentially, one would always run `stack command —help` followed by `stack —help` in order to piece together what comes before `command` and what comes after.

docs/rfc/0698-cli-infrastructure.adoc
FintanH: You only get one innovation point per project :stuck_out_tongue_closed_eyes:

kim: NoveltyCoin (ℕ)

docs/rfc/0698-cli-infrastructure.adoc
FintanH: I linked up with @alexjg on a call and we discussed the potential options -- he was kind enough to lay them out and I agreed they seemed to covered the cases. We then discussed the pros and cons, finally agreeing on the desired option. I'll lay out the final decision first, then lay out the other options considered. 

The global parameters are prefixed with either `link-` or `rad-` (let the bikeshedding commence) to signify that they're owned by radicle-link. These parameter names are reserved. Each subcommand MUST accept these parameters (unless it really does not make sense to, e.g. `rad profile` wouldn't need to set a `--rad-profile` for creating one. The `rad` executable also accpets these parameters and passed them on to the subcommand, e.g. `rad --rad-profile abc issues create` becomes `rad-issues --rad-profile abc create`. We outline the following precedence rules: command-line param takes precedence over its env var equivalent, the first command-line param takes precedence over any others, i.e. `rad --rad-profile abc issues --rad-profile def create` will result in `rad-issues --rad-profile abc issues create`.

I'll now outline the options so others can review them and possibly argue for their case instead:
1. The recommended approach outlined above
2. The global parameter is parsed lexically before the subcommand and the env var is set for the subcommand process to inspect. One big con of this is that the argument does not show up in `--help` and would need to be called out specifically in that executable's documentation.
3. We allow a global parameter to be parsed anywhere in the invocation of an executable. The con is that it's a PITA to program.
4. We don't outline _any_ global parameters and subcommands are left to their own devices. One con of this approach is that for scripting it's more ergonomic to be able to do `rad --rad-profile abc` over repeating it in the various subcommands.

Let me know your thoughts on the recommendation and options, and perhaps if we missed an option that might be more ergonomic :)

kim: I approve and appreciate the effort to get this right.

> more ergonomic

I'm... I can't sit on or otherwise arrange my body around a CLI, so I'm not sure what this means :stuck_out_tongue:

FintanH: > I approve and appreciate the effort to get this right.

Thanks :blush: I'll amend this section. Since the env vars are already prefixed with `RAD_`, I'll mandate that the params will follow suit, and can be amended at another time if we decide to namespace things differently.

> I'm... I can't sit on or otherwise arrange my body around a CLI, so I'm not sure what this means

I can recommend some flexibility and mobility exercises which might help ;)

docs/rfc/0698-cli-infrastructure.adoc
```diff
@@ -86,20 +86,51 @@ options, dubbed as "global parameters" here, will be a defined set --
 exported from `radicle-link` for re-use -- that the base `rad` command
 will always receive and pass onto the subcommand being called.
 
+These parameters will be prefixed by `rad-`, signfiying that they are
+from the `radicle-link` namespace and are reserved names.
+
+Subcommands, including the <<Core>> subcommands, MUST accept these
```
alexjg, line 92: Given the exception it might be better to make this SHOULD?

FintanH: You're dead right =] Fixed, thanks!

FintanH:
FINAL CALL, ALL ABOARD :steam_locomotive:

alexjg (dismissed):

docs/rfc/0698-cli-infrastructure.adoc
```diff
@@ -0,0 +1,386 @@
...
+
+The integral part of this proposal is that there will be one binary,
+`rad`, that will be the foundation for all other subcommands. It will
+define the global parameters and how to extend it.
+
+We will discuss these in the following two sections.
+
+== The Fellowship of the `rad`
+
+[quoute]
```
kim, line 63:
```suggestion
[quote]
```

docs/rfc/0698-cli-infrastructure.adoc
```diff
@@ -0,0 +1,386 @@
...
+this RFC as commands that write to the storage generally need access
+to the secret key, whether for signing data or proof of ownership to
+the storage. For most (if not all) read-only commands, no access to
+the secret key is needed.
+
+To achieve the goals of having a read-only storage and using
+`<<ssh-agent>>` for signing, we propose the following should be
+implemented:
+
+. `ssh-agent` compatability with `<<radicle-keystore>>`, which has been
```
kim, line 169:
```suggestion
. `ssh-agent` compatibility with `<<radicle-keystore>>`, which has been
```

kim (approved):
:ship:

docs/rfc/0698-cli-infrastructure.adoc
FintanH: [quouououote]
thank you :bow:

PR #699, opened 2021-06-11
storage: introduce read-only type

### LINKED ISSUE #461: storage: bring back read-only access

FintanH:
Fixes #461

We introduce the read-only variant of Storage. The use of this variant
implies that it will be used for read-only purposes, and so it does not
hold a signer.

Storage is re-formulated to be a combination of the read-only store
along with the signer and fetchers.

### DIFF
--- a/librad/src/git/storage.rs
+++ b/librad/src/git/storage.rs
@@ -4,20 +4,12 @@
 // This file is part of radicle-link, distributed under the GPLv3 with Radicle
 // Linking Exception. For full terms see the included LICENSE file.
 
-use std::{
-    convert::TryFrom,
-    fmt::Debug,
-    marker::PhantomData,
-    path::{Path, PathBuf},
-};
+use std::{convert::TryFrom, fmt::Debug, marker::PhantomData, path::PathBuf};
 
-use git_ext::{self as ext, blob, is_not_found_err, RefLike, RefspecPattern};
-use std_ext::result::ResultExt as _;
+use git_ext::is_not_found_err;
 use thiserror::Error;
 
-use super::types::{reference, Many, One, Reference};
 use crate::{
-    identities::git::{Identities, Urn},
     paths::Paths,
     peer::PeerId,
     signer::{BoxedSigner, Signer, SomeSigner},
@@ -27,37 +19,32 @@ pub mod config;
 pub mod fetcher;
 pub mod glob;
 pub mod pool;
+pub mod read;
 pub mod watch;
 
 pub use config::Config;
 pub use fetcher::{Fetcher, Fetchers};
 pub use glob::Pattern;
 pub use pool::{Pool, PoolError, Pooled, PooledRef};
+pub use read::{Error, Storage as ReadOnly};
 pub use watch::{NamespaceEvent, Watcher};
 
 #[derive(Debug, Error)]
 #[non_exhaustive]
-pub enum Error {
-    #[error("signer key does not match the key used at initialisation")]
-    SignerKeyMismatch,
-
-    #[error("malformed URN")]
-    Ref(#[from] reference::FromUrnError),
-
+pub enum OpenError {
     #[error(transparent)]
     Config(#[from] config::Error),
 
     #[error(transparent)]
-    Blob(#[from] ext::blob::Error),
-
-    #[error(transparent)]
     Git(#[from] git2::Error),
+
+    #[error("signer key does not match the key used at initialisation")]
+    SignerKeyMismatch,
 }
 
 /// Low-level operations on the link "monorepo".
 pub struct Storage {
-    backend: git2::Repository,
-    peer_id: PeerId,
+    inner: read::Storage,
     signer: BoxedSigner,
     fetchers: Fetchers,
 }
@@ -76,7 +63,7 @@ impl Storage {
     /// the same way two `git` processes can access the same repository.
     /// However, if you need multiple [`Storage`]s to be shared between
     /// threads, use a [`Pool`] instead.
-    pub fn open<S>(paths: &Paths, signer: S) -> Result<Self, Error>
+    pub fn open<S>(paths: &Paths, signer: S) -> Result<Self, OpenError>
     where
         S: Signer + Clone,
         S::Error: std::error::Error + Send + Sync + 'static,
@@ -84,7 +71,7 @@ impl Storage {
         Self::with_fetchers(paths, signer, Default::default())
     }
 
-    pub fn with_fetchers<S>(paths: &Paths, signer: S, fetchers: Fetchers) -> Result<Self, Error>
+    pub fn with_fetchers<S>(paths: &Paths, signer: S, fetchers: Fetchers) -> Result<Self, OpenError>
     where
         S: Signer + Clone,
         S::Error: std::error::Error + Send + Sync + 'static,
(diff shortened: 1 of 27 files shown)


alexjg (approved):

librad/src/git/storage/read.rs
```diff
@@ -0,0 +1,347 @@
+// Copyright © 2019-2020 The Radicle Foundation <hello@radicle.foundation>
```
kim, line 1:
```suggestion
// Copyright © 2019-2021 The Radicle Foundation <hello@radicle.foundation>
// Copyright © 2021      The Radicle Link Contributors
```

librad/src/git/storage/read.rs
```diff
@@ -0,0 +1,347 @@
...
+    /// [`Storage`] can be sent between threads, but it can't be shared between
+    /// threads. _Some_ operations are safe to perform concurrently in much
+    /// the same way two `git` processes can access the same repository.
+    /// However, if you need multiple [`Storage`]s to be shared between
+    /// threads, use a [`Pool`] instead.
+    pub fn open(paths: &Paths) -> Result<Self, Error> {
+        crate::git::init();
+
+        let backend = match git2::Repository::open_bare(paths.git_dir()) {
+            Err(e) if is_not_found_err(&e) => {
```
kim, line 64: We can’t initialise the repo without a signer. Well, maybe we can with just a peer id? To use it rw, you need to bring the secret part.

librad/src/git/storage/read.rs
```diff
@@ -0,0 +1,347 @@
...
+    }
+
+    pub(in crate::git) fn identities<'a, T: 'a>(&'a self) -> Identities<'a, T> {
+        Identities::from(self.as_raw())
+    }
+
+    // TODO: we would need to wrap a few more low-level git operations (such as:
+    // create commit, manipulate refs, manipulate config) in order to be able to
+    // model "capabilities" in terms of traits.
+    pub(in crate::git) fn as_raw(&self) -> &git2::Repository {
```
kim, line 315: Perhaps this can be `pub(super)`, so it’s not possible to get an almighty handle without having a signer?

librad/src/git/storage/pool.rs
```diff
@@ -57,8 +57,14 @@ impl AsMut<Storage> for PooledRef {
     }
 }
 
-impl From<Object<Storage, Error>> for PooledRef {
-    fn from(obj: Object<Storage, Error>) -> Self {
```
kim, line 61: Hm wait, so you need the secret key to create a pool, but you can downgrade the borrowed object to read only.

In #693 I ran into the situation that the storage has to be initialised eagerly, so the locking gymnastics for the lazy unit could also go away entirely (if we have some witness you need to bring to the pool). Then, we could have either ro or rw pools.

librad/src/git/storage.rs
```diff
@@ -380,25 +184,16 @@ impl AsRef<Storage> for Storage {
...
-        for name in names {
-            match name {
-                Err(e) => return Some(Err(e.into())),
-                Ok(name) => match ext::RefLike::try_from(name).ok() {
-                    Some(refl) => return Some(Ok(refl)),
-                    None => continue,
-                },
-            }
-        }
+impl std::ops::Deref for Storage {
```
FintanH, line 193: So, I know we use this pattern in a few places but it's also [considered an anti-pattern](https://github.com/rust-unofficial/patterns/blob/master/anti_patterns/deref.md). Removing it would result in a good bit of boilerplate. What's our general stance on this?

kim: It's not advisable indeed, except in YOLO circumstances. As those anti-patternists say, `*x.deref()` does not have the same type as `x`, which can yield surprises when rust decides to auto-dereference.

I think in this case the only way to avoid having to borrow explicitly at the call site is to turn `Storage` into a pair of traits.

FintanH: Why do you say a pair of traits? One of them being the read-only operations like `has_*` and the reference lookups, but what would the other one consist of?

kim: Well, say `trait Storage: ReadOnlyStorage`. Should I have written “couple” to denote it’s two traits? xoxo

Maybe `struct Storage` could be an impl of `ReadOnlyStorage`, which would save one trait and may make the transition easier?

FintanH: > xoxo

:kissing_closed_eyes: 

> Well, say trait Storage: ReadOnlyStorage. Should I have written “couple” to denote it’s two traits?

What I'm wondering is what do you see going into the `Storage` trait? :) The only operations left are constructors and `as_raw`. So maybe whatever the `as_raw` is doing in the rest of the codebase (where writes are happening) should be considered going into that trait?

> Maybe struct Storage could be an impl of ReadOnlyStorage, which would save one trait and may make the transition easier?

That makes sense. So just to check we're on the same page -- instead of things taking `S: AsRef<ReadOnly>` they would take `S: ReadOnlyStorage`, right?

kim: > What I'm wondering is what do you see going into the Storage trait?

Instead of `as_raw`, we should probably strive for the respective operations be captured in either their own traits, or as methods on a rw storage. This will become more pressing in a hybrid git2/gitoxide world I suppose. It’s a lot of work for internal interfaces, which is why I didn’t do it. Perhaps we could capture `as_raw` as a trait in the meantime, because that can be made crate-visible only.

Btw, I’ll also need to take another look at whether none of the git2 types returned from the ro storage can be used mutably.

> instead of things taking S: AsRef<ReadOnly> they would take S: ReadOnlyStorage, right?
 
 Uhm, `AsRef` is actually fine, if you’re fine with calling `as_ref` everywhere. What boilerplate did you see arise without `Deref`?

FintanH: > Instead of as_raw, we should probably strive for the respective operations be captured in either their own traits

Ya, I'm with you :+1: 

> Btw, I’ll also need to take another look at whether none of the git2 types returned from the ro storage can be used mutably.

Oh that's a great point :)

> Uhm, AsRef is actually fine, if you’re fine with calling as_ref everywhere. What boilerplate did you see arise without Deref?

Actually you're right. The boilerplate was due to methods like `has_urn` being used on `Storage` directly.

I suppose one other alternative to the trait is having top-level functions, for example:
```rust
// storage/read.rs
pub fn has_urn(storage: &S, urn: &Urn) -> Result<bool, Error> 
where S: AsRef<ReadOnly>,
{
    let storage = storage.as_ref();
    /* etc */
}
```

That would mean making a lot of changes to method calls in the codebase though. Thought I'd throw it out there nonetheless.

kim: > top-level functions

Yeah, I mean, that’s the pattern used for all the higher-level effects on storage. I thought that one day I’ll add capability traits instead of passing around the concrete `Storage`, but well...

librad/src/git/storage/pool.rs
FintanH: I'm not actually quite sure what you mean here :) This initial iteration was just to get things working.

I think the simplest solution is to have a ro pool and keep this pool as the rw pool. Do you see a better way of doing it?

FintanH: So I added the witness construction but held off on modifying `net::Peer` any more than adding a function: `using_read_only`, which is just an easy way to say one is using `ReadOnly` rather than `Storage`

kim (approved):
U+1F44D

We can maybe get rid of the `Initialised` mutex, but good to do separately

PR #709, opened 2021-06-18
replication: create rad/id and rad/self symref

### LINKED ISSUE #708: Feature: replicate Person data
On a high level as a user of the protocol, I want `Person` data to be replicated whenever a project is replicated.

For example: a maintainer creates a project and that project gets replicated onto a seed node. Then a contributor follows this project and replicates it onto their machine. In this case I expect the maintainer's `Person` data to be available on the contributor's machine in the `state::list_users` list.

Required by: https://github.com/radicle-dev/radicle-upstream/pull/1912.

FintanH:
Fixes #708 

We add `rad/id` at the top level and create a symbolic ref between `rad/id` and `rad/self` for each tracked peer.

In the `daemon` code I changed the `get_user` function to return a regular `Person` document and added `get_local` for getting the local identity.

I chose to add the test to `daemon/replication` because I thought it would have fitted with the `menage` tests but they're no bueno right now.

### DIFF
--- a/daemon/src/state.rs
+++ b/daemon/src/state.rs
@@ -358,17 +358,30 @@ where
 /// Get the user found at `urn`.
 ///
 /// # Errors
-///
-///   * Resolving the user fails.
-///   * Could not successfully acquire a lock to the API.
-pub async fn get_user<S>(peer: &Peer<S>, urn: Urn) -> Result<Option<LocalIdentity>, Error>
+///   * Storage access fails
+pub async fn get_user<S>(peer: &Peer<S>, urn: Urn) -> Result<Option<Person>, Error>
 where
     S: Clone + Signer,
 {
-    peer.using_storage(move |store| match identities::person::get(store, &urn)? {
-        None => Ok(None),
-        Some(person) => local::load(store, person.urn()),
-    })
+    peer.using_storage(move |store| identities::person::get(store, &urn))
+        .await?
+        .map_err(Error::from)
+}
+
+/// Get the local user found at `urn`.
+///
+/// # Errors
+///   * Storage access fails
+pub async fn get_local<S>(peer: &Peer<S>, urn: Urn) -> Result<Option<LocalIdentity>, Error>
+where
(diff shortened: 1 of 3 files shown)


daemon/src/state.rs
```diff
@@ -355,17 +355,30 @@ where
...
+}
+
+/// Get the local user found at `urn`.
 ///
-///   * Resolving the user fails.
-///   * Could not successfully acquire a lock to the API.
-pub async fn get_user<S>(peer: &Peer<S>, urn: Urn) -> Result<Option<LocalIdentity>, Error>
+/// # Errors
+///   * Storage access fails
+pub async fn get_local<S>(peer: &Peer<S>, urn: Urn) -> Result<Option<LocalIdentity>, Error>
```
alexjg, line 372: While we're changing this stuff, is there a reason this should take a `Urn`? We could just use `local::default` to get the `LocalIdentity` associated with the store right?

kim: You can have 0..n local identities

kim (dismissed):
I _think_ that this should‘ve worked without this patch, but whatever....

librad/src/git/replication.rs
```diff
@@ -426,6 +426,40 @@ fn ensure_rad_id(storage: &Storage, urn: &Urn, tip: ext::Oid) -> Result<ext::Oid
     id_ref.oid(storage).map(Into::into).map_err(Error::Store)
 }
 
+fn adopt_rad_self(storage: &Storage, urn: &Urn, peer: PeerId) -> Result<(), Error> {
+    let rad_self = Reference::rad_self(Namespace::from(urn), peer);
+
+    // We only need to crate the rad/id there's a rad/self
```
xla, line 432:
```suggestion
    // We only need to create the rad/id there's a rad/self
```

xla (dismissed):
utACK

🗾 ♻️ 🍮 ⛳️

alexjg (approved):

PR #711, opened 2021-06-21
identities: updated delegation causes replication failure

FintanH:
We have the following steps:
* Peer 1 creates a Person document
* Peer 2 replicates this Person
* Peer 1 updates Person document
* Peer 1 creates Project delegating to new Person document
* Peer 2 attempts to replicate the Project, but fails with error below

Error:
```
Revision 7bb4bbce268d89a62e1962f5dded7a9f68c9e665 of 4b84e8407d602317b01f22207908dbea1d9d4e17 not in ancestry path of eada5419806f643d402c92b790c9b471098f1e81
```

The error originates from [`updated_person`](https://github.com/radicle-dev/radicle-link/blob/master/librad/src/identities/git.rs#L781), which seems to not consider the fact that the `Person` in the delegate is a fast-forward of the `Person` that Peer 2 knows about at the top-level. Instead, it only checks that the `Person` delegate was a previous version of the `Person` at the top-level.

Should this check be done for fast-forwards as well and do the following if it is?
```rust
self.as_person().verify(known.revision)
```

### DIFF
--- a/librad/src/git/identities/person.rs
+++ b/librad/src/git/identities/person.rs
@@ -5,7 +5,7 @@
 
 use std::{convert::TryFrom, fmt::Debug};
 
-use radicle_git_ext::{is_not_found_err, OneLevel};
+use radicle_git_ext::{self as ext, is_not_found_err, OneLevel};
 
 use super::{
     super::{
@@ -187,6 +187,21 @@ where
     S: AsRef<storage::ReadOnly>,
 {
     Ok(verified(storage).newer(a, b)?)
+}
+
+pub fn fast_forward(storage: &Storage, latest: &VerifiedPerson) -> Result<Option<ext::Oid>, Error> {
+    let urn = latest.urn().with_path(None);
+    let id_ref = super::common::IdRef::from(&urn);
+    let canonical = id_ref.oid(storage)?;
+    let tip = latest.content_id;
+    Ok(if storage.as_raw().graph_descendant_of(*tip, *canonical)? {
+        id_ref
+            .update(storage, tip, &format!("fast-forward to {}", tip))
+            .map_err(|e| Error::Store(e.into()))?;
+        Some(tip)
+    } else {
+        None
+    })
 }
 
 fn identities<S>(storage: &S) -> Identities<Person>
--- a/librad/src/git/replication.rs
+++ b/librad/src/git/replication.rs
@@ -205,8 +205,7 @@ where
                     let rad_id = unsafe_into_urn(
                         Reference::rad_id(Namespace::from(&urn)).with_remote(remote_peer),
                     );
-                    let proj = identities::project::verify(storage, &rad_id)?
-                        .ok_or(Error::MissingIdentity)?;
+                    let proj = project::verify_with_delegate(storage, &rad_id, Some(remote_peer))?;
                     let project::SetupResult {
                         updated_tips: mut project_tips,
                         identity: id_status,
@@ -262,8 +261,7 @@ where
             let (result, updated) = match identity {
                 SomeIdentity::Project(proj) => {
                     let delegate_views = project::delegate_views(storage, proj, None)?;
-                    let proj = identities::project::verify(storage, &urn)?
-                        .ok_or(Error::MissingIdentity)?;
+                    let proj = project::verify_with_delegate(storage, &urn, None)?;
                     let mut updated_delegations = project::all_delegates(&proj);
                     let rad_id = unsafe_into_urn(Reference::rad_id(Namespace::from(&urn)));
                     let project::SetupResult {
@@ -768,8 +766,8 @@ mod project {
                         let peer_id = PeerId::from(*key);
                         let (urn, project) = if &peer_id == local_peer_id {
                             let urn = proj.urn();
-                            let verified = identities::project::verify(storage, &urn)?
-                                .ok_or(Error::MissingIdentity)?;
+                            let verified =
+                                project::verify_with_delegate(storage, &urn, remote_peer)?;
                             (urn, verified)
                         } else {
                             let remote_urn = unsafe_into_urn(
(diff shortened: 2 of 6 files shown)


kim:
I think this is related to #709, which, however, doesn't actually fix the root cause. At some point, it did work that a fetch would update the top-level person-typed namespace, such that the symref can never be ahead or behind. I'm not sure why or when this broke, but it is possible that we shouldn't rely on it anyways: there is still a TODO which punts on the problem that _technically_ more top-level namespaces need to be updated, because the tracking graph may have changed.

A possible fix  could thus be that #709 resets the top-level `rad/id` to the local `remotes/../rad/self` iff the latter can be fast-forwarded onto the former. I also think that the fetch turns any symrefs into direct refs if they get updated as a result of the fetch, so the symrefs need to be forced always.

kim:
> I think in this case because the rad/id of the Person is replicated before the
> Project it doesn't get a chance to create the symref, because it will fail the
> verification with the above error before it gets to the symref creation point.

Ah yes... peer 2 already has the top-level namespace, that's why. We would need
to do another roundtrip to first fetch everything in `rad/ids` in order to make
sure that the top-level is up-to-date.

That's annoying.

I think what I proposed could still be viable -- after the `Peek`, the top-level
can be reset to the corresponding `rad/ids/XXX` if and only if that'd be a
fast-forward.

This may render the identity invalid, however, so we would need to verify it
before replicating it again (and not replicate if it became invalid). In this
case, the project would also fail to verify -- which has dubious semantics.

So I'm not sure. Maybe the correct thing to do is to only verify the project
against the state of `rad/ids`. If both `rad/namespaces/XXX/refs/rad/id` and
`refs/rad/ids/XXX` verify and are reachable from each other, the former can be
reset to the more recent one, and the latter turned into a symref. If the latter
verifies, but the former doesn't, then the project is in some kind of limbo
state.

librad/src/identities/git.rs
```diff
@@ -775,11 +775,29 @@ impl<'a> Identities<'a, Project> {
...
+            let commit = self.repo.find_commit(head)?;
+            self.is_in_ancestry_path(known.content_id.into(), commit.tree_id())
+        };
         // Nb. technically we could coerce `known` into a `VerifiedPerson` if its
         // `content_id` equals `latest_head`. Let's not introduce an unsafe
         // coercion, but rely on caching to be implemented efficiently.
         if self.is_in_ancestry_path(latest_head, known.revision.into())? {
             self.as_person().verify(latest_head)
+        // If the latest is a fast-forward then we can update the top-level
+        } else if is_fast_forward(&known, latest_head)? {
```
FintanH, line 788: I'm not sure about the placement of this, but I was struggling to find where the logic could go. What do you think @kim?

kim: Into `replication`. 

I said [here](https://github.com/radicle-dev/radicle-link/pull/711#issuecomment-866714328) that it is probably a mistake to fail verification if _top level_ delegates fail verification. The reason is this:

Imagine you create that project with valid delegates, which are then "checkpointed" as `rad/ids/*`. Now, one of them creates an invalid revision _after_ this checkpoint. If the project fails verification by auto-resolving to the top-level, it can _never_ recover from that (eg. by removing the faulty delegate), because we never replicate it again.

To solve this, we'd need to allow passing the `find_latest_head` function to `git::identities::project::verify`, and make it resolve to `rad/ids/<urn.id>` on the first pass. After replication is done doing its thing, we run verification again, this time resolving to `refs/namespaces/<urn.id>/rad/id`, and return a _different_ value from `replicate` if that fails.

test/src/test/integration/librad/scenario/updated_delegate.rs
```diff
@@ -0,0 +1,106 @@
+// Copyright © 2019-2020 The Radicle Foundation <hello@radicle.foundation>
```
xla, line 1: Is there not an updated license header?

xla (approved):
Respect your elders.

🔒 🛫 🏃 🏙

test/src/test/integration/librad/scenario/updated_delegate.rs
FintanH: Ah yes, thank you!

librad/src/git/identities/person.rs
```diff
@@ -189,6 +189,21 @@ where
     Ok(verified(storage).newer(a, b)?)
 }
 
+pub fn fast_forward<S>(
```
kim, line 192: I get how you're guided by the types, but you end up with a lot of IO:

* see if `known` exists
* load and verify `known`
* do ancestry check between `known` and `latest`
* update ref

There is also a correctness problem: the reason for introducing this is to reset `refs/namespaces/XXX/rad/id` to `refs/namespaces/YYY/rad/ids/XXX` _iff_ the latter can be fast-forwarded onto the former. _However_, this **must** be a fast-forward in the `git` sense, not the `tree` history sense, otherwise `YYY` can rewrite the history of `XXX`!

(No, we cannot actually prevent this entirely without a blockchain, but it is surely _safer_ to disallow history rewrites on identity branches wherever possible)

So I think this can be simplified to a function which takes a `VerifiedPerson` and:

1. Looks up the oid of the `rad/id`(ie. `refname_to_id(RefLike::try_from(pers.urn().with_path(None))?)`)
2. Calls `graph_descendant_of(pers.content_id, canonical_oid)`
3. If true, reset the ref from 1. to `pers.content_id`

Nb.: `VerifiedPerson` is already verified (qed), so if 1. and 2. pass we have also proven validity of the top-level person.

librad/src/git/replication.rs
```diff
@@ -205,8 +205,13 @@ where
                     let rad_id = unsafe_into_urn(
                         Reference::rad_id(Namespace::from(&urn)).with_remote(remote_peer),
                     );
-                    let proj = identities::project::verify(storage, &rad_id)?
-                        .ok_or(Error::MissingIdentity)?;
+                    let proj = identities::project::verify_with(storage, &rad_id, |delegate| {
+                        let refname =
```
kim, line 209: It is a bit hard to see from the code how the lookup functions are different for the various invocations. Do you think this could be extracted out somehow (maybe curried function "constructors" or something)?

librad/src/git/identities/person.rs
FintanH: Thanks for the help, that makes more sense. Hopefully, the simplified version is right :crossed_fingers:

FintanH:
Is there anything blocking this @kim? Or is it a matter of you working on replication v3?

kim (approved):

PR #712, opened 2021-06-22
It ain't over till it's over

kim:
Revise blocking tasks once more.

Also add signal handlers to `ephemeral-peer` and `seed`. `daemon` could also use
them, but it currently spawns the protocol stack nowhere near a `main` function.


FintanH (approved):
Thanks for the clear docs :ok_hand:

PR #714, opened 2021-06-23
test: use `blocking` and single-threaded runtime

kim:
Due to the use of `blocking` in `executor`, the multi-threaded tokio
runtime is no longer required. Gives a very slight speedup for test.

Also use `blocking` for blocking tasks in tests.


xla (approved):
✒️ ☔️ ♑️ 🏙

FintanH:
Were the initial test failures flakes, or what's the story there?

kim:
I'm not sure ... those tests are timing dependent, and occasionally failed before. I think this "wait for tin can message, or my watch shows 6pm" pattern should maybe be revisited...

FintanH (approved):

PR #717, opened 2021-06-29
profile: refactor Profile

FintanH:
While scoping out the introduction of creating, getting, and setting
profiles, I saw some room for refactoring to better see the flow of
profile creation and loading.

This patch introduces the `profile::id` module which contains the
`ProfileId` type. It surfaces the many ways to construct a ProfileId:
* From a UUID
* Read from RAD_PROFILE
* Read from active_profile
* Read from active_profile or default to creating one

Before this change, a profile ID was simply a String, but now that it is
behind a newtype wrapper, it's more obvious that it must go through
validation during construction.

The other introduction to this change is the RadHome enum which captures
where the Profile is living: XDG or some given Root.

The construction of a Profile is now computed in terms of RadHome and
ProfileId.

### DIFF
--- a/librad/src/profile.rs
+++ b/librad/src/profile.rs
@@ -5,21 +5,22 @@
 
 use std::{
     env,
-    fs,
     io,
-    path,
     path::{Path, PathBuf},
 };
-use uuid::Uuid;
 
 use crate::paths::{project_dirs, Paths};
 
+pub mod id;
+pub use id::ProfileId;
+
+const RAD_HOME: &str = "RAD_HOME";
+const RAD_PROFILE: &str = "RAD_PROFILE";
+
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
-    #[error("invalid profile ID in RAD_PROFILE environment variable: {id}")]
-    InvalidProfileIdFromEnv { id: String },
-    #[error("invalid profile ID loaded from {path}: {id}")]
-    InvalidProfileIdFromFile { id: String, path: PathBuf },
+    #[error(transparent)]
+    ProfileId(#[from] id::Error),
     #[error(transparent)]
     Io(#[from] io::Error),
 }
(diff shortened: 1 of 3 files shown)


librad/src/profile.rs
```diff
@@ -29,10 +30,37 @@ pub enum Error {
...
+pub enum RadHome {
+    /// The system's XDG family of directories.
+    XDG,
+    /// A given path, usually set through `RAD_HOME`.
+    Root(PathBuf),
+}
+
+impl RadHome {
+    /// If `RAD_HOME` is defined then the path supplied there is used and
+    /// `RadHome::Rad` is constructed. Otherwise, `RadHome::XDG` is constructed.
```
xla, line 47:
```suggestion
    /// `RadHome::Root` is constructed. Otherwise, `RadHome::XDG` is constructed.
```

librad/src/profile.rs
```diff
@@ -91,83 +101,31 @@ impl Profile {
...
-        let paths = Paths::from_root(root.join(&id))?;
+        let paths = match home {
+            RadHome::XDG => Paths::new(&id.0)?,
+            RadHome::Root(root) => Paths::from_root(root.join(&id.0))?,
+        };
 
         Ok(Self { id, paths })
     }
 
+    pub fn from_root(root: &Path, profile_id: Option<ProfileId>) -> Result<Self, Error> {
```
xla, line 118: Is this method strictly necessary? The `from_home` method seems to be sufficient for both variants of the `RadHome` enum, which both are constructible by callers.

librad/src/profile/id.rs
```diff
@@ -0,0 +1,178 @@
+// Copyright © 2021 The Radicle Link Contributors
```
xla, line 1: :black_heart:

librad/src/profile.rs
FintanH: It's there to be backward-compatible and a lot of the testing of this uses `from_root` so it seems handy to have.

kim (approved):
This looks good to me as a refactor.

The only thing I'm not 100% convinced of is the naming `XDG` -- as a matter of fact, we are only adhering to XDG on Linux. System directory conventions are subtly not compatible, so suggesting that it's XDG everywhere is potentially confusing. See [here](https://stackoverflow.com/questions/43853548/xdg-basedir-directories-for-windows/49783629#49783629) for some reflections by the author of the `directories` crate.

While we're at it, I'm never opposed to opening a can of worms: was it a mistake maybe to not just use XDG on _all_ systems? I'm happily using XDG on macOS for a lot of things, but I get that hell may break loose on Windows. It may also be that `Upstream` would end up looking in XDG for link data, but `~/Library/Application Support` for chrome state, overall complicating the matter. Discuss!

FintanH:
> The only thing I'm not 100% convinced of is the naming XDG -- as a matter of fact, we are only adhering to XDG on Linux.

Good point, I think `ProjectDirs` is a better name because that's essentially what it's using in the end anyway.

> See here for some reflections by the author of the directories crate.

Thanks for the link, that was enlightening :)

> was it a mistake maybe to not just use XDG on all systems?

Are we not? :thinking: I was under the impression that using [`ProjectDirs`](https://docs.rs/directories/3.0.2/directories/struct.ProjectDirs.html) does this, or maybe I'm misunderstanding your point?

xla:
> Discuss!

Adhering to the platforms prescribed default layouts has advantages like least surprise effect. It also should be to aid with packaging. Where it falls apart is when there is divergence as we experienced in the past with `BIN_DIR` and likely will encounter with the `RUNTIME_DIR`, both only covered by the XDG spec.

> Are we not? thinking I was under the impression that using [`ProjectDirs`](https://docs.rs/directories/3.0.2/directories/struct.ProjectDirs.html) does this, or maybe I'm misunderstanding your point?

It uses `XDG` on linux, `Knwon Folders` on Windows and `Standard Directories` on macOS.

kim:
> Knwon Folders

PR #720, opened 2021-06-30
profile: management of profiles

FintanH:
Adding functionality for management of `Profile`s. Specifically the following functions:
* `new`
* `get`
* `set`
* `list`

TODO:
- [x] Add documentation to functions
- [x] Test unhappy path of profile id not existing

### DIFF
--- a/librad/src/profile.rs
+++ b/librad/src/profile.rs
@@ -19,6 +19,8 @@ const RAD_PROFILE: &str = "RAD_PROFILE";
 
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
+    #[error("the profile {0} does not exist")]
+    DoesNotExist(ProfileId),
     #[error(transparent)]
     ProfileId(#[from] id::Error),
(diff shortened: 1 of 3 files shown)


librad/src/profile.rs
```diff
@@ -69,6 +71,49 @@ impl RadHome {
...
+            return Err(Error::DoesNotExist(id));
+        }
+        id.set_active(home)?;
+        Self::from_home(home, Some(id))
+    }
+
+    /// List all the `Profile`s that can be found under `home`.
+    ///
+    /// Note: It is expected that only [`ProfileId`]s exist under `home`.
+    pub fn list(home: &RadHome) -> Result<Vec<Self>, Error> {
```
FintanH, line 105: `BTreeSet` would be more truthful, since they'll be unique.

FintanH: Nevermind, `Self` is not `Ord`

kim (dismissed):
lgtm

PR #721, opened 2021-07-01
test: replication of user

FintanH:
The `track` and `fetch` calls in the `replication_includes_user` seem to
race on git resources causing this test to be flaky. Awaiting the
fetched events after the track seems to fix this.

### DIFF
--- a/test/src/test/integration/daemon/replication.rs
+++ b/test/src/test/integration/daemon/replication.rs
@@ -472,7 +472,7 @@ fn replication_includes_user() -> Result<(), anyhow::Error> {
     logging::init();
 
     let mut harness = Harness::new();
-    let alice = harness.add_peer("alice", RunConfig::default(), &[])?;
+    let mut alice = harness.add_peer("alice", RunConfig::default(), &[])?;
     let bob = harness.add_peer(
         "bob",
         RunConfig::default(),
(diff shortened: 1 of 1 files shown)


test/src/test/integration/daemon/replication.rs
```diff
@@ -499,6 +499,12 @@ fn replication_includes_user() -> Result<(), anyhow::Error> {
         .await?;
 
         state::track(&alice.peer, project.urn(), bob.peer_id).await?;
+
+        // FIXME: we have to wait for these events otherwise track and fetch race. A
+        // similar double call is required in the track_peer test.
+        assert_fetched(&mut alice.events).await?;
+        assert_fetched(&mut alice.events).await?;
```
xla, line 506: Why is it two events that are awaited here?

xla (commented):
Is there an intuition what the resources are raced on? Does that mean if this is held wrong by a caller it might lead to the same behaviour, or is it just that the assertions in the test fail?

test/src/test/integration/daemon/replication.rs
FintanH: The same thing happens here https://github.com/radicle-dev/radicle-link/blob/master/test/src/test/integration/daemon/replication.rs#L454-L457. Now "why", I'm not too sure :grimacing:

xla (approved):
⚰ ✈️ ⌨ 🛬

FintanH:
> Is there an intuition what the resources are raced on? Does that mean if this is held wrong by a caller it might lead to the same behaviour, or is it just that the assertions in the test fail?

Thinking about this, and I wonder if it's the `query` that occurs in the `track` followed by the `fetch`. This will cause two fetches to occur and one of them will bail because there's a mutex on the URN. I wonder if the test could be simplified to doing either a just `track` or `fetch`.

xla:
> Thinking about this, and I wonder if it's the `query` that occurs in the `track` followed by the `fetch`. This will cause two fetches to occur and one of them will bail because there's a mutex on the URN. I wonder if the test could be simplified to doing either a just `track` or `fetch`.

These littered queries and announces really weren't a good idea in the first place, what have we done. :D

xla:
Still relevant?

FintanH:
> Still relevant?

Buh, I think the test still occassionally fails but so do a lot of tests :laughing: I'll just close for now

PR #724, opened 2021-07-02
git-protocol: git protocol v2 machinery

kim:
Introduces a new crate with the lower-level machinery for handling
replication based on git protocol v2.

### DIFF
--- a/git-protocol/src/fetch.rs
+++ b/git-protocol/src/fetch.rs
@@ -1,0 +1,396 @@
+// Copyright © 2021 The Radicle Link Contributors
+//
+// This file is part of radicle-link, distributed under the GPLv3 with Radicle
+// Linking Exception. For full terms see the included LICENSE file.
+
+use std::{
+    future::Future,
+    io,
+    mem,
+    pin::Pin,
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
+    task::{Context, Poll},
+};
+
+use bstr::{BString, ByteSlice as _};
+use futures_lite::{
+    future,
+    io::{AsyncBufRead, AsyncRead, AsyncWrite},
+};
+use git_features::progress::{self, Progress};
+use git_protocol::fetch::{
+    response,
+    Action,
+    Arguments,
+    Delegate,
+    DelegateBlocking,
+    LsRefsAction,
+    Response,
+};
+use git_transport::client;
+use pin_project::{pin_project, pinned_drop};
+use versions::Version;
+
+pub use git_hash::ObjectId;
+pub use git_protocol::fetch::Ref;
+
+use crate::packwriter::PackWriter;
+
+/// A known reference to fetch.
+///
+/// Can be used to bypass `ls-refs` entirely via the `ref-in-want` feature, if
+/// all wanted refs are known a priori.
+#[derive(Debug)]
+pub struct WantRef {
+    /// Name of the ref.
+    ///
+    /// Must be in canonical form, eg. `refs/heads/pu`, not `pu`.
+    pub name: BString,
+    /// The local tip of the ref, if available.
+    ///
+    /// If the ref is already present locally, the remote end will be told that
+    /// we `have` the `oid` already, so as to allow it to assemble a smaller
+    /// packfile.
+    pub have: Option<ObjectId>,
+}
+
+/// A reference to fetch.
+///
+/// This is the result of the `filter` function passed to [`Fetch`]: the refs
+/// advertised by the remote (if and when `ls-refs` is executed) are passed
+/// through the `filter` for packfile negotiation.
+#[derive(Debug)]
+pub struct WantHave {
+    /// Tip of the ref the server said it has, and which we want.
+    pub want: ObjectId,
+    /// Local tip of the ref, if we have it.
+    pub have: Option<ObjectId>,
+}
+
+#[derive(Debug)]
+pub struct Options {
+    /// The remote (logical) repository to fetch from.
+    ///
+    /// Normally, this is the path to a repo on the remote side (eg.
+    /// `/git.git`). `radicle-link` serves only a single namespaced repo, so
+    /// this value should be the name of a namespace.
+    pub repo: BString,
+
+    /// [Extra Parameters][extra] to send with the initial transport header.
+    ///
+    /// [extra]: https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/technical/pack-protocol.txt#n52
+    pub extra_params: Vec<(String, Option<String>)>,
+
+    /// Prefixes of refs to ask the server to advertise via `ls-refs`.
+    ///
+    /// If this is `None`, the `ls-refs` command will not be sent. If it is
+    /// `Some`, but contains an empty [`Vec`], the server is asked to return
+    /// all refs it knows about. Otherwise, the server is asked to only
+    /// return refs matching the given prefixes.
+    pub ref_prefixes: Option<Vec<BString>>,
+
+    /// Known refs to ask the server to include in the packfile.
+    ///
+    /// If both `want_refs` and `ref_prefixes` are empty, it is unlikely that
+    /// the server will send a packfile at all.
+    pub want_refs: Vec<WantRef>,
+
+    /// Whether to indicate to the server that we're done after having received
+    /// a packfile.
+    ///
+    /// If `true`, send a FLUSH packet after the packfile was received. This
+    /// will cause the remote end to close the stream, so it cannot be
+    /// reused. `false` does not send a FLUSH packet, allowing to send
+    /// further commands over the same stream.
+    pub done_after_pack: bool,
+}
+
+/// Result of a succesful [`fetch`].
+#[derive(Debug)]
+pub struct Outputs<T> {
+    /// The refs contained in the packfile (if not empty).
+    ///
+    /// These are the refs advertised in response to `ls-refs`, filtered through
+    /// the `filter_refs` function of [`Fetch`], and any `wanted-ref`s in
+    /// response to `want-ref`.
+    pub refs: Vec<Ref>,
+    /// If a packfile was received successfully, some info about it.
+    pub pack: Option<T>,
+}
+
+impl<T> Default for Outputs<T> {
+    fn default() -> Self {
+        Self {
+            refs: Vec::default(),
+            pack: None,
+        }
+    }
+}
+
+/// [`Delegate`] driving the fetch end of the [pack protocol].
+///
+/// [pack protocol]: https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/technical/pack-protocol.txt
+pub struct Fetch<F, P, O> {
+    opt: Options,
+    filter_refs: F,
+    pack_writer: P,
+    out: Outputs<O>,
+}
+
+impl<F, P, O> Fetch<F, P, O> {
+    pub fn new(opt: Options, filter_refs: F, pack_writer: P) -> Self {
+        Self {
+            opt,
+            filter_refs,
+            pack_writer,
+            out: Outputs::default(),
+        }
+    }
+
+    pub fn with_filter_refs<G>(self, filter_refs: G) -> Fetch<G, P, O> {
+        Fetch {
+            opt: self.opt,
+            filter_refs,
+            pack_writer: self.pack_writer,
+            out: self.out,
+        }
+    }
+
+    pub fn outputs(&self) -> &Outputs<O> {
+        &self.out
+    }
+
+    pub fn take_outputs(&mut self) -> Outputs<O> {
+        mem::take(&mut self.out)
+    }
+}
+
+impl<F, P> DelegateBlocking for Fetch<F, P, P::Output>
+where
+    F: Fn(&Ref) -> io::Result<Option<WantHave>>,
+    P: PackWriter,
+{
+    fn handshake_extra_parameters(&self) -> Vec<(String, Option<String>)> {
+        self.opt.extra_params.clone()
+    }
+
+    fn prepare_ls_refs(
+        &mut self,
+        caps: &client::Capabilities,
+        args: &mut Vec<BString>,
+        _: &mut Vec<(&str, Option<&str>)>,
+    ) -> io::Result<LsRefsAction> {
+        let act = match &self.opt.ref_prefixes {
+            None => LsRefsAction::Skip,
+            Some(prefixes) => {
+                // Work around `git-upload-pack` not handling namespaces properly
+                // before git/2.31
+                //
+                // cf. https://lore.kernel.org/git/pMV5dJabxOBTD8kJBaPuWK0aS6OJhRQ7YFGwfhPCeSJEbPDrIFBza36nXBCgUCeUJWGmpjPI1rlOGvZJEh71Ruz4SqljndUwOCoBUDRHRDU=@eagain.st/
+                let must_namespace = remote_git_version(caps)
+                    .map(|version| version < Version::new("2.31.0").unwrap())
+                    .unwrap_or(false);
+                for prefix in prefixes {
+                    let prefix = if must_namespace {
+                        format!("ref-prefix refs/namespaces/{}/{}", self.opt.repo, prefix)
+                    } else {
+                        format!("ref-prefix {}", prefix)
+                    };
+                    args.push(prefix.into())
+                }
+                LsRefsAction::Continue
+            },
+        };
+
+        Ok(act)
+    }
+
+    fn prepare_fetch(
+        &mut self,
+        _: git_transport::Protocol,
+        caps: &client::Capabilities,
+        _: &mut Vec<(&str, Option<&str>)>,
+        refs: &[Ref],
+    ) -> io::Result<Action> {
+        if !self.opt.want_refs.is_empty() && !remote_supports_ref_in_want(caps) {
+            return Err(io::Error::new(
+                io::ErrorKind::Unsupported,
+                "`want-ref`s given, but server does not support `ref-in-want`",
+            ));
+        }
+
+        // If we don't send any `have`s, and no `want-ref`s, the server will
+        // just ignore the fetch command and wait for the next. This would cause
+        // `git_protocol::fetch` to hang indefinitely, so abort here.
+        let act = if refs.is_empty() && self.opt.want_refs.is_empty() {
+            Action::Cancel
+        } else {
+            Action::Continue
+        };
+
+        Ok(act)
+    }
+
+    fn negotiate(
+        &mut self,
+        refs: &[Ref],
+        args: &mut Arguments,
+        _: Option<&Response>,
+    ) -> io::Result<Action> {
+        for r in refs {
+            if let Some(WantHave { want, have }) = (self.filter_refs)(r)? {
+                self.out.refs.push(r.to_owned());
+                args.want(want);
+                if let Some(oid) = have {
+                    args.have(oid);
+                }
+            }
+        }
+
+        for WantRef { name, have } in &self.opt.want_refs {
+            // Work around `git-upload-pack` not handling namespaces properly,
+            // cf. https://lore.kernel.org/git/CD2XNXHACAXS.13J6JTWZPO1JA@schmidt/
+            let want_ref = format!("refs/namespaces/{}/{}", self.opt.repo, name);
+            args.want_ref(BString::from(want_ref).as_bstr());
+            if let Some(oid) = have {
+                args.have(oid);
+            }
+        }
+
+        // send done, as we don't bother with further negotiation
+        Ok(Action::Cancel)
+    }
+
+    fn indicate_client_done_when_fetch_completes(&self) -> bool {
+        self.opt.done_after_pack
+    }
+}
+
+#[async_trait(?Send)]
+impl<F, P> Delegate for Fetch<F, P, P::Output>
+where
+    F: Fn(&Ref) -> io::Result<Option<WantHave>>,
+    P: PackWriter,
+{
+    async fn receive_pack(
+        &mut self,
+        pack: impl AsyncBufRead + Unpin + 'async_trait,
+        prog: impl Progress,
+        _: &[Ref],
+        resp: &Response,
+    ) -> io::Result<()> {
+        // Strip any namespaces leaked by the other end due to workarounds
+        let namespace = format!("refs/namespaces/{}/", self.opt.repo);
+        self.out.refs.extend(
+            resp.wanted_refs()
+                .iter()
+                .map(|response::WantedRef { id, path }| Ref::Direct {
+                    path: path
+                        .strip_prefix(namespace.as_bytes())
+                        .map(BString::from)
+                        .unwrap_or_else(|| path.clone()),
+                    object: *id,
+                }),
+        );
+        let out = self.pack_writer.write_pack(pack, prog)?;
+        self.out.pack = Some(out);
+
+        Ok(())
+    }
+}
+
+/// Future created by the [`fetch`] function.
+///
+/// Ensures that a running inner [`PackWriter`] is cancelled when the
+/// [`Fetching`] future is dropped without also dropping the [`AsyncRead`] data
+/// source.
+#[pin_project(PinnedDrop)]
+struct Fetching<T> {
+    stop: Arc<AtomicBool>,
+    #[pin]
+    task: T,
+}
+
+#[pinned_drop]
+impl<T> PinnedDrop for Fetching<T> {
+    fn drop(self: Pin<&mut Self>) {
+        self.stop.store(true, Ordering::Release)
+    }
+}
+
+impl<T> Future for Fetching<T>
+where
+    T: Future,
+{
+    type Output = T::Output;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
+        self.project().task.poll(cx)
+    }
+}
+
+pub fn fetch<F, B, P, R, W>(
+    opt: Options,
+    filter_refs: F,
+    build_pack_writer: B,
+    recv: R,
+    send: W,
+) -> impl Future<Output = io::Result<Outputs<P::Output>>>
+where
+    F: Fn(&Ref) -> io::Result<Option<WantHave>> + Send + 'static,
+    B: FnOnce(Arc<AtomicBool>) -> P,
+    P: PackWriter + Send + 'static,
+    P::Output: Send + 'static,
+    R: AsyncRead + Unpin + Send + 'static,
+    W: AsyncWrite + Unpin + Send + 'static,
+{
+    use git_transport::{
+        client::git::{ConnectMode, Connection},
+        Protocol,
+    };
+
+    let stop = Arc::new(AtomicBool::new(false));
+    let task = blocking::unblock({
+        let mut conn = Connection::new(
+            recv,
+            send,
+            Protocol::V2,
+            opt.repo.clone(),
+            None::<(String, Option<u16>)>,
+            ConnectMode::Daemon,
+        );
+        let pack_writer = build_pack_writer(Arc::clone(&stop));
+
+        move || {
+            let mut delegate = Fetch::new(opt, filter_refs, pack_writer);
+            future::block_on(git_protocol::fetch(
+                &mut conn,
+                &mut delegate,
+                |_| unreachable!("credentials helper requested"),
+                progress::Discard,
+            ))
+            .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
+
+            Ok(delegate.out)
+        }
+    });
+
+    Fetching { stop, task }
+}
+
+fn remote_git_version(caps: &client::Capabilities) -> Option<Version> {
+    let agent = caps.capability("agent").and_then(|cap| {
+        cap.value()
+            .and_then(|bs| bs.to_str().map(|s| s.to_owned()).ok())
+    })?;
+    Version::new(agent.strip_prefix("git/")?)
+}
+
+fn remote_supports_ref_in_want(caps: &client::Capabilities) -> bool {
+    caps.capability("fetch")
+        .and_then(|cap| cap.supports("ref-in-want"))
+        .unwrap_or(false)
+}
--- a/git-protocol/src/packwriter.rs
+++ b/git-protocol/src/packwriter.rs
@@ -1,0 +1,214 @@
+// Copyright © 2021 The Radicle Link Contributors
+//
+// This file is part of radicle-link, distributed under the GPLv3 with Radicle
+// Linking Exception. For full terms see the included LICENSE file.
+
+use std::{
+    io,
+    path::{Path, PathBuf},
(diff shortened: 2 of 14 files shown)


kim:
@Byron would you mind taking a look?

It seems to work basically, but then hangs forever reading the packfile. By replacing `copy` with a loop, I can see that it reads exactly 95 bytes in three `read` calls, and then waits indefinitely ... for the flush pkt?

kim:
@Byron nevermind -- somewhat surprisingly, I need to spawn `upload-pack` in stateful mode, otherwise it doesn't send the capabilities advertisement. Now, I need to close its `stdin` after the packfile is sent to make the process terminate... :thinking:

git-protocol/src/fetch.rs
```diff
@@ -0,0 +1,493 @@
...
+        recv,
+        send,
+        Protocol::V2,
+        opt.repo.clone(),
+        None::<(String, Option<u16>)>,
+        ConnectMode::Daemon,
+    );
+    blocking::unblock(move || {
+        let mut delegate = Fetch::new(opt, filter_refs, receive_pack);
+        future::block_on(git_protocol::fetch::fetch(
```
alexjg, line 418: Is this `block_on` inside `unblock` because the delegate can have blocking calls as per the documentation on `git_protocol::fetch::fetch`?

kim:
@Byron nevermind the nevermind -- it's pretty bad, though, as the client can keep `upload-pack` running by simply not closing the connection. I wonder if I should just send made-up capabilities, so I can use `--stateless-rpc`.

git-protocol/src/fetch.rs
kim: I heard you like to block so I put some block in your unblock so you can block non-blocking!

kim: No seriously, this is basically a nested executor, such that blocking code (namely the machinery which writes the packfile to disk) can incrementally pull data from an `AsyncBufRead`.

In german, we call this "von hinten durch die Brust ins Auge".

alexjg: Ha, amazing idiom. Apparently the closest idiomatic English translation is to "do something ass backwards", which I don't really feel captures the implied complexity of the literal translation.

So this is because `Bundle::write_to_directory` requires `BufRead` and so we have to use `BlockOn` for that? I guess until we get async filesystem IO that's going to be an issue.

FintanH: I also saw:
> Going around your ass to get to your elbow

Which is appreciate a lot

Byron: > So this is because Bundle::write_to_directory requires BufRead and so we have to use BlockOn for that? I guess until we get async filesystem IO that's going to be an issue.

Indeed, it's expected that this blocks. With `blocking` it shouldn't be an issue to run it within async code without blocking the executor. The only real question then is how the operation can be cancelled. Generally this is supposed to be done with these [little interrupt atomics](https://github.com/Byron/gitoxide/blob/38aca4076037a6f8288c2cf483f134ea16c328d5/gitoxide-core/src/pack/receive.rs#L201) that able to make the input reader fail. This could also be achieved by wrapping the input reader into a custom implementation, but `gitoxide` provides support to interrupt all potentially long running operations and may implement iterators on top of that. The latter play nicely with `blocking` as it will be able to stop ~~polling~~ calling `next()` on the iterator if the receiver of the computation is dropped.

Byron:
> @Byron nevermind the nevermind -- it's pretty bad, though, as the client can keep upload-pack running by simply not closing the connection. I wonder if I should just send made-up capabilities, so I can use --stateless-rpc.

Sorry for the hassle!

I will take a closer look tomorrow. It could be that this setup is different enough from the tests (which are against the daemon and against github) to cause something to be a little off. Failing to flush a `flush` packet line could certainly be a cause of this. The daemon might theoretically run into the same issue but somehow cleans out `upload-pack` when the connection is dropped, and I never noticed.

kim:
It does work as designed, except that the `git-upload-pack` executable doesn’t allow to turn on capability advertisement without _also_ turning on refs advertisement: https://github.com/git/git/blob/master/builtin/upload-pack.c#L53

git-protocol/src/upload_pack.rs
```diff
@@ -0,0 +1,134 @@
...
+            .map_err(invalid_data)?
+            .map_err(invalid_data)?;
+        match pkt {
+            PacketLine::Data(data) => std::str::from_utf8(data)
+                .map_err(invalid_data)?
+                .parse()
+                .map_err(invalid_data),
+            _ => Err(invalid_data("not a header packet")),
+        }
+    }?;
```
Byron, line 85: Nitpick: I was actually getting stuck on this `?` and realized that it helped my understanding to put it one line up, as close to its effective scope as possible.

git-protocol/src/upload_pack.rs
```diff
@@ -0,0 +1,134 @@
...
+            .stdin(Stdio::piped())
+            .stderr(Stdio::inherit())
+            .kill_on_drop(true)
+            .reap_on_drop(true)
+            .spawn()?;
+
+        let mut stdin = child.stdin.take().unwrap();
+        let mut stdout = child.stdout.take().unwrap();
+
+        try_join!(
```
Byron, line 118: If this would be done with [Stream::collect](https://docs.rs/futures-lite/1.12.0/futures_lite/stream/trait.StreamExt.html#method.try_collect) the `futures-util` dependency could be removed.

It _should_ be the same.

Byron (commented):
For now I took a look at all code except for the tests which are merely postponed to another session.

All looks good to me and it's nice to see that  `git_protocol::client::fetch()`  does indeed compose.

I can't say anything about what to do with `git-upload-pack`, you seem to know what works and what doesn't, but if you think that `gitoxide` isn't speaking `git` correctly confusing the server in the process then it's definitely something to reproduce and fix.

git-protocol/src/fetch.rs
```diff
@@ -0,0 +1,493 @@
...
+        prog: impl Progress,
+    ) -> io::Result<Self::Output> {
+        use git_pack::{bundle::write::Options, data::input::Mode, index::Version, Bundle};
+
+        let opts = Options {
+            thread_limit: self.opt.max_indexer_threads,
+            index_kind: Version::V2,
+            iteration_mode: Mode::Verify,
+        };
+        Bundle::write_to_directory(
```
Byron, line 172: ⚠️Big Warning⚠️: Thin packs can't yet be resolved here, even though the machinery is in place it's not yet forwarded and/or fully implemented. The reason for that is that `gitoxide` only supports clones for now where packs are never thin.

I put that [on the list](https://github.com/Byron/gitoxide/issues/104) now to not forget.

test/src/test/integration/git_protocol.rs
```diff
@@ -0,0 +1,123 @@
...
+            executor::block_on,
+            future::{select, Either},
+            pin_mut,
+        };
+
+        pin_mut!(client);
+        pin_mut!(server);
+        block_on(async move {
+            match select(client, server).await {
+                Either::Left((res, _)) => res,
```
Byron, line 100: This is interesting: When waiting for the server and running the test it hangs even though the client is done. Here I would immediately assume that something is missing in the protocol, like a finishing empty request.

```suggestion
                Either::Left((res, server)) => {
                    server.await;
                    res
                }
```

Edit: _Thinking about it, it's a bit strange that the server still has a connection to listen on as `client_send` should already be dropped, stdin should also become unavailable for reading. Maybe related to async, or I am missing something_.

Having re-read the docs it became clear that [`gitoxide` is missing the empty request](https://github.com/git/git/blob/master/Documentation/technical/protocol-v2.txt#L136:L140) at the end of the operation. In V1 the session automatically ends after a packfile was received, but in V2 that's not the case anymore being a command-oriented protocol now. This is only an issue in persistent V2 connections as well like the ones used here.

I vaguely recall that the server shutting down isn't desired behaviour too and that sending a flush packet/empty request does indeed shut the server connection down. Could you elaborate on why that is, @kim ?

test/src/test/unit/git_protocol/upload_pack.rs
```diff
@@ -0,0 +1,129 @@
...
+use radicle_link_git_protocol::upload_pack;
+
+mod header {
+    use super::*;
+    use std::str::FromStr as _;
+
+    #[test]
+    fn service_must_be_upload_pack() {
+        assert_eq!(
+            upload_pack::Header::from_str("git-receive-pack "),
```
Byron, line 15: I really like the _simple_ errors which make testing less cumbersome to write. Using structured/enum based errors anywhere might have its perks but it for sure doesn't make testing easier.

Byron (commented):
Looking at the integration test was very helpful as I came to believe that the server hanging is a shortcoming of `gitoxide` that should be investigated and hopefully fixed. I don't fully understand what the expectations in the server are though as I think you mentioned it should not shut down when sending the flush packet, nor do I fully understand the interplay between stateless RPC and the daemon sending capabilities (instead of the upload-pack program).

git-protocol/src/fetch.rs
```diff
@@ -0,0 +1,493 @@
...
+pub type PackReceived = git_pack::bundle::write::Outcome;
+
+/// The default [`ReceivePack`].
+///
+/// Writes the packfile into the given output directory, along with a v2 index.
+/// The packfile is verified.
+pub struct DefaultReceivePack {
+    opt: ReceivePackOptions,
+    out: Option<PathBuf>,
+    stop: AtomicBool,
```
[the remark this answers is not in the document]
kim: Except that I actually need to tie flipping the bool to the lifetime of the `unblock` future :facepalm: The outer executor may drop it, but that won’t cancel the task.

git-protocol/src/fetch.rs
```diff
@@ -0,0 +1,493 @@
...
+        &mut self,
+        pack: impl AsyncBufRead + Unpin + 'async_trait,
+        prog: impl Progress,
+        _: &[Ref],
+        resp: &Response,
+    ) -> io::Result<()> {
+        self.out.wanted_refs = resp
+            .wanted_refs()
+            .iter()
+            .map(|response::WantedRef { id, path }| Ref::Direct {
```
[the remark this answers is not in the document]
kim: `oid` 😄 

Or else `id`, I like short identifiers.

git-protocol/src/fetch.rs
```diff
@@ -0,0 +1,493 @@
...
+
+    let conn = Connection::new(
+        recv,
+        send,
+        Protocol::V2,
+        opt.repo.clone(),
+        None::<(String, Option<u16>)>,
+        ConnectMode::Daemon,
+    );
+    blocking::unblock(move || {
```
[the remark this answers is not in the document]
kim: It is potentially the case that this could be hidden from the user, but my understanding is this:

 the `Executor` in `blocking` is responsible for managing blocking tasks on a thread pool. In order to run async code on the thread we end up on, we need to make an “async context” available to it. Just using a thread local (async!) executor happens to work, because our `Async{Read,Write}` will receive wakeups from the outer event loop. Another way to write the same would be to use `runtime.enter()` (both `tokio` and `async-*` have equivalents). This would cause the the inner async tasks to be scheduled on the _outer_ executor.

git-protocol/src/fetch.rs
```diff
@@ -0,0 +1,493 @@
...
+    .await
+}
+
+/// Like [`futures_lite::io::Take`], but returns an error if and when the
+/// `limit` is exceeded.
+///
+/// Note that, unlike [`futures_lite::io::Take`], if a single poll reads past
+/// the limit, the excess bytes are _not_ discarded. Instead, an error is
+/// returned on the next poll.
+pub struct TryTake<R> {
```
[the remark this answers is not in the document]
kim: Actually… I was hoping to read that I can use plain `take()` to bail out if the packfile exceeds the given limit, because the bundle writer is streaming and recognises a premature EOF without further processing. Is that the case?

git-protocol/src/upload_pack.rs
kim: I don’t think that’s the same: `try_join!` drives all futures in a loop, while `collect` only the head of the stream. Technically, we would even want `FuturesUnordered` or `select!`.

test/src/test/integration/git_protocol.rs
kim: I’m not sure actually if `futures_ringbuf` implements the dropping behaviour we’d expect here correctly. Perhaps a named pipe would be preferable for these kinds of scenarios, but the only one that turned up from a quick search was a semi-functional implementation in recent `tokio`.


To elaborate on the behaviour we want here: the stateful variants of the protocol are either `git-daemon` or SSH. Both imply some degree of trust between the parties. It is indeed the case that `link` could re-use the same connection (ie QUIC stream) to ask for a subsequent packfile (recall: we need to get at some signed data first, and then get the signed heads). However, we can not currently look into what’s going on from the server side, because we’re just shuffling bytes to `stdio` of `git-upload-pack`.

That creates a security concern: what if a client just does nothing, but keeps the connection (and thus, the upload-pack process) alive? It does not seem like the `—timeout` option has an effect in this case, but perhaps worth looking into why this is the case.

Since QUIC streams are (relatively) cheap, I would at this point prefer to run things in stateless mode, much like HTTP, but without the overhead of assembling HTTP headers. (_Aside: I have no idea how a multi-round negotiation could possibly work over HTTP_) Alternatively, we could wrap the receive stream and close it if either no bytes or too many bytes were received for some period of time, but that’s a lot more complicated. Ultimately, the “server” end will need to understand which commands are being sent in what order, and terminate the session if that doesn’t adhere to what a well-behaved `link` peer would do.

git-protocol/src/upload_pack.rs
kim: I had found https://github.com/smol-rs/futures-lite/issues/6, but that seems to have fizzled. Will need to contemplate if I like `try_zip(a, try_zip(b, c))` (although we use something like this elsewhere).

git-protocol/src/fetch.rs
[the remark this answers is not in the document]
Byron: Yes, I thought, without having tried it, that the unblock future returns a task which when drop will cancel the blocked evaluation - this is effective only for iterators. What if that task would be wrapped into another `Drop` implementing struct that then switches the `stop` flag? That should do something.
Similar things would probably have to happen for timeouts, but maybe there it's enough to close the connection and have everything stop that way.

git-protocol/src/fetch.rs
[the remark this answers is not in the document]
Byron: I just took a look but realized that in the first fully peeled Tag variant using `id` would be ambiguous as there are multiple `oids`, [one for the tag, one for the object referred to by the tag](https://github.com/Byron/gitoxide/blob/d6da891419f66208a8820185dd165e62b7a01a6e/git-protocol/src/fetch/refs.rs#L46-L49). From there the naming is actually consistent.
I could imagine to name things `object_id` and `tag_id` respectively but feel it's not really improving anything.
Naming is hard.

git-protocol/src/fetch.rs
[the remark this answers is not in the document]
Byron: I would have hoped to also hear an opinion in the linked thread, but thus far there is no involvement of the maintainer.
To me it looks a good case for `git-blocking` to eventually _unblock_ `blocking`s development 😅.

git-protocol/src/fetch.rs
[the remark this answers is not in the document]
Byron: Yes, it is streaming and will fail loudly if an entry can't be fully parsed. In the possible event of the amount of bytes taken coinciding with an entry boundary it will still fail as it wants to read a set amount of entries as advertised in the pack header, along with the trailer.
Thus partial packs will always result in an error.

git-protocol/src/upload_pack.rs
Byron: Additionally I think it's fair asking the value question, because even if `futures-util` would be removed here it's likely to be in a lot of other places within the project.
As a disclaimer, I might have a bias trying to replace `futures-util` with `futures-lite` to save 0.5s of compilation time, but it's clearly not worth it 😅.

Background: When working on an Intel Macbook Air I put a lot of effort into optimizing compile times of `prodash`, which now works without `futures-util`. That time seems to be stuck with me.

test/src/test/integration/git_protocol.rs
Byron: Thank you, all noted. I take it that the `git-upload-pack` process should definitely shut down at the end of the interaction, but it not doing so may also be related to the ring-buffer not working as expected on drop. Maybe that calls for a test on process level so the connection is real, or… bring up a real tcp socket along with a real connection. Of course, one would still have to hope that `tokio` tcp connections act accordingly on drop.

I will also run a test or two to verify it's not `gitoxide` misbehaving, and I have a strong feeling it is _in some circumstances_ and this being one of them.

> That creates a security concern: what if a client just does nothing, but keeps the connection (and thus, the upload-pack process) alive?
> Ultimately, the “server” end will need to understand which commands are being sent in what order, and terminate the session if that doesn’t adhere to what a well-behaved link peer would do.

These look like excellent requirements for the server implementation :).

> (Aside: I have no idea how a multi-round negotiation could possibly work over HTTP)

This one _I think_ I know - the client sends all the wants and haves from the previous round. The `Arguments` type [is implementing this](https://github.com/Byron/gitoxide/blob/0d5b911ad5f47ab8f044d6bbe660a6d1dfeecb5f/git-protocol/src/fetch/arguments/async_io.rs#L35). For V2 it always does that as it's supposed to be stateless, in V1 it's optional depending on whether the transport says so (i.e. V1 over HTTP is stateless, hence needs to resend all previous haves and wants).

test/src/test/integration/git_protocol.rs
```diff
@@ -0,0 +1,123 @@
...
+    .unwrap();
+
+    tmp
+}
+
+#[test]
+fn standard_clone() {
+    let upstream = upstream();
+
+    let (client_recv, server_send) = futures_ringbuf::Endpoint::pair(1024, 1024);
```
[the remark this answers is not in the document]
kim: I was holding this the wrong way: the `Endpoint`s are already wired up correctly, and we're supposed to use `AsyncReadExt::split` to get read-/write-halves.

Unfortunately, this still doesn't close `upload-pack`'s `stdin` after we're done fetching. It does not seem to be `futures_ringbuf`'s fault either.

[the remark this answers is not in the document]
Byron: It's probably `gitoxide` not getting the protocol right. [The fix](https://github.com/Byron/gitoxide/projects/1#card-64313808) will be applied tomorrow and I will test it against this PR - previously I was able to reproduce the hanging issue, too.

test/src/test/integration/git_protocol.rs
kim: I played around with this more and found:

* `--timeout` only [applies to v1](https://git.kernel.org/pub/scm/git/git.git/tree/builtin/upload-pack.c#n41)
* in v2 mode, `git-upload-pack` (without any involvement from `git-daemon`) will terminate the command loop if either [EOF](https://git.kernel.org/pub/scm/git/git.git/tree/serve.c#n217) or a [flush pkt](https://git.kernel.org/pub/scm/git/git.git/tree/serve.c#n236), and no command, is encountered
* in v2 mode, `--stateless-rpc` means that `git-upload-pack` will wait for exactly one command, process it, and then exit
* I still don't know why dropping one end of a `futures_ringbuf::Endpoint` pair doesn't lead to delivering EOF to `upload-pack`. If you drop `server` [here](https://github.com/najamelan/futures_ringbuf/blob/dev/examples/endpoint.rs#L25) and then try to read again, it works as expected.

So what I ended up with is to simply send a flush after we're done fetching (d4dd9d636cfc2416d25735196a369af6ded8d953). I conclude that this is the as-intended way to use v2 -- after all, we _do_ want to send "non-standard" command sequences eventually. Probably the flush should even be up to the caller.

I wonder where you see `gitoxide` misbehaving -- are you thinking of using a `Action` variant to decide whether to send a flush after a packfile was received?

git-protocol/src/fetch.rs
kim: Based on your current priorities, do you think I should use `libgit2` for now?

git-protocol/src/fetch.rs
```diff
@@ -0,0 +1,493 @@
...
+        }
+
+        for WantRef { name, have } in &self.opt.want_refs {
+            args.want_ref(name.as_bstr());
+            if let Some(oid) = have {
+                args.have(oid);
+            }
+        }
+
+        // send done, as we don't bother with further negotiation
```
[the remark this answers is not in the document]
kim: I _think_ it will at worst lead to a larger-than-necessary packfile, no?

test/src/test/integration/git_protocol.rs
Byron: > I wonder where you see gitoxide misbehaving -- are you thinking of using a Action variant to decide whether to send a flush after a packfile was received?

I see what you mean. From my point of view the operation is done once `git_protocol::fetch(…)` is completed and seeing what happened here it would be the least surprising to send a flush packet automatically when it's finished. I would add a default-implemented method to the delegate to allow to override this if there are other plans, like fetching multiple different packs.

git-protocol/src/fetch.rs
Byron: If there is a more-or-less equivalent call in `libgit2` to write a pack without much hassle, so using it is low-cost, I'd be inclined to say yes just to avoid me being a bottleneck.

Technically everything is prepared to support thin-pack resolution in _the machinery_, but it's not hooked up yet as there was no need previously. I put the `git-pack` improvements all the way up so it's done right after git-ref writing is finished.

git-protocol/src/fetch.rs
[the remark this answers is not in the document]
Byron: Probably, and some duplicate objects will be removed next time `git gc` is run (or some other command to clean things up as I don't know exactly what it does by default).
This also means that having a complete negotiation algorithm isn't needed right now and could be seen as performance improvement.

git-protocol/src/fetch.rs
```diff
@@ -433,32 +433,42 @@ where
...
+            opt.repo.clone(),
+            None::<(String, Option<u16>)>,
+            ConnectMode::Daemon,
+        );
+        let receive_pack = build_receive_pack(&stop);
+
+        move || {
+            let mut delegate = Fetch::new(opt, filter_refs, receive_pack);
+            future::block_on(async {
+                let res = git_protocol::fetch::fetch(
```
Byron, line 464: Warning: the new and only path for this `git_protocol::fetch(…)`. In future, if there are multiple paths the shorter one should be preferred as [all of these dual-paths](https://github.com/Byron/gitoxide/projects/1#card-65290657) will eventually be cleaned up.

git-protocol/src/fetch.rs
```diff
@@ -433,32 +433,42 @@ where
...
+                    &mut conn,
+                    &mut delegate,
+                    |_| unreachable!("credentials helper requested"),
+                    progress::Discard,
+                )
+                .await
+                .map_err(|e| io::Error::new(io::ErrorKind::Other, e));
+                flush_to_write(conn.into_inner().1).await.ok();
```
Byron, line 472: The delegate will now send a flush packet at the end of a successful operation, which probably leads to issues now as it would try to flush again in this line. 

If control over this should remain here, there is a [delegate method to override](https://github.com/Byron/gitoxide/blob/928f75ad939e35a159d3d2751d5d0f9d00d796af/git-protocol/src/fetch/delegate.rs#L122-L131).

git-protocol/src/fetch.rs
Byron: Tomorrow will be dedicated to thin-pack resolution, I wouldn't be surprised it's available when you begin your day. Stay tuned :).

Byron: Thin pack resolution is nearly there and [will look like this](https://github.com/Byron/gitoxide/blob/9f43bf029624f7c94346646465e366609b89e2e1/git-pack/tests/pack/data/output/count_and_entries.rs#L367). It's implemented [as an iterator](https://github.com/Byron/gitoxide/blob/9f43bf029624f7c94346646465e366609b89e2e1/git-pack/src/data/input/lookup_ref_delta_objects.rs#L40) which injects the looked-up entries into an existing pack-entry iteration.

However, even though I took my shot I wasn't lucky enough to get it right from the start and will need to add the utterly missing unit test tomorrow to figure out what part of the offset computations are off by one ;). This also means that I expect this to be ready by tomorrow.

Byron: So, even though it should be possible to recalculate pack-offsets and offset-deltas on the fly in the light of changing pack entry sizes due to entry-insertions and other size changes, for tonight I have to declare defeat :/.

If it would have worked, you would be the first to actually receive and write thin packs with `gitoxide`, as fetch-related tests and capabilities are still missing entirely, so in any case there is quite some work ahead to validate all of this.

kim: As per 0ed3c655322b4c03ceac4b9c0d148d6e9948b939, it's not _too_ painful to use libgit2 in the meantime. It does lock the odb (for whatever reason), so we need to keep an owned `Repository` around.

Byron: It's probably not a terrible idea to have `libgit2` ready at the flick of a feature toggle either, yet I hope that this code paths can soon be removed entirely.

This thin-pack problem has been annoying me in the past days but now I finally see the implementation to come to fruition. It's amazing how much work has to be done to do this correctly. The biggest hurdle is the absolutely unnecessary number of objects in the pack header, which causes extra work for particular advantage especially since the file content is secured by a trailing hash anyway. Definitely something I would change in a V3 pack format.

Without having studied how git handles thin pack resolution at all I can at least say that the `gitoxide` implementation is fully streaming even with thin-pack resolution, so resolving referenced base objects happens on the fly as well as writing the pack to disk, all while it's being indexed in preparation for producing the hashes required for the pack index file. When cloning `gitoxide` is already faster than git, and when fetching I think it will be the same.

Byron: With e7b9d9613874cd1ebaf740dc08db467c461a4751 thin packs can be resolved. 
The bundle writer now [has an additional parameter](https://github.com/Byron/gitoxide/blob/1852e3ea98a462958862ab05f110649e3b06e2b5/git-pack/src/bundle/write/mod.rs#L38) to allow resolving arbitrary objects using an object db. Typically [it's implemented like this](https://github.com/Byron/gitoxide/blob/990ea4866be2d22ae2043da2dcd9577b748de255/git-pack/tests/pack/data/output/count_and_entries.rs#L373-L375) if a thin pack is in the realm of possibilities.

I think it has all the tests it needs already to assure it's working as expected.
The pack-generation machinery was also update to be able to produce thin packs itself which is where most of the thorough validation is coming from right now.

kim:
This is now ready, I think.

FintanH (dismissed):
Looks good :ok_hand:

test/src/test/integration/git_protocol.rs
```diff
@@ -0,0 +1,384 @@
...
+            &build_pack_writer,
+        )
+        .unwrap();
+
+        assert!(server_out.success());
+        assert!(client_out.pack.is_some());
+    }
+
+    let remote_repo = git2::Repository::open(&remote).unwrap();
+    remote_repo.set_namespace("foo").unwrap();
```
Byron, line 328: Adding namespace support is on the list of features to add to the gitoxide loose ref store now.

test/src/test/integration/git_protocol.rs
```diff
@@ -0,0 +1,384 @@
...
+    let local = tempdir().unwrap();
+
+    thin_pack_with(&remote, &local, |stop| {
+        let local_repo = git2::Repository::init(&local).unwrap();
+        packwriter::Libgit::new(packwriter::Options::default(), local_repo, stop)
+    })
+}
+
+#[test]
+fn thin_pack_gitoxide() {
```
Byron, line 375: I am definitely loving these tests and am glad it does indeed work as expected.
Currently thin pack support is tested on a thin pack of `gitoxide`'s own making _only_.

test/src/test/integration/git_protocol.rs
```diff
@@ -0,0 +1,384 @@
...
+
+fn upstream() -> TempDir {
+    let tmp = tempdir().unwrap();
+
+    let repo = git2::Repository::init_bare(&tmp).unwrap();
+    let auth = git2::Signature::now("apollo", "apollo@cree.de").unwrap();
+
+    let tree = {
+        let empty = repo.treebuilder(None).unwrap();
+        let oid = empty.write().unwrap();
```
Byron, line 35: Actually all of work done here except for `Repository::init_bare(…)` and `.set_namespace()` can already be done in `gitoxide`. Maybe once all is supported I could contribute a PR which oxidizes the test setup.

Byron (approved):
Probably it's best to use `libgit2` pack writing for now as the `gitoxide` implementation of pack writing exists in isolation, that is isn't yet aware of any additional requirements that might apply to writing packs while dealing with locks, multi-pack indices and the likes.
Some more research should be done to be sure there is no unwanted side effects.

On the other hand, all it does it to write tmp files and rename the finished files to move into their desired spot. This _should_ be find if git is indeed backwards compatible with itself.

I also left  a comment about using `gitoxide` for setting up the test fixture, maybe that could be done in a separate PR contribute by me if there is interest.

Besides that: YAY 🎉, `gitoxide` lands in Radicle-Link :)

git-protocol/src/packwriter.rs
```diff
@@ -0,0 +1,214 @@
...
+        let odb =
+            Store::at(self.git_dir.clone()).map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
+        let opts = Options {
+            thread_limit: self.opt.max_indexer_threads,
+            index_kind: Version::V2,
+            iteration_mode: Mode::Verify,
+        };
+        Bundle::write_to_directory(
+            BlockOn::new(TryTake::new(pack, self.opt.max_pack_bytes)),
+            Some(self.git_dir.join("objects/pack")),
```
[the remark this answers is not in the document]
kim: :disappointed:

git-protocol/src/fetch.rs
```diff
@@ -0,0 +1,397 @@
...
+fn remote_git_version(caps: &client::Capabilities) -> Option<Version> {
+    let agent = caps.capability("agent").and_then(|cap| {
+        cap.value()
+            .and_then(|bs| bs.to_str().map(|s| s.to_owned()).ok())
+    })?;
+    Version::new(agent.strip_prefix("git/")?)
+}
+
+fn remote_supports_ref_in_want(caps: &client::Capabilities) -> bool {
+    let ref_in_want = BString::from("ref-in-want");
```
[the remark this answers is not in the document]
kim: Nice!

test/src/test/integration/git_protocol.rs
kim: That'd be awesome.

PR #725, opened 2021-07-05
Fix rere related failing replication

alexjg:
Introduce a test case which reproduces a strange failure to replicate which only occurs when `rere` is enabled with the patched libgit which respects force flags. Issue to describe the problem in more detail incoming.


alexjg:
The fix was in the patch of libgit that we're using. I've left the test in because I think it's useful to have some coverage of this path - it being a somewhat common one.

FintanH (approved):
Glad you found the issue :) I retried the test suite because it was daemon tests failing.

test/src/test/integration/librad/scenario/tracked_references.rs
```diff
@@ -0,0 +1,134 @@
+// Copyright © 2019-2020 The Radicle Foundation <hello@radicle.foundation>
```
xla, line 1: This is the old license header, let's stick to the updated one for all new files.

test/src/test/integration/librad/scenario/tracked_references.rs
```diff
@@ -0,0 +1,134 @@
...
+        identities,
+        storage::ReadOnlyStorage,
+        tracking,
+        types::{Namespace, Reference},
+        util::quick_commit,
+    },
+    git_ext::tree,
+    reflike,
+};
+use std::ops::Index as _;
```
xla, line 21: Minor nit, some consistency in import order would make my inner OCdaemon happy.

test/src/test/integration/librad/scenario/tracked_references.rs
```diff
@@ -0,0 +1,134 @@
...
+
+fn config() -> testnet::Config {
+    testnet::Config {
+        num_peers: nonzero!(2usize),
+        min_connected: 2,
+        bootstrap: testnet::Bootstrap::from_env(),
+    }
+}
+
+#[test]
```
xla, line 31: The attribute seems a bit lost with a big block of comment between it and the function it applies to.

test/src/test/integration/librad/scenario/tracked_references.rs
```diff
@@ -0,0 +1,134 @@
...
+/// Peers should be able to see references created by peers in their tracking
+/// graph. To test this we do the following:
+///
+/// - Create a testnet with two peers, peer1, and peer2
```
xla, line 35:
```suggestion
```

Can be assumed.

xla (commented):
utACK. Seems legit. :>

test/src/test/integration/librad/scenario/tracked_references.rs
alexjg: You're referring to this right?:

```
// Copyright © 2021 The Radicle Link Contributors
//
// This file is part of radicle-link, distributed under the GPLv3 with Radicle
// Linking Exception. For full terms see the included LICENSE file.
```

xla: yus

test/src/test/integration/librad/scenario/tracked_references.rs
alexjg: Interesting, what consistency are you referring to? `rust-analyzer` arranged these imports in alphabetical order, are you thinking of something like nested imports appearing below non-nested imports?

xla: Dropping an example which show-cases order: std, thrid-party crates, first-party crates, super, crate:

https://github.com/radicle-dev/radicle-link/blob/33c0db4402fb81c1e10bd7742531d972afc98379/librad/src/net/protocol.rs#L6-L33

alexjg: Thanks! Interestingly you have to have a new line between each category or `cargo fmt` will just silently reorder everything to alphabetical order for you.

xla (approved):
↩️ 🌝 📖 🎌

PR #727, opened 2021-07-12
daemon: use tracing crate and improve logging

geigerzaehler:
We replace the `log` crate with `tracing` for `daemon` and clean up some logging statements.

### DIFF
--- a/daemon/src/peer/gossip.rs
+++ b/daemon/src/peer/gossip.rs
@@ -26,8 +26,8 @@ where
         rev: rev.map(|rev| Rev::Git(rev.into())),
         origin: None,
     }) {
-        Ok(()) => log::trace!("successfully announced for urn=`{}`, rev=`{:?}`", urn, rev),
-        Err(_payload) => log::warn!("failed to announce for urn=`{}`, rev=`{:?}`", urn, rev),
+        Ok(()) => tracing::trace!(%urn, ?rev, "successfully announced URN"),
+        Err(_payload) => tracing::warn!(%urn, ?rev, "failed to announce URN"),
     }
 }
 
@@ -41,11 +41,7 @@ where
         rev: None,
         origin,
     }) {
-        Ok(()) => log::trace!(
-            "successfully queried for urn=`{}`, origin=`{:?}`",
-            urn,
-            origin
-        ),
-        Err(_payload) => log::warn!("failed to query for urn=`{}`, origin=`{:?}`", urn, origin),
+        Ok(()) => tracing::trace!(%urn, ?origin, "successfully queried URN"),
+        Err(_payload) => tracing::warn!(%urn, "failed to query URN"),
     };
 }
--- a/daemon/src/peer/include.rs
+++ b/daemon/src/peer/include.rs
@@ -14,8 +14,7 @@ pub async fn update<S>(peer: Peer<S>, urn: Urn)
 where
     S: Clone + Signer,
 {
-    match state::update_include(&peer, urn.clone()).await {
-        Ok(path) => log::debug!("Updated include file @ {}", path.display()),
-        Err(err) => log::debug!("Failed to update include file for `{}`: {}", urn, err),
+    if let Err(err) = state::update_include(&peer, urn.clone()).await {
+        tracing::error!(%urn, error = ?err, "Failed to update include file");
     }
 }
--- a/daemon/src/peer/run_state.rs
+++ b/daemon/src/peer/run_state.rs
@@ -199,7 +199,7 @@ impl RunState {
     /// new state and in some cases produes commands which should be
     /// executed in the appropriate subroutines.
     pub fn transition(&mut self, input: Input) -> Vec<Command> {
-        log::trace!("TRANSITION START: {:?} {:?}", input, self.status);
+        tracing::trace!(?input, status = ?self.status, "transition start");
 
         let cmds = match input {
             Input::Announce(announce_input) => self.handle_announce(announce_input),
@@ -211,7 +211,7 @@ impl RunState {
             Input::Stats(stats_input) => self.handle_stats(stats_input),
         };
 
-        log::trace!("TRANSITION END: {:?} {:?}", self.status, cmds);
+        tracing::trace!(?cmds, status = ?self.status, "transition end");
 
         cmds
     }
@@ -356,7 +356,7 @@ impl RunState {
                     urn,
                 },
             ) => {
-                log::warn!("Cloning failed with: {}", reason);
+                tracing::warn!(?reason, "cloning failed");
                 self.waiting_room
                     .cloning_failed(&urn, remote_peer, SystemTime::now(), reason)
             },
--- a/daemon/src/peer/run_state/running_waiting_room.rs
+++ b/daemon/src/peer/run_state/running_waiting_room.rs
@@ -328,7 +328,7 @@ impl RunningWaitingRoom {
             },
             // FIXME(alexjg): Figure out how to report this error to the client
             Err(error) => {
-                log::warn!("WaitingRoom::Error : {}", error);
+                tracing::warn!(?error, "waiting room error");
                 Vec::new()
             },
         }
--- a/daemon/src/peer/subroutines.rs
+++ b/daemon/src/peer/subroutines.rs
@@ -88,7 +88,7 @@ where
         };
         let waiting_room = match waiting_room::load(&store) {
             Err(err) => {
-                log::warn!("Failed to load waiting room: {}", err);
+                tracing::warn!(?err, "Failed to load waiting room");
                 WaitingRoom::new(request::waiting_room::Config {
                     delta: config::DEFAULT_WAITING_ROOM_TIMEOUT,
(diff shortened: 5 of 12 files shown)


daemon/src/peer/run_state.rs
```diff
@@ -199,7 +199,7 @@ impl RunState {
     /// new state and in some cases produes commands which should be
     /// executed in the appropriate subroutines.
     pub fn transition(&mut self, input: Input) -> Vec<Command> {
```
alexjg, line 201: Might be worth add ing a `#[instrument]` to this function and removing the "start" and "end" traces as they serve the same purpose.

geigerzaehler: We’re also printing `self.status` and the resulting `cmds` which would not be possible with `instrument`. Is that ok?

alexjg: I think you should be able to do that with something like (based on my reading of https://docs.rs/tracing/0.1.26/tracing/attr.instrument.html#examples-1):

```rust
#[instrument(skip(self), fields(status=self.status)]
```

geigerzaehler: Nice! https://github.com/radicle-dev/radicle-link/pull/727/commits/e29d2bd7c682e38f55a5c68d1cdfa9ef28525609

alexjg: Ugh, I'm actually wrong about this. The `instrument` macro generates a span which records the fields when the span is entered. I forgot that the `status` changes as a result of the `transition` and that won't be recorded by the span. Your original code was better I think. My apologies.

daemon/src/peer/gossip.rs
```diff
@@ -26,8 +26,8 @@ where
         rev: rev.map(|rev| Rev::Git(rev.into())),
         origin: None,
     }) {
-        Ok(()) => log::trace!("successfully announced for urn=`{}`, rev=`{:?}`", urn, rev),
-        Err(_payload) => log::warn!("failed to announce for urn=`{}`, rev=`{:?}`", urn, rev),
+        Ok(()) => tracing::trace!(?urn, ?rev, "successfully announced URN"),
```
FintanH, line 29:
```suggestion
        Ok(()) => tracing::trace!(%urn, ?rev, "successfully announced URN"),
```

`Urn`s are easier to debug with their `Display`s being used :)

daemon/src/peer/gossip.rs
```diff
@@ -41,11 +41,7 @@ where
...
         origin,
     }) {
-        Ok(()) => log::trace!(
-            "successfully queried for urn=`{}`, origin=`{:?}`",
-            urn,
-            origin
-        ),
-        Err(_payload) => log::warn!("failed to query for urn=`{}`, origin=`{:?}`", urn, origin),
+        Ok(()) => tracing::trace!(?urn, "successfully queried URN"),
+        Err(_payload) => tracing::warn!(?urn, "failed to query URN"),
```
FintanH, line 45: Any reason we removed the `origin` from the message?

daemon/src/peer/include.rs
```diff
@@ -15,7 +15,7 @@ where
     S: Clone + Signer,
 {
     match state::update_include(&peer, urn.clone()).await {
```
FintanH, line 17: Could change this to to `if let Err(err)`

daemon/src/peer/run_state/running_waiting_room.rs
```diff
@@ -328,7 +328,7 @@ impl RunningWaitingRoom {
             },
             // FIXME(alexjg): Figure out how to report this error to the client
             Err(error) => {
-                log::warn!("WaitingRoom::Error : {}", error);
+                tracing::warn!(?error, "waiting room error");
```
FintanH, line 331: Curious as to why you use `?` instead of `%` in a lot of places?

daemon/src/peer/sync.rs
```diff
@@ -26,18 +26,18 @@ where
         .collect::<Vec<_>>();
 
     for urn in urns {
-        log::debug!("Starting fetch of {} from {}", urn, remote_peer);
+        tracing::debug!(?urn, ?remote_peer, "startin fetch");
```
FintanH, line 29:
```suggestion
        tracing::debug!(?urn, ?remote_peer, "starting fetch");
```

Alternatively, `"startin'"` :stuck_out_tongue_winking_eye:

daemon/src/peer/run_state/running_waiting_room.rs
geigerzaehler: Usually the `Debug` implementation provides more information about the printed value. For errors in particular using `Display` may omit information like constructor names if `#[error(transparent)]` is used or any field that is not explicitly mentioned in the error message. Of, course for URN it makes more sense to use `%`.

daemon/src/peer/gossip.rs
geigerzaehler: Fixed in e0f8ba665088506a52784ac5335cbb639d40818b

daemon/src/peer/gossip.rs
geigerzaehler: Fixed in e0f8ba665088506a52784ac5335cbb639d40818b

daemon/src/peer/include.rs
geigerzaehler: Fixed in e0f8ba665088506a52784ac5335cbb639d40818b

daemon/src/peer/run_state.rs
geigerzaehler: Reverted in ea907e4862c809649133b12ef0fa01299fd43d40

daemon/src/peer/subroutines.rs
```diff
@@ -405,7 +405,7 @@ where
                 .ok();
         },
         Err(err) => {
-            log::error!("sync error for {}: {:?}", peer_id, err);
+            tracing::error!(?peer_id, ?err, "sync error");
```
FintanH, line 408:
```suggestion
            tracing::error!(%peer_id, ?err, "sync error");
```

daemon/src/peer/subroutines.rs
```diff
@@ -456,11 +456,11 @@ async fn clone<S>(
...
         },
         Err(err) => {
-            log::warn!(
-                "an error occurred for the command 'Clone' for the URN '{}' from {}:\n{}",
-                urn,
-                remote_peer,
-                err,
+            tracing::warn!(
+                %urn,
+                ?remote_peer,
```
FintanH, line 461:
```suggestion
                %remote_peer,
```

daemon/src/peer/sync.rs
```diff
@@ -17,7 +17,7 @@ pub async fn sync<S>(peer: &Peer<S>, remote_peer: PeerId) -> Result<(), Error>
 where
     S: Clone + Signer,
 {
-    log::debug!("Starting sync from {}", remote_peer);
+    tracing::debug!(?remote_peer, "Starting sync");
```
FintanH, line 20:
```suggestion
    tracing::debug!(%remote_peer, "Starting sync");
```

daemon/src/peer/sync.rs
```diff
@@ -26,18 +26,18 @@ where
         .collect::<Vec<_>>();
 
     for urn in urns {
-        log::debug!("Starting fetch of {} from {}", urn, remote_peer);
+        tracing::debug!(?urn, ?remote_peer, "starting fetch");
```
FintanH, line 29:
```suggestion
        tracing::debug!(%urn, %remote_peer, "starting fetch");
```

daemon/src/peer/sync.rs
```diff
@@ -26,18 +26,18 @@ where
...
         match state::fetch(peer, urn.clone(), remote_peer, vec![], None).await {
             Ok(result) => {
-                log::debug!(
-                    "Finished fetch of {} from {} with the result {:?}",
-                    urn,
-                    remote_peer,
-                    result.updated_tips
+                tracing::debug!(
+                    %urn,
+                    ?remote_peer,
```
FintanH, line 34:
```suggestion
                    %remote_peer,
```

daemon/src/peer/sync.rs
```diff
@@ -26,18 +26,18 @@ where
...
+                    updated_tips = ?result.updated_tips,
+                    "finished fetch",
                 );
                 include::update(peer.clone(), urn).await;
             },
-            Err(e) => log::debug!("Fetch of {} from {} errored: {}", urn, remote_peer, e),
+            Err(error) => tracing::error!(%urn, ?remote_peer, ?error, "fetch error"),
```
FintanH, line 40:
```suggestion
            Err(error) => tracing::error!(%urn, %remote_peer, ?error, "fetch error"),
```

FintanH (commented):
We should also display `PeerId`s :)

FintanH (dismissed):

geigerzaehler:
@alexjg: Can I get a second approval? :pray:

alexjg (approved):

PR #728, opened 2021-07-16
rad-profile: the rad-profile CLI library

FintanH:
Currently `ssh_add` is a stub since there is some reworking in `thrussh` required to get https://github.com/radicle-dev/radicle-keystore/pull/17 in.


FintanH:
This is ready for some inspection :female_detective:

link-crypto/Cargo.toml
```diff
@@ -33,6 +32,11 @@ features = ["std", "derive"]
 path = "../git-ext"
 features = ["serde", "minicbor"]
 
+[dependencies.radicle-keystore]
+git = "https://github.com/radicle-dev/radicle-keystore"
+rev = "2a28dd785f7cd360cdc962e4d343fa9f7ac39814"
```
FintanH, line 37: Although we should probably wait for merging until a minor release is made[0]

[0]: https://github.com/ZcashFoundation/ed25519-zebra/issues/45#issuecomment-820181394

FintanH:
oh lol, what do we do for Windows? :laughing: `thrussh` has no implementation for `not(unix)` :grimacing:

kim:
>    thrussh has no implementation for not(unix)

Is this because std doesn't export UNIX domain sockets under windows? That's
more a matter of policy it seems [0]. Oh look, a yak!

[0]: https://github.com/rust-lang/rust/issues/56533

FintanH:
> Is this because std doesn't export UNIX domain sockets under windows? That's
more a matter of policy it seems [0]. Oh look, a yak!

Thanks for the pointer :) Well, this would mean that we'd have to also get `tokio` and `smol` to add the Windows support so that in turn it can be supported by `thrussh`. 

image

kim:
>    get tokio and smol to add the Windows support

Yeah right.

I think this is probably the point in time where we no longer pretend we could
support Windows eventually (outside WSL, that is). Or else, someone steps up to
become a maintainer for Windows-compat.

FintanH:
> I think this is probably the point in time where we no longer pretend we could
support Windows eventually (outside WSL, that is). Or else, someone steps up to
become a maintainer for Windows-compat.

I think we can at least punt on this for now with https://github.com/radicle-dev/radicle-link/pull/728/commits/a04a95407a54b0ea4a6120ecc8edf3433abe635a, but ya it doesn't seem feasible with all of us working on unix based machines as well.

The only other option I see is hiding all ssh-agent functionality behind a `cfg(unix)` and forcing windows users to use the `prompt` method of key access :upside_down_face:

rad-profile/src/cli.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright © 2021 The Radicle Link Contributors
+//
+// This file is part of radicle-link, distributed under the GPLv3 with Radicle
+// Linking Exception. For full terms see the included LICENSE file.
+
+pub mod args;
+pub mod main;
+
+pub use main::main;
```
xla, line 9: How does this interact with the `pub mod main;` above?

rad-profile/src/cli/args.rs
```diff
@@ -0,0 +1,78 @@
...
+}
+
+/// Create a new profile, generating a new secret key and initialising
+/// configurations and storage.
+#[derive(Debug, StructOpt)]
+pub struct Create {}
+
+/// Get the currently active profile.
+#[derive(Debug, StructOpt)]
+pub struct Get {}
```
xla, line 35: Could this take an optional `id` to get another profile than the currently active?

xla: If not it might be more accurately named `GetActive`.

xla (commented):
Cool beans! Are we concerned that there are no destructive actions at the moment, e.g. `profile delete` or `ssh remove`?

link-crypto/Cargo.toml
FintanH: Would be preferable, I think, for the upstream release to be made. They said they'd make a decision tomorrow :)

rad-profile/src/cli.rs
FintanH: I believe Rust knows what's up. I was playing around with a `src/main.rs` locally and I made this change to make it more sensible than `rad_exe::main::main()` :smile:

rad-profile/src/cli/args.rs
FintanH: Ya, that's a good point. Need to revisit why I didn't do that initially :)

FintanH:
> Cool beans! Are we concerned that there are no destructive actions at the moment, e.g. profile delete or ssh remove?

Very good point. Aside: it's funny `ssh-add` can also delete a key haha. But I don't see why we couldn't add destructive operations.

kim:
>    it's funny ssh-add can also delete a key

From the agent, which is ephemeral storage. Think of it more as "unload key".

I believe it was mentioned in RFC #682 that secure long term storage of key
material is still a concern.

FintanH:
> From the agent, which is ephemeral storage. Think of it more as "unload key".

Ya, good way of putting it. I was just musing about the name of `add` :)

> I believe it was mentioned in RFC #682 that secure long term storage of key
material is still a concern.

Long-term storage as well as rotation I would imagine

xla (dismissed):
🎤 🗑 📏 🛎

xla (commented):
This yak looks sufficiently shaved.

⛏ 📘 📙 🎠

xla (approved):
⚱ 🖨 🏺 🍂

PR #729, opened 2021-07-20
daemon: only log with trace level for subroutines

geigerzaehler:
Logging the `Subrioutines` input is quite noisy because we have frequent “tick” events that often have no effect. If they have an effect we should add logging statements in those places.


kim (approved):
lgtm :shrug:

FintanH (approved):
Whoops, missed this. Love a one-line change :ok_hand:

FintanH (dismissed):

PR #730, opened 2021-07-20
librad: pull periodic tasks instead of spawning

geigerzaehler:
We create a stream with periodic membership tasks that needs to be pulled instead of spawning a tasks that pumps the membership tasks into a channel. This has the advantage that now detached task needs to be spawned and that we stop issuing membership tasks when future retuned by `accept()` is dropped.

### DIFF
--- a/librad/src/net/protocol/membership/hpv.rs
+++ b/librad/src/net/protocol/membership/hpv.rs
@@ -11,7 +11,7 @@ use std::{
 };
 
 use data::BoundedVec;
-use futures::channel::mpsc;
+use futures::Stream;
 use parking_lot::RwLock;
 use rand::seq::IteratorRandom as _;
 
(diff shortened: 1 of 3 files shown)


librad/src/net/protocol/membership/hpv.rs
```diff
@@ -124,20 +123,18 @@ where
     Addr: Clone + Debug + PartialEq,
 {
     pub fn new(
-        spawner: &executor::Spawner,
         local_id: PeerId,
         rng: Rng,
         params: Params,
-    ) -> (Self, mpsc::Receiver<Periodic<Addr>>)
+    ) -> (Self, BoxStream<'static, Periodic<Addr>>)
```
kim, line 129: Nit: can we defer the boxing until absolutely needed (ie. return `impl Stream` here)?

kim (approved):

kim (dismissed):

FintanH (approved):

PR #731, opened 2021-07-20
daemon: call protocol cleanup when peer is dropped

geigerzaehler:
`Peer::run` now properly cleans up the network stack when run to
completion. To this end it accepts a shutdown signal as an argument.
This seems to fix [deadlocks][1] that occur when the daemon is torn down
multiple times.

Instead of copying the approach from `seed/src/lib.rs` we use a shutdown
future and port this approach to the seed. The reason for this is that
the seed also did not handle shutdown properly. If a termination signal
was sent it did not break the main loop. Moreover we did not react to
shutdown requests while awaiting `peer.bind()`.

[1]: radicle-dev/radicle-upstream#2133

### DIFF
--- a/daemon/src/peer.rs
+++ b/daemon/src/peer.rs
@@ -9,8 +9,9 @@
 use std::{io, net::SocketAddr, vec};
 
 use futures::{
-    future::{FutureExt as _, TryFutureExt as _},
+    future::{Either, FutureExt as _, TryFutureExt as _},
     stream::StreamExt as _,
+    Future,
 };
 use tokio::{
     sync::{broadcast, mpsc, watch},
@@ -101,7 +102,8 @@ where
 {
     /// Constructs a new [`Peer`].
     ///
-    /// To kick-off the peer's subroutines be sure to use [`run`][`Self::run`].
+    /// To kick-off the peer's subroutines be sure to use
+    /// [`start`][`Self::start`].
     ///
     /// # Errors
     ///
@@ -122,7 +124,8 @@ where
 
     /// Construct a new [`Peer`] using an existing [`net::peer::Peer`].
     ///
-    /// To kick-off the peer's subroutines be sure to use [`run`][`Self::run`].
+    /// To kick-off the peer's subroutines be sure to use
+    /// [`start`][`Self::start`].
     #[must_use = "give a peer some love"]
     pub fn with_peer(
         peer: net::peer::Peer<S>,
@@ -159,8 +162,14 @@ where
         self.subscriber.subscribe()
     }
 
-    /// Start up the internal machinery to advance the underlying protocol,
-    /// react to significant events and keep auxiliary tasks running.
+    /// Returns a future that runs the peer and a handle that shuts the peer
+    /// down when it is dropped.
+    ///
+    /// The function only returns when an error occurs or the [`Shutdown`] value
+    /// is dropped. becomes ready.
+    ///
+    /// The future returned by this function must be run to completion for the
+    /// daemon to shut down properly.
     ///
     /// # Errors
     /// * Failed to accept peer connections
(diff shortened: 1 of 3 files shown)


daemon/src/peer.rs
```diff
@@ -204,10 +204,13 @@ where
...
                             .send(bound.listen_addrs())
                             .expect("subroutines is gone");
-                        let (_, run) = net::protocol::accept(bound, disco.clone().discover());
+                        let (close_endpoint, run) =
+                            net::protocol::accept(bound, disco.clone().discover());
+                        let close_endpoint = OnDrop::new(close_endpoint);
                         if let Err(e) = run.await {
                             tracing::error!(error = ?e, "accept error");
                         }
+                        drop(close_endpoint);
```
FintanH, line 213: For my educational purposes, why are these gymnastics required? Why can't we just call `close_endpoint` here?

kim:
I think this might not be the whole story: if it is the case that there still is a live reference to the endpoint after the future returned from `run` is dropped, then it must be that either the `accept` future is not dropped, or somehow destructors are not run (causing resource leakage). When calling the stop closure, you _still_ need to ensure that the `run` future is run to completion, which may only accidentally be the case currently.

Iow, it is likely preferable to move that re-bind loop as close to `main` as possible, and install proper signal handlers (like [here](https://github.com/radicle-dev/radicle-link/blob/master/seed/src/lib.rs#L202)).

daemon/src/peer.rs
kim: Alex Crichton calls this a "drop bomb" (or bomb drop?): the compiler cannot drop the value earlier. But dropping the future while we're at `run.await` won't run any code after that point.

geigerzaehler:
I’ve updated the PR. I adjusted the approach from `seed/src/lib.rs` to make it more robust. Details are in the updated description.

FintanH (approved):

geigerzaehler:
The corresponding [PR in upstream](https://github.com/radicle-dev/radicle-upstream/pull/2163) shows that this seems to work. (The single test failure there is another issue.)

daemon/src/peer.rs
```diff
@@ -162,16 +163,27 @@ where
...
+    ///
     /// # Errors
     /// * Failed to accept peer connections
     /// * A subroutine panicked or was cancelled
     ///
     /// # Panics
     /// * If the subroutine is gone when the protocol network is still setting
     ///   up shop.
-    pub async fn run(self) -> Result<(), Error> {
+    pub async fn run(self, shutdown: impl Future<Output = ()> + Send) -> Result<(), Error> {
```
kim, line 179: It's not entirely clear to me why this needs to be a future -- arguably an `AtomicBool` or somesuch is much more convenient for the caller, as it can be flipped in a `Drop` impl without hoop jumping. Afaiu, the only problem with the previous approach was that the loop was not broken by matching on the `Done` error.

geigerzaehler: It would be possible to change the argument to `AtomicBool`. I am right that would require us to spawn another task or thread that watches the bool and then calls `stop_accepting()` when it flips to true? The seed code was accomplishing this by spawning a thread. This is something I wanted to avoid and why I went for a future. Happy to implement this differently though.

On the other hand creating the shutdown future with a `oneshot` is not to complicated.
```rust
let (shutdown_tx, shutdown_rx) = tokio::sync::oneshot::channel();
peer.run(shutdown_rx);
// ...
shutdown_tx.send()
```

versus with `AtomicBool`
```rust
let shutdown = Arc::new(AtomicBool::new(false));
peer.run(shutdown.clone());
// ...
shutdown.store(true, Ordering::Relaxed)
```

> Afaiu, the only problem with the previous approach was that the loop was not broken by matching on the Done error.

There was another issue in the seed: When the signal was received while awaiting `.bind()` the loop would not be broken because the old `accept` was stopped but not the new one.

kim: Mya, I'm not objecting to performing any sort of oneshot gymnastics _internally_, but I think you're unnecessarily putting the responsibility for preventing premature dropping of futures on the caller. As I can see from the companion PR, this results in avoidable complexity.

As a refresher, there are two main reasons for the requirement to run the network loop future to completion:

1. When working with multiple runtimes (such as in tests), deadlocks occur because spawned blocking tasks frequently need to re-enter the async runtime in order to pull data from a QUIC stream. Due to tokio's decidedly braindead shutdown behaviour, this can result in the await point never resolving because tokio already dropped all non-blocking futures. Because blocking tasks cannot be cancelled, the program hangs indefinitely. To avoid this, we close the endpoint and keep driving all futures until the terminate by themselves (eg. by receiving an IO error).
2. When working with a single runtime (such as in an executable), it is still important (up to and excluding `SIGKILL`) to let the QUIC stack drain all outstanding connections. Otherwise, peers won't notice that the connection is half-closed, and need to sit out the idle period. We currently are lenient in that we allow a peer to establish more than one connection, but that may change again.

With that out of the way, I think the simplest (read: hardest to fuck up) API for this is to _return_ a callable which shuts down the stack, a la:

```rust
pub fn run(self) -> (Shutdown, impl Future<Output = Result<!, Error>)
```

Where `Shutdown` contains either some `AtomicRef`-like with the current shutdown function, or is wired up using a channel. Either way, it's trigger would be a blocking function, or a `Drop` impl at your discretion.

The UX is then roughly:

```rust
let (kill, run) = peer.run();
spawn(run);
// do stuff
drop(kill)
```

Which arguably is a lot simpler than what you have now.

geigerzaehler: > With that out of the way, I think the simplest (read: hardest to fuck up) API for this is to _return_ a callable which shuts down the stack

Makes sense. I’ve implemented this in https://github.com/radicle-dev/radicle-link/pull/731/commits/166cd93ffabf3e69f6d0077ae4eee17afccb7756.

kim (approved):
Great, thanks!

FintanH (dismissed):

kim (approved):

PR #736, opened 2021-07-27
librad: fix deadlock when waiting for idle

geigerzaehler:
We fix a deadlock when we wait for the quic endpoint to become idle.

The deadlock happens when we wait for `endpoint.wait_idle()` while we are simultaneously still trying to connect to a remote that does not respond (i.e. it neither accepts nor refuses the connection) in `connect` called by `librad::net::protocol::io::discovered()`.

To make `endpoint.wait_idle()` future finish we need to advance the connection task so we drop it after the future has finished instead of before.

### DIFF
--- a/librad/src/net/protocol.rs
+++ b/librad/src/net/protocol.rs
@@ -255,9 +255,9 @@ where
         async move {
             let res = io::connections::incoming(state, incoming).await;
             drop(_git_factory);
-            drop(tasks);
             tracing::debug!("waiting on idle connections...");
             endpoint.wait_idle().await;
+            drop(tasks);
             tracing::debug!("protocol shut down");
             res
         }


FintanH (commented):
Is there any way of confirming that this fixes a deadlock?

geigerzaehler:
> Is there any way of confirming that this fixes a deadlock?

I’ve prepared a branch `thomas/reproduce-deadlock-fix` on Upstream. To reproduce the error checkout the second to last commit on that branch and run `yarn run test:integration:debug`. After cypress opens up click on the networking test to run it. After the test has passed, wait for the log line `establishing connection remote_addr=64.190.62.111:123` in the terminal. Then hit `Ctrl+R` in the test window to re-run the test. The second run fails and in the console you can see that after a GET request to `/v1/control/reset` we don’t see the log line that tells us that the network protocol has shut down.

If you move to the head of the branch the proxy uses the updated librad version and everything works as expected.

kim (dismissed):
Good catch

FintanH (approved):

PR #738, opened 2021-07-27
tests: use test writer for fmt subscriber

geigerzaehler:
We use the test writer for `tracing_subscriber` so that log output is captured by the tests. This also allows us to increase the default verboseness of the log outputs which provides more information when the tests fail.

### DIFF
--- a/test/src/logging.rs
+++ b/test/src/logging.rs
@@ -28,14 +28,21 @@ use tracing_subscriber::{EnvFilter, FmtSubscriber};
 pub fn init() {
     if env_logger::builder().is_test(true).try_init().is_ok() {
         if env::var("RUST_LOG").is_err() {
-            env::set_var("RUST_LOG", "error");
+            env::set_var("RUST_LOG", "debug");
         }
 
-        let mut builder = FmtSubscriber::builder().with_env_filter(EnvFilter::from_default_env());
+        let mut builder = FmtSubscriber::builder()
+            .with_env_filter(EnvFilter::from_default_env())
+            .with_test_writer();
         if log_enabled!(target: "librad", Level::Trace) {
             builder = builder.with_thread_ids(true);
         } else if env::var("TRACING_FMT").is_err() {
-            env::set_var("TRACING_FMT", "compact")
+            let default_format = if env::var("CI").is_ok() {
+                "compact"
+            } else {
+                "pretty"
+            };
+            env::set_var("TRACING_FMT", default_format);
         }
 
         match env::var("TRACING_FMT").ok().as_deref() {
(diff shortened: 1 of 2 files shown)


kim:
Yes we _do_ use the capturing test writer. The defaults have been adjusted over time for CI, and it did not prove useful to dump debug level logs when tests fail — you’ll typically need to isolate the failures anyway, and then you can override via env variables.

So no, I don’t think this is better, because we had these settings already and dialled down the verbosity.

test/src/logging.rs
```diff
@@ -28,14 +28,16 @@ use tracing_subscriber::{EnvFilter, FmtSubscriber};
...
     if env_logger::builder().is_test(true).try_init().is_ok() {
         if env::var("RUST_LOG").is_err() {
-            env::set_var("RUST_LOG", "error");
+            env::set_var("RUST_LOG", "debug");
         }
 
-        let mut builder = FmtSubscriber::builder().with_env_filter(EnvFilter::from_default_env());
+        let mut builder = FmtSubscriber::builder()
+            .with_env_filter(EnvFilter::from_default_env())
+            .with_test_writer();
```
kim, line 36: Wait, is that the new thing? `env_logger::builder().is_test(true)` did previously have the same effect, so what's the interaction between the two?

geigerzaehler: `.is_test(true)` on `env_logger()` has no effect on the tracing subscriber. I’m still seeing error logs for test cases that pass by default. And if I change the default log level I see logs for every test case.

geigerzaehler:
> So no, I don’t think this is better, because we had these settings already and dialled down the verbosity.

Fair enough. I’ll revert the verbosity and output change.

test/src/logging.rs
kim: Oh interesting, I always assumed this would log _though_ `env_logger`. So this never worked as intended? Is the `.is_test(true)` needed at all?

kim:
I tried this out, and it does indeed make a lot more sense now. No output at all is generated if the test passes, otherwise all the buffered logs are dumped. Having a log level higher than "error" does make sense then.

I'm still skeptical of the "pretty" format on CI, because that can easily get truncated and then you're looking at ANSI escape sequences in the "raw" view. Maybe we can default to "compact" when the `CI` environment variable is set, and use "pretty" otherwise.

test/src/logging.rs
geigerzaehler: I think creating an `env_logger` is not needed at all. If the [`tracing-log`](https://docs.rs/tracing-subscriber/0.2.19/tracing_subscriber/index.html#optional-dependencies) feature on `tracing_subscriber` is enabled (which it is by default), the `fmt` subscriber will print all events emitted by `log` according to the docs.

geigerzaehler:
> Having a log level higher than "error" does make sense then.

Any preference for a specific level? I was leaning towards debug.

> Maybe we can default to "compact" when the `CI` environment variable is set, and use "pretty" otherwise.

:+1:

kim:
Debug seems good yes. We can fine tune it later if needed.

kim (dismissed):

PR #739, opened 2021-07-29
clib: introducing common CLI data types and functions

FintanH:

### DIFF
--- a/clib/Cargo.toml
+++ b/clib/Cargo.toml
@@ -1,0 +1,23 @@
+[package]
+name = "link-clib"
+version = "0.1.0"
+authors = ["The Radicle Team <dev@radicle.xyz>"]
+edition = "2018"
+license = "GPL-3.0-or-later"
+
+[lib]
+doctest = true
+test = false
+
+[dependencies]
+radicle-keystore = "0.1"
+serde_json = "1.0"
+serde = "1.0"
+thiserror = "1.0"
+
+[dependencies.librad]
+path = "../librad"
+
+[dependencies.minicbor]
+version = "0.9.1"
+features = ["std"]
--- a/clib/Cargo.toml
+++ b/clib/Cargo.toml
@@ -9,6 +9,9 @@ license = "GPL-3.0-or-later"
 doctest = true
 test = false
 
+[features]
+unsafe = []
+
 [dependencies]
 radicle-keystore = "0.1"
 serde_json = "1.0"
--- a/clib/README.md
+++ b/clib/README.md
@@ -1,0 +1,12 @@
+# CLIB
+
+When a CLI needs a LIB, one gets a CLIB.
+
+This package provides the common data types and functions that can be
+used across the different set of CLI packages.
+
+* `clib::keys` - common functions for setting up of and retrival from the secret key
+  storage.
+* `clib::ser` - serialization formats required for CLI output.
+* `clib::storage` - common functions for setting up read-only and
+  read-write storage.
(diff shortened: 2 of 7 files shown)


clib/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "clib"
```
kim, line 2: no prefix?

clib/README.md
```diff
@@ -0,0 +1,12 @@
+# CLIB
+
+When a CLI needs a LIB, one gets a CLIB.
```
kim, line 3: must.. resist.. temptation to look up in urbandictionary

kim (approved):

clib/README.md
FintanH: Unintentional, but yeh... Best not hahaha

clib/Cargo.toml
FintanH: I was on the fence since I like the shortness of `clib::` and mirroring `librad`. But I'm happy to do `radicle-clib` or `clibrad` :smile:

clib/src/keys.rs
```diff
@@ -0,0 +1,44 @@
...
+use radicle_keystore::{
+    crypto::{Crypto, KdfParams, Pwhash, SecretBoxError},
+    file,
+    pinentry::Prompt,
+    FileStorage,
+    Keystore as _,
+};
+
+/// The filename for storing the secret key.
+pub const LIBRAD_KEY_FILE: &str = "librad.key";
```
alexjg, line 20: This pops up in a number of places in different codebases, is the intention that this be it's canonical home?

clib/src/keys.rs
```diff
@@ -0,0 +1,44 @@
...
+
+pub type Error = file::Error<SecretBoxError<std::io::Error>, IntoSecretKeyError>;
+
+/// Create a [`Prompt`] for unlocking the key storage.
+pub fn prompt() -> Pwhash<Prompt<'static>> {
+    let prompt = Prompt::new("please enter your passphrase: ");
+    Pwhash::new(prompt, KdfParams::recommended())
```
alexjg, line 27: In upstream there's a feature flag to switch this to test KDF params. Do we want to have a similar thing here for when you're testing CLI applications? Especially when testing CLI applications against test Upstream sandboxes.

clib/src/ser.rs
```diff
@@ -0,0 +1,42 @@
...
+pub enum Format {
+    /// Requires the data type to implement [`Serialize`].
+    Json,
+    /// Requires the data type to implement [`Encode`].
+    Cbor,
+}
+
+impl Format {
+    /// Serialize the `val` to a `String`.
+    pub fn to_string<T>(&self, val: &T) -> Result<String, Error>
```
alexjg, line 33: This feels (to me) confusingly named because of the ubiquity of `ToString::to_string`, would `format` be better?

clib/src/keys.rs
FintanH: Ya, that would be the intended purpose. Or we move it somewhere _even more_ canonical, and this lib uses it.

clib/src/ser.rs
FintanH: You're right, I was just cargo culting with `serde_json::to_string` being the naming. I like the `format` suggestion, thanks :)

clib/Cargo.toml
kim: As I said elsewhere, I think we should just prefix every crate with `link-`, and never need to change it again

alexjg (approved):

kim (approved):

PR #740, opened 2021-07-29
Retain outgoing uni streams

kim:
See message of most recent commit.

_Note: `quinn:TransportConfig::max_concurrent_*_streams` applies to the remote end, so we can change this to a more restrictive value only after this was rolled out_

### DIFF
--- a/librad/src/net/protocol/io/send/rpc.rs
+++ b/librad/src/net/protocol/io/send/rpc.rs
@@ -3,9 +3,9 @@
 // This file is part of radicle-link, distributed under the GPLv3 with Radicle
 // Linking Exception. For full terms see the included LICENSE file.
 
-use std::net::SocketAddr;
+use std::{net::SocketAddr, ops::DerefMut as _};
 
-use futures::sink::SinkExt as _;
+use futures::{SinkExt as _, TryFutureExt as _};
 use futures_codec::FramedWrite;
 
 use crate::net::{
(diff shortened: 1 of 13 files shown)


alexjg:
How does this work with the upgrade process and the receive streams? Does it rely on the same stream always being used for the same purpose (i.e the `idx` argument to `borrow_uni` always being the same for each stream type)?

librad/src/net/protocol/io/send/rpc.rs
```diff
@@ -52,19 +52,37 @@ where
...
+            quic::BorrowUniError::Upgrade(f) => error::Rpc::Upgrade(f),
+        }
+    }
 
     match rpc.into() {
         Membership(msg) => {
-            let upgraded = upgrade::upgrade(stream, upgrade::Membership).await?;
-            FramedWrite::new(upgraded, codec::Membership::new())
+            let mut stream = conn
+                .borrow_uni(0, |s| {
```
FintanH, line 67: For reading clarity, it might be useful to `const` these, e.g.
```rust
const MEMBER_IDX: usize = 0;
const UPGRADE_IDX: usize = 1;
```

FintanH (dismissed):
Looks good, I think I get it :)

kim:
> How does this work with the upgrade process and the receive streams?

They’re unidirectional, and there is no negotiation RTT (anymore).

> Does it rely on the same stream always being used for the same purpose (i.e the idx argument to borrow_uni always being the same for each stream type)?

Yes. I looked for typed map implementations, but they weren’t very convincing. Perhaps we can turn the index into `Into<usize>`, so callers can use an enum.

alexjg (approved):

PR #741, opened 2021-08-04
git-protocol: make stateless

kim:
While working on replication v3, it becomes apparent that layering more
state handling on top of a stateful transport is getting quite
convoluted. Thus make all git protocol v2 interactions stateless, ie. a
"delegate" represents a single command.

Note that this requires to assemble the capabilities advertisement
ourselves, as `git upload-pack` cannot be taught to send it when using
`--stateless-rpc`.


kim:
/cc @Byron

You'll notice that, while it works, this is slightly shoehorned onto the stateful-by-default `git_protocol::fetch`. Specifically, I think `indicate_end_of_interaction` [here](https://github.com/Byron/gitoxide/blob/main/git-protocol/src/fetch_fn.rs#L152-L160) should be sent unconditionally to abort the `fetch` command early. It's not a big deal though, as closing the stream should have the same effect.

git-protocol/src/ls.rs
```diff
@@ -0,0 +1,149 @@
...
+    fn prepare_ls_refs(
+        &mut self,
+        caps: &client::Capabilities,
+        args: &mut Vec<BString>,
+        _: &mut Vec<(&str, Option<&str>)>,
+    ) -> io::Result<LsRefsAction> {
+        let must_namespace = must_namespace(caps);
+        for prefix in &self.opt.ref_prefixes {
+            let mut arg = BString::from("ref-prefix ");
+            if must_namespace {
```
Byron, line 83: It's interesting that namespaces are enforced on client side, I would have expected the server to always enforce the correct namespace to be set.

git-protocol/src/transport.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    fn to_url(&self) -> String {
+        self.inner.to_url()
+    }
+
+    fn supported_protocol_versions(&self) -> &[Protocol] {
+        &[Protocol::V2]
+    }
+
+    fn is_stateful(&self) -> bool {
```
Byron, line 66: I see how having a way to control 'is_stateful()` for `git::Connection` could have avoided this entire boiler plate. Maybe that should just be added now. The reason I wasn't doing it when adding `custom_url()` was that git connections are always stateful, maybe I am conflating things though as there is a persistent connection yet V2 assure stateless and command-centric operation.

I understand that `is_stateful() -> false` is present to prevent `git_protocol::fetch(…)` to end the interaction and allow a fetch interaction to follow right after using the same connection.

Byron (commented):
> You'll notice that, while it works, this is slightly shoehorned onto the stateful-by-default git_protocol::fetch. Specifically, I think indicate_end_of_interaction here should be sent unconditionally to abort the fetch command early. It's not a big deal though, as closing the stream should have the same effect.

Indeed, currently the whole 'indicate_end_of_interaction' logic always depends on statefulness of the transport, but sometimes depends on the protocol version as well. [The latter](https://github.com/Byron/gitoxide/blob/928f75ad939e35a159d3d2751d5d0f9d00d796af/git-protocol/src/fetch_fn.rs#L193-L195) is based on journey tests failing otherwise, a strong indicator of the logic being correct at least.

In the referenced code I think instead of unconditionally indicating the end of interaction, maybe the delegate should be asked instead? After all, 'ls_refs' followed by 'fetch' can't currently be done unless there is a stateless transport. This is the situation I think was encountered here.

If I didn't get it too wrong I think the delegate [should be asked here](https://github.com/Byron/gitoxide/blob/928f75ad939e35a159d3d2751d5d0f9d00d796af/git-protocol/src/fetch_fn.rs#L153) similarly how it's already done [at the end of the pack-receive operation](https://github.com/Byron/gitoxide/blob/928f75ad939e35a159d3d2751d5d0f9d00d796af/git-protocol/src/fetch_fn.rs#L193).

This would allow the custom proxy-transport to be removed and I think would be (more) correct. If you agree this could be implemented today.

git-protocol/src/ls.rs
kim: It's a bug in `upload-pack`.

git-protocol/src/transport.rs
kim: `is_stateful()` needs to return `false` for the case of executing `ls-refs`, but `Action::Cancel`ling `fetch` _iff_ the remote end is run as `--stateless-rpc`. In this case, the remote will exit after one command has been processed, which results in `EPIPE` if the client tries to write something after the command was already done.

This is essentially how an HTTP transport would work -- whether the layer 3 connection is persistent or not is opaque. You're right that a full fetch interaction is not actually stateless from the client's perspective: it needs to remember the advertised refs in order to be able to update the local tips after a packfile was received, and also the wants/ACKs for multi-round negotiation. I am seeking to abstract those two concerns from the lower-level `Delegate`, because the `Delegate` impls are otherwise the same per command, if that makes sense.

FintanH (approved):
:+1:

git-protocol/src/transport.rs
Byron: > is_stateful() needs to return false for the case of executing ls-refs, but Action::Cancelling fetch iff the remote end is run as --stateless-rpc. In this case, the remote will exit after one command has been processed, which results in EPIPE if the client tries to write something after the command was already done.

Thanks for elaborating, I think I have understood (much better) now.
Since the server side hangs up right after ls-refs due to `--stateless-rpc`, with the way it's currently implemented stateful connections would always want to send a flush message even though the other end has hung up. 

Maybe just a step of many, I think the `Transport` wrapper wouldn't be required if the delegate would also be asked if the interaction should indeed be ended. [This is now implemented](https://github.com/Byron/gitoxide/blob/65ce8e1812ce820e3a0c40e39170339bf73234e5/git-protocol/src/fetch_fn.rs#L155) and I hope it will help. After removal of the `Transport` wrapper, due to the [implementation of the delegate](https://github.com/radicle-dev/radicle-link/pull/741/files#diff-644094fb16c526094bcd4e7e097f291e53590308e20dcd31c5456f767e5bbdc0R114) a flush message should not be sent independently of the transport. Also note that I found a 'bug' in the trait implementation which caused the `indicate_client_done_when_fetch_completes()` method not to be called at all unless the `Delegate` implementation was passed in _owned_ form. It took me a moment of intense staring at the code to figure out [why](https://github.com/Byron/gitoxide/commit/65ce8e1812ce820e3a0c40e39170339bf73234e5#diff-79c081e45e3128b5199c2fbd9adc5537cb7daf5f64f5a57e4175cf7f8c7a6c14R163-R165) it wouldn't call the overridden method.

kim: Thanks. 

It remains a bit unclear under which circumstances one would implement `Delegate::indicate_client_done_when_fetch_completes` as `true` and `Transport::is_stateful` as `false` (effectively not sending any flush packets). For clarity I'm leaving the wrapper in place thus.

Byron: A valid point.

`Transport::is_stateful()` is mostly useful to know if command arguments have to be resent during negotiation (effectively making the command payload larger each round).
The other use is about terminating persistent connection gracefully, so it does appear that `is_stateful()` is a misnomer in the second case. I think the first case would also work with a changed name, like `is_persistent()`, and make more sense overall.

`Delegate::indicate_client_done_when_fetch_completes` would allow reusing such a persistent connection as no final flush packet is sent. This is required in case of V2 I think or else the server will be upset because the client hung up unexpectedly. It's true that having it return `true` while the transport is not (persistent|stateful) is ineffective, but then again the `Delegate` is not involved with the choice of transport at all. Maybe it shouldn't be deciding this either, but instead have it handled by the caller of `git_protocol::fetch()` to decide on a call-by-call basis. After all, the Delegate shouldn't care either.

Both changes seem like they would improve the situation and maybe tempt you to drop the `Transport` proxy in favor of an additional argument to `git_protocol::fetch(…, ReusePersistentConnection::No)` :).

kim: Just some thoughts: 

The [spec](https://github.com/git/git/blob/2d755dfac9aadab25c3e025b849252b8c0a61465/Documentation/technical/protocol-v2.txt#L150-L154) says (emphasis mine):

> Protocol version 2 is **stateless by default**.  This means that all commands
must only last a single round and be stateless **from the perspective of the
server side**, unless the client has requested a capability indicating that
state should be maintained by the server.  Clients MUST NOT require state
management on the server side in order to function correctly.

It is helpful to emphasise that, conversely, the **client is stateful** if and only if the protocol is used interactively. Note, however, that approx. 95% of all git requests in the wild can be performed non-interactively using v2:

```
command=fetch
0001
want-ref refs/heads/main
# optional: 
# have $(git rev-parse refs/heads/main)
done
0000
```

Another observation is that the concept of connection persistence is a historical artifact, and irrelevant in v2: as an example, there is nothing which would prevent an implementation of an SSH transport which keeps the (TCP) connection, but executes multiple commands in succession. Likewise for HTTP beyond 1.1. Only a `git-daemon` on the remote end (note: not the `git://` transport per-se) is a bit different, because the server expects a diligently well-behaving client.

So it seems to me like most of the complexity of the API `git-protocol` now presents stems from the desire to support a networked `upload-pack` written [16 years ago](https://github.com/git/git/commit/a87e8be2aece466e3f41dfde81c80709e1d56cd3) :eyes: 

I believe I said this before, but I think it would be _much_ simpler to present a command-oriented API instead of requiring users to implement callbacks for all possible branching points. For example:

```rust
trait Transport {
    async fn request<C: Command>(&self, cmd: C) -> Result<C::Response, Error>;
    async fn close(self);
}

trait BuildTransport {
    type Transport: Transport;

    async fn connect() -> Result<(Self::Transport, Option<LsRefsResponse>), Error>;
}

impl Client<T: Transport> {
    pub fn new(trans: T, refs: Option<LsRefsResponse>) -> Result<Self, Error> { /**/ }

    pub async fn ls_refs(&self, cmd: LsRefs) -> Result<LsRefsResponse, Error> { /**/ }
    pub async fn fetch(&self, cmd: Fetch) -> Result<FetchResponse<impl AsyncBufRead + Unpin>, Error> { /**/ }
}

enum FetchResponse<P> {
    Negotiating(Response),
    HavePack {
        resp: Response,
        pack: P
    }
}
```

To fall back to v1, the `LsRefsResponse` can be memoized on `connect`. To perform multi-round negotiation, it is up to the caller to call `fetch` repeatedly. It is up to the `Transport` impl to decide on each command whether it reuses a connection or creates a new one (or some kind of lightweight stream). Accordingly, it sends flush packets or doesn't. While `Transport` is aware of the protocol, implementations for any `AsyncRead` / `AsyncWrite` can be provided via `Stateless`, `Persistent` newtypes.

Byron: What a write-up! Thanks a lot for the analysis, I would definitely love to improve on what's currently presented in `git-transport` and/or `git-protocol` as no matter how hard one tries, certain abstractions don't seems to fit in that well. 

Particularly the notion of statefulness and even persistency of connections across requests come to mind here. Since the current complexity stems from supporting all connection types along with protocol V1/V2 I admittedly can't tell/simulate if changing it into the model above will cover all cases sufficiently well. Probably you are also suggesting there is not need to support the old world anyway, and in case of V1 I couldn't agree more. When it comes to `git-daemon` then at least it has proven very useful for testing 😅 .

My intuition about this is as follows: The current complexity supports everything at the costs mentioned, hence it should support simpler abstractions on top like the one proposed here. With `git-repository` being in a position where it is supposed to make certain commands easy to do without sacrificing much control or performance I think it would be the one to either provide such abstractions or implement them under the hood. Ideally, `git-repository` would support custom transports while making requests as simple as above.

Right now one has to use these plumbing-level crates for everything which is much more cumbersome than the facilities provided by `git2` for example, but with `git-repository` I expect APIs that are en-par (or nicer) than that.

I will keep this conversation in mind when implementing `gixp fetch` and see if it's a good time to push protocol related functionality into `git-repository`, opening an avenue to improving this.

PR #749, opened 2021-08-11
git-protocol: switch to gitoxide deps from crates.io

kim:
The functionality used here was tagged and released yesterday, so we do
not have a reason to use git dependencies anymore.


FintanH (approved):
:kissing_closed_eyes::ok_hand:

alexjg (approved):

PR #751, opened 2021-08-16
daemon: don’t persist waiting room on tick

geigerzaehler:
We stop persisting the waiting room on the `Tick` input. The waiting room state itself does not change when this input is processed. Since the waiting room tick occurs every 500ms by default the serialization load might be heavy.

Similarly, we eliminate the `Tick` waiting room event since it offer no valuable information.

### DIFF
--- a/daemon/src/peer/run_state/running_waiting_room.rs
+++ b/daemon/src/peer/run_state/running_waiting_room.rs
@@ -80,8 +80,6 @@ pub enum Event {
         /// The attempts that were made before the timeout
         attempts: Option<usize>,
     },
-    /// One tick of the waiting room
-    Tick,
 }
 
 /// `RunningWaitingRoom` is an adapter from the interface of `WaitingRoom` to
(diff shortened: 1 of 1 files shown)


daemon/src/peer/run_state/running_waiting_room.rs
```diff
@@ -192,27 +190,18 @@ impl RunningWaitingRoom {
...
-        let state_before = self.waiting_room.requests();
+        let mut cmds = Vec::with_capacity(2);
 
         if let Some(urn) = self.waiting_room.next_query(timestamp) {
             cmds.push(Command::Request(command::Request::Query(urn)));
-            cmds.push(Command::PersistWaitingRoom(self.waiting_room.clone()));
         }
         if let Some((urn, remote_peer)) = self.waiting_room.next_clone() {
             cmds.push(Command::Request(command::Request::Clone(urn, remote_peer)));
-            cmds.push(Command::PersistWaitingRoom(self.waiting_room.clone()));
```
FintanH, line 205: It's been a minute since I looked at this code, but perhaps you know the answer since you've dug in a bit. At what point does the waiting room get persisted now that we remove these?

geigerzaehler: The waiting room is persisted every time a request changes it’s state. In most cases this happens through `simple_command_helper`

https://github.com/radicle-dev/radicle-link/blob/90a024f35c86fb21fd8866c1f85848da48cf869d/daemon/src/peer/run_state/running_waiting_room.rs#L299

or in the transition function directly

https://github.com/radicle-dev/radicle-link/blob/90a024f35c86fb21fd8866c1f85848da48cf869d/daemon/src/peer/run_state/running_waiting_room.rs#L123

FintanH (dismissed):

daemon/src/peer/run_state/running_waiting_room.rs
alexjg: One risk of removing the persist on tick is that it becomes easier to fail to save the waiting room state as the logic is more complex. Is it worth adding a few tests to assert that persist commands are emitted in the various paths through the waiting room logic?

geigerzaehler: >  Is it worth adding a few tests to assert that persist commands are emitted in the various paths through the waiting room logic?

I let you make that call. I’m not opposed to this but to me the code should enforce this which would require some refactoring. Even if we add some tests, it is not guaranteed that persistence is ensured when more code paths are added.

geigerzaehler: @alexjg Could you have another look at this?

alexjg: I'm happy to leave the testing for now I think, the impact of a failed persistence is fairly low and as you say, some refactoring is probably required to test this in a satisfactory manner.

FintanH (approved):

PR #754, opened 2021-08-17
refactor: extract crypto-related modules into crate

kim:
In preparation for a cleaner separation of the replication v3, the
pervasive PeerId type is made available as a standalone dependency
`link-crypto`.  Since it is merely a newtype around a public key, all
other
crypto-related modules (`keys`, `signer`) are bundled as well.

Because virtually every source file needs to be touched for this change,
it tries to be as mechanical as possible, and doesn't introduce any
other changes apart from moving files and adjusting imports.
Specifically, no attempt is made to keep crate naming consistent
(`link-` instead of `radicle-` or `radicle-link-` prefix). This would
need to be addressed in a later patch.


alexjg (dismissed):

kim:
Edit: slim down direct dependencies of `librad`

kim:
Edit: fix broken doc links

FintanH (approved):
:+1: :ok_hand: :call_me_hand:

PR #760, opened 2021-08-18
Once popped, never stopped

kim:
This is a seven-part series to extract `librad` modules into top-level crates in
order to improve modularity. The following approach is followed:

* crates are still re-exported through `librad`, preserving the module hierarchy
  where possible / sensible
* the module hierarchy in the `test` crates is left untouched, ie. all tests
  happen through `librad`
* the move is motivated by allowing the v3 replication to live in it's own crate
  (and thus be able to pick smaller dependencies). We don't go beyond that for
  now, namely the big `librad` module hierarchies `net` and `git` are left
  intact. This is because splitting them would cause even more churn, and it is
  not immediately obvious how to best split them.
* to avoid naming conflicts, protocol related crates are prefixed with `link-`
  (incl. their directories), while crates which could also be released
  standalone don't have a prefix. No attempt is made in this series to adjust
  the names of already existing crates, although that might be submitted shortly
  after


link-canonical/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "link-canonical"
+version = "0.1.0"
+authors = ["Kim Altintop <kim@eagain.st>"]
```
FintanH, line 4: Is there a reason we're not using "The Radicle Team" out of interest?

kim: Because it did not show up in `git log` :shrug:

FintanH (dismissed):
Nice, this looks good :) I, for one, welcome lots of smaller crates :+1:

link-canonical/Cargo.toml
kim: The cargo reference for [`[authors]`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-authors-field) says: 

> The exact meaning is open to interpretation — it may list the original or primary authors, current maintainers, or owners of the package.

My personal interpretation is that maintainers are practically the same as package owners (on crates.io), and the `authors` field is never actually rendered anywhere (at least not that I'm aware of). We also use "The Radicle Contributors" in the license header, so as to not having to update it whenever some substantial modification occurs (although one would be free to do so). Additionally, "dev@radicle.xyz" is not actually a contact address.

So, "original or primary authors" seems sensible to me.

librad/src/lib.rs
```diff
@@ -25,14 +25,15 @@ extern crate lazy_static;
 #[macro_use]
 extern crate radicle_macros;
 
+pub extern crate link_canonical as canonical;
```
kim, line 28: Actually, `pub use` would be equivalent when there's no `#[macro_use]`. Not sure, it doesn't seem like `extern crate` would be deprecated from the language anytime soon.

link-canonical/Cargo.toml
FintanH: Ya, sounds good to me too :+1: I was just curious :)

Issue #734, opened 2021-07-26
Seed is flooding peers with new ingress streams


geigerzaehler:
When connecting to either `setzling` or `sprout` we see the following log line every 200ms
```
Jul 26 12:15:18.816  INFO accept{peer_id=hyb79bxwfz8mbmqarrok53fj9cnmmey6oujg1yfoysjjo1rmzojuhw}:
  disco:discovered{remote_id=hynkyndc6w3p8urucakobzna7sxwgcqny7xxtw88dtx3pkf7m3nrzc}:
  incoming{remote_id=hynkyndc6w3p8urucakobzna7sxwgcqny7xxtw88dtx3pkf7m3nrzc remote_addr=34.141.175.140:12345}:
  librad::net::protocol::io::streams: new ingress stream
```

In addition, we also see the following log line for a while until it goes away (maybe after replicating a first project)
```
Jul 26 12:07:56.240  WARN radicle_daemon::peer::run_state::running_waiting_room:
  waiting room error error=MissingUrn(Urn { id: Oid(1183cf230ec31058fa955e69b4172086161641de), path: Some(RefLike("rad/id")) })
```

Steps to reproduce

1. Start the most recent version (0.2.9 or master) of Upstream from the command line with a fresh profile using `RAD_HOME`
2. Create a profile
3. Log at the command line output

kim:
Is it the same `remote_addr` always?

geigerzaehler:
> Is it the same `remote_addr` always?

Yes it is. Once I’m connected to more peers this happens with every peer. Also the Oid for the waiting room warning is always the same.

kim:
I don’t know of anything which would trigger packets at a 200ms interval, except perhaps an amplification effect of something malfunctioning with the “waiting room”

xla:
@geigerzaehler If you could conjure up debug logs from seed that correspond to the upstream peer that would help with narrowing down the interaction that is causing this.

cloudhead:
> I’ve noticed that once my peer is connected to multiple peers I’m seeing this log line for every peer. This puts significant pressure on the CPU resulting in 50% utilization. This is a high severity issue for Upstream users since the app and their machines become unresponsive

Does this then rule out a potential issue in the seed node/library?

geigerzaehler:
> @geigerzaehler If you could conjure up debug logs from seed that correspond to the upstream peer that would help with narrowing down the interaction that is causing this.

[...]

PR #762, opened 2021-08-19
seed: consider if tracking relationship was already established

kim:
When calling `Node::track_project`, it might very well be that the
project/peer is already tracked, which would result in unnecessary
broadcast traffic. Thus, don't emit any events in this case.

Also, change the order of `track` and `replicate` to ensure we
do actually fetch the newly-tracked peer's view. Additionally, we can
skip `replicate` if we already tracked the peer, as subsequent broadcast
messages won't be considered uninteresting.

Ref #734

### DIFF
--- a/seed/src/lib.rs
+++ b/seed/src/lib.rs
@@ -399,7 +399,7 @@ impl Node {
 
                     if mode.is_trackable(peer_id, urn) {
                         // Attempt to track, but keep going if it fails.
-                        if Node::track_project(api, urn, provider).await.is_ok() {
+                        if let Ok(true) = Node::track_project(api, urn, provider).await {
                             let event = Event::project_tracked(urn.clone(), *peer_id, api).await?;
                             api.announce(Payload {
                                 urn: urn.clone(),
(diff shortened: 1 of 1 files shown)


seed/src/lib.rs
```diff
@@ -440,18 +443,25 @@ impl Node {
...
             .await?
         };
 
         match &result {
-            Ok(()) => {
+            Ok(true) => {
                 tracing::info!("Successfully tracked project {} from peer {}", urn, peer_id);
             },
+            Ok(false) => {
+                tracing::info!("Project {} from peer {} already tracked", urn, peer_id);
```
kim, line 463: Not sure, maybe remain silent?

FintanH: Or `debug` if we want to look into this specific call path for some reason

FintanH (approved):
Nice, that makes sense!

cloudhead (dismissed):
Yes!

kim:
Please pull out your truth tables and verify I got the booleans right

PR #763, opened 2021-08-19
git-protocol: rename crate to `link-git-protocol`

kim:
The mouthfulness of `radicle-link-git-protocol` was actually the
motivation for dropping the `radicle-` prefix in #763 and #760, so let's
do this now.


FintanH (dismissed):
Alles klar :+1:

alexjg (dismissed):

PR #764, opened 2021-08-19
refactor: part 8/7

FintanH:
librad re-exports link-crypto, which in turn exports
radicle-keystore. This can lead to a slew of compiler errors if the
radicle-keystore dependency changes for one of the other crates, e.g.
temporarily developing on top of a git SHA.

This patch removes the dependency on radicle-keystore from the rest of
the crates and imports the keystore from librad instead.


kim (approved):
o7

alexjg (approved):

PR #766, opened 2021-08-20
clib: ssh-agent storage

FintanH:
Depends on https://github.com/radicle-dev/radicle-keystore/pull/21

### DIFF
--- a/clib/src/storage.rs
+++ b/clib/src/storage.rs
@@ -6,6 +6,7 @@
 use thiserror::Error;
 
 use librad::{
+    crypto::BoxedSigner,
     git::storage::{error, read, ReadOnly, Storage},
     profile::Profile,
 };
@@ -35,10 +36,10 @@ pub mod prompt {
     ///
     /// The decryption will happen by prompting the person for their passphrase
     /// at the command line.
-    pub fn storage(profile: &Profile) -> Result<Storage, Error> {
+    pub fn storage(profile: &Profile) -> Result<(BoxedSigner, Storage), Error> {
         let paths = profile.paths();
         let signer = keys::signer_prompt(profile)?;
-        Ok(Storage::open(paths, signer)?)
+        Ok((signer.clone(), Storage::open(paths, signer)?))
     }
 }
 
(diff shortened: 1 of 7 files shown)


alexjg (dismissed):

clib/src/storage.rs
```diff
@@ -22,28 +22,41 @@ pub enum Error {
...
+
+pub mod ssh {
+    use thrussh_agent::client::ClientStream;
+
+    use super::*;
+
+    /// Initialise [`Storage`].
+    ///
+    /// The signing key will be retrieved from the ssh-agent. If the key was not
+    /// added to the agent then this result in an error.
```
alexjg, line 53: Is the intention that clients will pattern match on the `SshConnect` error to detect whether the key was not added to the agent in order to display appropriate error messages to the user? If so it might be worth adding some documentation of how exactly to do that here?

FintanH: I actually had no intention of that :sweat_smile: I'm actually just assuming it's going to result in an error -- because what else would it do? :grimacing: Once I do some playing around with it I can be more opinionated about possibly handling the error :)

alexjg: Right, I guess the `SshAgent` could list identities to ensure that the key is actually present in the agent. Although that only ensures that the key is present when you create the signer and it could be removed later due to a lifetime constraint so it probably doesn't buy much. I guess you also could do that on each `Signer::sign` call to produce nicer error messages at the expense of  an additional round trip to the agent.

But yeah, I can also see how this will probably become clearer once you've started playing with it.

kim: From playing around with `thrussh-keys`, the error messages are actually terrible and basically contain no information about what might have gone wrong. It is possible that this is just how `ssh-agent` works.

FintanH: Ya, it's probably a nice idea to add something to the `rad-profile` set of commands to see if your key is in the list, ala `rad profile ssh-ask` :)

alexjg: The protocol RFC has a single generic error response so I imagine there's not much extra `thrussh-keys`can do. Although I think it would make sense to add some utility methods like "check if a key exists before requesting signing with it" which can do as I mentioned above and list identities first in order to provide nice error messages.

FintanH: I can make a note of this in an issue so that I don't forget when we do start playing around with it

FintanH: Captured this here: https://github.com/radicle-dev/radicle-link/issues/767

kim (dismissed):
lgtm

PR #770, opened 2021-08-24
daemon: remove sync

FintanH:
With rere and eventually grafting, we will not need the sync mechanism.

Currently we have been deploying seeds with rere disabled, but this
makes upgrading less maintainable as we need to keep track of an extra
branch. We did this because of some emergent effects between rere and
sync, so this also allows us to move forward with rere.


geigerzaehler (dismissed):

xla (dismissed):
🍙 🔺 💂 🎎

kim:
Hm interesting. Is that the code path which could've been altered as [proposed here](https://github.com/radicle-dev/radicle-link/issues/734#issuecomment-901725862)?

FintanH:
> Hm interesting. Is that the code path which could've been altered as proposed here?

Well, there's also this shouting match https://github.com/radicle-dev/radicle-link/blob/master/daemon/src/peer/announcement.rs#L133-L152 which announces new commits.

I wasn't around when the rere and sync shenanigans were happening, I believe. Maybe @xla could shed some light there if he remembers.

PR #771, opened 2021-08-25
fix: waiting room handles gossip only if it knows URN

geigerzaehler:
We only send information about gossip messages to the waiting room if the waiting room is waiting for them.

### DIFF
--- a/daemon/src/peer/run_state.rs
+++ b/daemon/src/peer/run_state.rs
@@ -281,7 +281,9 @@ impl RunState {
                         provider: PeerInfo { peer_id, .. },
                         result,
                     } => {
-                        cmds.extend(self.waiting_room.found(&urn, peer_id, SystemTime::now()));
+                        if self.waiting_room.get(&urn).is_some() {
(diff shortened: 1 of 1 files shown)


FintanH (approved):

daemon/src/peer/run_state.rs
```diff
@@ -317,7 +317,9 @@ impl RunState {
                         provider: PeerInfo { peer_id, .. },
                         result,
                     } => {
-                        cmds.extend(self.waiting_room.found(&urn, peer_id, SystemTime::now()));
+                        if self.waiting_room.get(&urn).is_some() {
+                            cmds.extend(self.waiting_room.found(&urn, peer_id, SystemTime::now()));
+                        }
 
                         if let PutResult::Applied(_) = result {
```
kim, line 324: Can it happen that the urn is not in the waiting room, but there is still an include file which may need updating?

FintanH: Once you've replicated a project, the include will still need updating. Say if the delegates change and thus the remotes will need changing.

kim (approved):

PR #775, opened 2021-08-27
rad-exe: top-level rad CLI

FintanH:
This is the first draft of the top-level rad CLI library -- to become the
`rad` executable.

In its current form, it defines the global arguments and performs the
sanitisation of those arguments, as per the RFC. It has mock `profile`
subcommand that does nothing. It allows for an external `rad-<command>`
to be called spawning a subprocess to execute it.


rad-exe/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+edition = "2018"
+license = "GPL-3.0-or-later"
+
+[lib]
+doctest = true
+test = false
+
+[dependencies]
+anyhow = "1.0"
+structopt = "0.3"
```
xla, line 14: clueless

xla: clap

rad-exe/src/cli/args.rs
```diff
@@ -0,0 +1,130 @@
...
+/// The rules are summarised as:
+///   * The `rad` global _always_ takes precedence, e.g. `rad --rad-profile deaf
+///     xxx --rad-profile beef` will result in `deaf`
+///   * If multiple `rad` globals are given, it is an error, i.e. `rad
+///     --rad-profile deaf --rad-profile beef`
+///   * If multiple globals are given as part of the external command, the last
+///     one is take, e.g. `rad xxx --rad-profile deaf --rad-profile beef` will
+///     result in `beef`.
+///   * Command line arguments take precedence over environment variables, e.g.
+///     `RAD_PROFILE=deaf rad --rad-profile def` will result in `beef`
```
xla, line 60: Should the rules be extended to also allude to the precedence of commands themselves not only the options/params?

rad-exe/src/cli/args.rs
```diff
@@ -0,0 +1,130 @@
...
+/// rad --rad-profile=deaf --rad-profile=beef xxx # beef
+/// rad xxx --rad-profile=deaf --rad-profile=beef # beef
+/// rad --rad-profile=dead xxx --rad-profile=deaf --rad-profile=beef # dead
+/// ```
+pub fn sanitise_globals(mut args: Args) -> Args {
+    match &mut args.command {
+        Command::External(external) => {
+            sanitise_option(
+                RAD_PROFILE_ARG,
+                "RAD_PROFILE",
```
xla, line 76: Could be a `const`.

*ducks for cover*

xla: And/or bundle this one and the other `RAD_`s up in an env module.

rad-exe/src/cli/main.rs
```diff
@@ -0,0 +1,48 @@
...
+    process::{exit, Command},
+};
+
+use structopt::StructOpt;
+
+use super::args::{self, sanitise_globals, Args};
+
+pub fn main() -> anyhow::Result<()> {
+    let args = sanitise_globals(Args::from_args());
+    println!("{:?}", args);
```
xla, line 17: Is this debug leftover?

xla (commented):
The structure so far makes sense to me.

rad-exe/Cargo.toml
FintanH: With an endorsement from Morgan Freeman, what could go wrong?

rad-exe/src/cli/args.rs
FintanH: Can you provide an example of what you mean?

rad-exe/src/cli/args.rs
FintanH: It's already defined here https://github.com/radicle-dev/radicle-link/blob/master/librad/src/profile.rs#L18 so I can make that `pub` :)

rad-exe/src/cli/main.rs
FintanH: Whoops :flushed:

rad-exe/src/cli/args.rs
xla: `pub` it all.

rad-exe/src/cli/args.rs
xla: Sorry, the initial comment was highly confusing. What I meant is that the rules state that globals take precedence and I wanted to check if the rules should also state that global/builtin commands take precedence.

FintanH: Ah, well first off, that's not handled by this function, it's handled by the `Command` struct. Secondly, that's stated in the RFC where we state that one shouldn't (in fact can't) override the behaviour of a core command. I say can't because the arg parser will match something like `rad profile` and execute the core command. Whereas it'll fall through if it's `rad xxx`.

Does that make sense?

xla: Tippi toppi.

kim:
> So the error that comes up is No key found 

I can't reproduce locally. Do you have a test run log with `RUST_LOG=debug`?

test/src/test/integration/git_helpers/remote.rs
```diff
@@ -24,6 +24,8 @@ use crate::{logging, rad::identities::create_test_project};
 
 const PASSPHRASE: &str = "123";
 
+// NOTE(finto): this test is running into issues with the key going missing part-way through the test (see https://github.com/radicle-dev/radicle-link/pull/775#issuecomment-907016375 for more commentary). I'm disabling it since we aim to have a git daemon server to replace this and the issue SHOULD NOT show up in real use scenarios, since I believe this is due to high levels of concurrency in the test suite.
```
xla, line 27: Maybe some newlines would help this comment.

FintanH: Yaw, I wonder why `cargo fmt` didn't do this for me :thinking:

FintanH:
> I can't reproduce locally. 

Are you running the whole test suite, i.e. `cargo test`? If you run the test individually it will work fine, or even just a subset of the tests. 

> Do you have a test run log with RUST_LOG=debug?

Here's the output from the logs when it fails https://gist.github.com/FintanH/8eb6de5e47c5038f26931ebbd5cddde6

kim:
>    Are you running the whole test suite, i.e. cargo test?

Yes

>    Here's the output from the logs when it fails
>    [1]https://gist.github.com/FintanH/8eb6de5e47c5038f26931ebbd5cddde6

Mkay, that's not too helpful. Perhaps add some GIT_TRACE_* vars?

FintanH:
> Yes

Huh... Doesn't Work On My Machine:tm:

> Mkay, that's not too helpful. Perhaps add some GIT_TRACE_* vars?

I updated the gist with the output trace outputs. I don't see anything suspicious :eyes:

kim:
Something is not right with the profile selection logic:

https://buildkite.com/monadic/radicle-link-seedling/builds/3300#401f270a-f190-4ac9-82d8-7a51890d740c

The key file the helper looks for is `/tmp/.tmpPaeAgS/ghi/keys/librad.key`, but the one the test thinks it has created is `/tmp/.tmpPaeAgS/461d13f9-ba32-4617-b6c1-7477caf7eee2/keys`

FintanH:
> The key file the helper looks for is /tmp/.tmpPaeAgS/ghi/keys/librad.key, but the one the test thinks it has created is /tmp/.tmpPaeAgS/461d13f9-ba32-4617-b6c1-7477caf7eee2/keys

Whoa! Where did that specific file path come from :o Was trying to get that to print before :sweat_smile: Thanks for the spot here, I'll look into this. Weird that it only happens on occasion :thinking:

kim:
Afaiu `ghi` is not actually a valid profile identifier. Perhaps some environmental variable gets overwritten, and is not validated?

FintanH:
> Afaiu `ghi` is not actually a valid profile identifier.

Any `String` is a valid profile identifier, but the [default ones generated are UUIDs](https://github.com/radicle-dev/radicle-link/blob/master/librad/src/profile/id.rs#L106-L109). 

> Perhaps some environmental variable gets overwritten, and is not validated?

This sounds like a good hunch :detective:

test/src/test/unit/rad_exe/args.rs
```diff
@@ -0,0 +1,169 @@
...
+    let args = sanitise_globals(args);
+    if let Command::External(external) = args.command {
+        let index = find_arg(RAD_PROFILE_ARG, &external);
+        assert_eq!("def", external[index.unwrap() + 1]);
+    }
+}
+
+#[test]
+fn rad_profile_second_precedence_multiple() {
+    env::set_var("RAD_PROFILE", "ghi");
```
FintanH, line 78: I might have found the culprit @kim :facepalm:

FintanH:
To summarise the issue, the latest tests for sanitising rad arguments involved setting the `RAD_PROFILE` environment variable. This would result in the argument being read during the test of the git remote helper (yay concurrency!)

The fix I've provided here is to use `rusty-fork`[[0]] which spawns those test cases into a subprocess, thus not interfering with the environment variables in the rest of the tests.

[0]: https://docs.rs/rusty-fork/0.3.0/rusty_fork/

kim:
Seems sensible.

kim:
You might be interested in adapting 9d51e6c06ff81b281c44890030237685cdf15928

The error renders as:

```
Error: key file: /tmp/.tmpPaeAgS/ghi/keys/librad.key

Caused by:
    No key found
```

so perhaps it needs a different copy.

rad-exe/Cargo.toml
alexjg: What's the reasoning behind using structopt rather than clap (which includes structopt these days)?

FintanH: The plan would be to use `clap-3.0.0` once it's out of its beta phase. Would you prefer to rely on `clap-3.0.0-beta.4`?

alexjg: Ah, yeah I was misled by the documentation on clap.rs into thinking that clap 3.0 was out of beta. Structopt makes sense then.

FintanH: Sweet :grin:

FintanH:
> so perhaps it needs a different copy.

How about changing the error variant here[[0]] to include the path?

[0]: https://github.com/radicle-dev/radicle-keystore/blob/master/src/file.rs#L93

kim:
Yeah sure, that works, too

xla (dismissed):
💅 😅 🌬 🙉

cloudhead (approved):

PR #777, opened 2021-08-27
git: ignore error due to concurrent sigrefs update

### LINKED ISSUE #761: when updating signed refs: failed to create commit: current tip is not the first parent
We sometimes see the following error 

```plain
ERROR 
  accept{peer_id=hyy3x75yx1ypiaz4pi1kiyxr7pa6fstwyhuxkj7qgz4teiwychzy8g}:
  incoming:incoming{remote_id=hyywnprcxyfw56jgatnjiqznq59z8kohjoimuqe57m6iooh7kfgzyg remote_addr=127.0.0.1:17000}:
  rere{urn=rad:git:hnrkjbbwjhfam4kfkjhtzfyiuyn4teibf6soo remote_peer=hyywnprcxyfw56jgatnjiqznq59z8kohjoimuqe57m6iooh7kfgzyg}:
  rere{urn=Urn { id: Oid(908689e170bd28aa4f237282b300b5145425f5a1), path: None } remote_peer=hyywnprcxyfw56jgatnjiqznq59z8kohjoimuqe57m6iooh7kfgzyg}:
  replicate{config=Config { fetch_limit: Limit { peek: 5120000, data: 5368709120 } }}:
    librad::git::replication::project: error=failed to create commit: current tip is not the first parent; class=Object (11); code=Modified (-15)
```
In the 25ms before this error we are seeing the following log statement three imtes
```plain
DEBUG 
  accept{peer_id=hyy3x75yx1ypiaz4pi1kiyxr7pa6fstwyhuxkj7qgz4teiwychzy8g}:
  incoming:incoming{remote_id=hyywnprcxyfw56jgatnjiqznq59z8kohjoimuqe57m6iooh7kfgzyg remote_addr=127.0.0.1:17000}:
  rere{urn=rad:git:hnrkjbbwjhfam4kfkjhtzfyiuyn4teibf6soo remote_peer=hyywnprcxyfw56jgatnjiqznq59z8kohjoimuqe57m6iooh7kfgzyg}:
  rere{urn=Urn { id: Oid(908689e170bd28aa4f237282b300b5145425f5a1), path: None } remote_peer=hyywnprcxyfw56jgatnjiqznq59z8kohjoimuqe57m6iooh7kfgzyg}:
  replicate{config=Config { fetch_limit: Limit { peek: 5120000, data: 5368709120 } }}:
  update{urn=rad:git:hnrkjbbwjhfam4kfkjhtzfyiuyn4teibf6soo}:
    librad::git::refs: updating signed refs for refs/namespaces/hnrkjbbwjhfam4kfkjhtzfyiuyn4teibf6soo/refs/rad/signed_refs
```

This seems to indicate that there is a race when updating the singed refs concurrently.

There seems to originate from this call. https://github.com/radicle-dev/radicle-link/blob/fcc3a7c11934fc49a433390a9b52e02fa59c4307/librad/src/git/refs.rs#L366-L373

After doing some investigation I found that the error occurs if `parent` is not the tip of `branch` (`../rad/signed_refs` in this case). `parent` is obtained a couple of lines before the call above so there is time for `branch` to get updated in the meantime.

kim:
When two `Refs::update` race, only one of them can succeed, because the
ref update only succeeds if the expected parent commit is what the ref
currently points to (ref updates are atomic). While there is a chance
that the winning update commits a slightly outdated view of the refs
tree, this is tolerable, because every modification of the namespace
will trigger a sigrefs update, and thus the state converges.

Not so great, however, is to propagate the commit failure due to a
modified parent as an error, as this may interrupt the replication flow,
leaving the namespace in a potentially half-arsed state.

Therefore, teach `Refs::update` to ignore this kind of error, and return
the most recent committed state instead.

Fixes #761

### DIFF
--- a/librad/src/git/local/transport.rs
+++ b/librad/src/git/local/transport.rs
@@ -39,6 +39,9 @@ pub enum Error {
     #[error("no rad/self present and no default identity configured")]
     NoLocalIdentity,
 
+    #[error("refs signature updated by another process")]
+    SigrefsRace,
+
     #[error(transparent)]
     OpenStorage(#[from] OpenStorageError),
 
@@ -244,7 +247,9 @@ impl LocalTransport {
                         let storage = _dyn.as_ref();
 
                         // Update `rad/signed_refs`
-                        Refs::update(storage, &urn)?;
+                        if let refs::Updated::ConcurrentlyModified = Refs::update(storage, &urn)? {
+                            return Err(Error::SigrefsRace);
+                        }
 
                         // Ensure we have a `rad/self`
                         let local_id = identities::local::load(storage, urn.clone())
--- a/librad/src/git/refs.rs
+++ b/librad/src/git/refs.rs
@@ -197,6 +197,23 @@ pub mod stored {
     }
 }
 
+/// Success result of [`Refs::update`]
+pub enum Updated {
+    /// The computed [`Refs`] were stored as a new commit.
+    Updated { refs: Refs, at: git2::Oid },
+    /// The stored [`Refs`] were the same as the computed ones, so no new commit
+    /// was created.
+    Unchanged { refs: Refs, at: git2::Oid },
+    /// Another process committed [`Refs`], so the computed ones were discarded.
+    ///
+    /// This should typically be treated as a warning for interactive updates.
+    /// See [0], [1] for further discussion.
+    ///
+    /// [0]: https://github.com/radicle-dev/radicle-link/pull/777
+    /// [1]: https://lists.sr.ht/~radicle-link/dev/%3C20210830224202.GE10879%40schmidt.localdomain%3E
+    ConcurrentlyModified,
+}
+
 /// The published state of a local repository.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Refs {
@@ -320,14 +337,8 @@ impl Refs {
 
     /// Compute the current [`Refs`], sign them, and store them at the
     /// `rad/signed_refs` branch of [`Urn`].
-    ///
-    /// If the result of [`Self::compute`] is the same as the alread-stored
-    /// [`Refs`], no commit is made and `None` is returned. If a concurrent
-    /// [`Self::update`] commits first, the result of [`Self::load`] is
-    /// returned. Otherwise, the new and persisted [`Refs`] are returned in a
-    /// `Some`.
     #[tracing::instrument(skip(storage, urn), fields(urn = %urn))]
-    pub fn update(storage: &Storage, urn: &Urn) -> Result<Option<Self>, stored::Error> {
+    pub fn update(storage: &Storage, urn: &Urn) -> Result<Updated, stored::Error> {
         let branch = Reference::rad_signed_refs(Namespace::from(urn), None);
         tracing::debug!("updating signed refs for {}", branch);
 
@@ -354,8 +365,10 @@ impl Refs {
 
         if let Some(ref parent) = parent {
             if parent.tree()?.id() == tree.id() {
-                tracing::debug!("signed refs already up-to-date");
-                return Ok(None);
+                return Ok(Updated::Unchanged {
+                    refs: signed_refs.refs,
+                    at: parent.id(),
+                });
             }
         }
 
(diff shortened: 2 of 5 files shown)


alexjg (dismissed):

kim:
Under fair^H intended use this should not be necessary: pushing to the monorepo is interactive (and single-user normally), so you get a warning printed to your terminal if the signed refs update failed. A fetch (unattended) is supposed to be exclusive per namespace, so that cannot race either.

kim:
Btw can you see how this whole thread shows how the categories “issue”, “pull request”, “discussion” are useless? I propose the terminology “project management theatre”.

FintanH (approved):

geigerzaehler:
> Under fair^H intended use this should not be necessary: pushing to the monorepo is interactive (and single-user normally), so you get a warning printed to your terminal if the signed refs update failed. A fetch (unattended) is supposed to be exclusive per namespace, so that cannot race either.

In our test case the error originates from the daemon that does a fetch and not from `git push`. As you pointed out, the race is probably caused by a concurrent `git push` but we don’t see the warning there. This means we do end up in a situation where the daemon does not publish it’s latest state until something updates the refs. This may lead to confusing behavior for any users and is also less than ideal for our test setups.

I understand that this is a stop-gap fix but I would prefer to raise the log level to a warning so that we know something’s amiss when the tests don’t work.

kim:
> I understand that this is a stop-gap fix

No it's not. What made you think so?

geigerzaehler:
> > I understand that this is a stop-gap fix
> 
> No it's not. What made you think so?

Shouldn’t we try to achieve a situation where the signed refs represent the latest state without requiring something else to force an update? This change doesn’t accomplish that.

kim:
@geigerzaehler I think you're not understanding https://github.com/radicle-dev/radicle-link/pull/777#issuecomment-907598021: the test situation is not the same as the usage situation. Whether or not the background replication fails to update the signed refs is irrelevant to the user -- the user is supposed to use `git push`. Like any other `git push`, this can fail because of concurrent modification, the only difference is that we can not currently roll back the entire push. If upstream modifies the monorepo directly, that's your problem.

kim:
> In our test case the error originates from the daemon that does a fetch and not from git push. 

Which would mean that the push won.

> This means we do end up in a situation where the daemon does not publish it’s latest state until something updates the refs.

As reported in https://github.com/radicle-dev/radicle-link/issues/761#issuecomment-908399619, our global lock did not actually have the desired effect, so it could very well be that this is due to competing fetches.

When there are competing updates, I am not aware of a way to guarantee that the most recent one eventually wins -- the background process can attempt to retry, but that's not the same as a guarantee. With the fix in place, the only competition is between the background fetcher and a user-initiated push. If the push wins, there should eventually be a new fetcher, because of gossip, rere, etc. If the push loses, the user will receive a warning, and is supposed to retry. If programmatic pushes in test code don't handle this, they need to be fixed.

geigerzaehler:
> If the push wins, there should eventually be a new fetcher, because of gossip, rere, etc. If the push loses, the user will receive a warning, and is supposed to retry.

I was missing that. Together with fixing the global lock this should be sufficient to ensure proper publication or inform the user a push failed.

geigerzaehler (approved):

FintanH (dismissed):

kim:
> If the push loses, the user will receive a warning, and is supposed to retry.

Wait no, we are not propagating the warning anymore with this change :facepalm:

We need to return an `Either` or something to be able to distinguish the
possible outcomes.

kim:
Unsure if CI now fails because of the mutex actually working :joy_cat:

geigerzaehler (approved):

kim:
> if CI now fails because of the mutex actually working

Yes it did, because the test did encode an explicit rere, which was rejected as
already-in-flight. I am leaving the assertion in place (even though it conflates
test concerns), because it is testing that concurrency mechanism. I am also too
lazy right now to come up with a dedicated test.

kim:
> I am leaving the assertion in place

Well, scrap that. It is timing-dependent.

PR #780, opened 2021-09-01
Core-corrosion: Adding gitoxide to git-protocol integration tests

Byron:
A draft PR to get feedback early. I's goal is to add additional Rust to the core of git which is currently accessed through a thin corrosive layer provided by the `git2` crate.

By the end of this PR I'd expect all usages of `git2` to be replaced with at least equivalently readable substitutes provided by `gitoxide` except for where `git2` is explicitly under test.

Doing so is extremely valuable to `gitoxide` as it gets to develop its high-level-yet-performant API based on actual usage of a proven and mature library.

### DIFF
--- a/link-git-protocol/Cargo.toml
+++ b/link-git-protocol/Cargo.toml
@@ -24,7 +24,7 @@ tempfile = "3.2.0"
 versions = "3.0.2"
 
 [dependencies.git-repository]
-version = "0.9.0"
+version = "0.9.1"
 features = [ "async-network-client", "unstable" ]
 
 [dependencies.git-features]
--- a/link-git-protocol/Cargo.toml
+++ b/link-git-protocol/Cargo.toml
@@ -27,6 +27,18 @@ versions = "3.0.2"
 version = "0.9.0"
 features = [ "async-network-client", "unstable" ]
 
+[dependencies.git-features]
+version = "0.16.1"
+
+[dependencies.git-hash]
+version = "0.6.0"
+
+[dependencies.git-odb]
+version = "0.21.0"
+
+[dependencies.git-pack]
+version = "0.11.0"
+
 [dependencies.git-packetline]
 version = "0.10.0"
 features = ["async-io"]
--- a/link-git-protocol/Cargo.toml
+++ b/link-git-protocol/Cargo.toml
@@ -27,18 +27,6 @@ versions = "3.0.2"
 version = "0.9.1"
 features = [ "async-network-client", "unstable" ]
 
-[dependencies.git-features]
-version = "0.16.1"
-
-[dependencies.git-hash]
-version = "0.6.0"
-
-[dependencies.git-odb]
-version = "0.21.0"
-
-[dependencies.git-pack]
-version = "0.11.0"
-
 [dependencies.git-packetline]
 version = "0.10.0"
 features = ["async-io"]
--- a/test/src/test/integration/link_git_protocol.rs
+++ b/test/src/test/integration/link_git_protocol.rs
@@ -6,53 +6,56 @@
 use std::{
     collections::BTreeSet,
     io,
-    path::Path,
+    path::{Path, PathBuf},
     sync::{atomic::AtomicBool, Arc},
 };
 
 use bstr::ByteSlice as _;
 use futures::{AsyncReadExt as _, TryFutureExt as _};
+use git::{
+    prelude::*,
+    refs::transaction::{Change, PreviousValue, RefEdit},
+};
+use git_repository as git;
 use link_git_protocol::{fetch, ls, packwriter, upload_pack, ObjectId, PackWriter, Ref};
 use tempfile::{tempdir, TempDir};
 
 fn upstream() -> TempDir {
     let tmp = tempdir().unwrap();
 
-    let repo = git2::Repository::init_bare(&tmp).unwrap();
-    let auth = git2::Signature::now("apollo", "apollo@cree.de").unwrap();
+    let repo = git::init_bare(&tmp).unwrap().into_easy();
+    let auth = git::actor::Signature::now_local_or_utc("apollo", "apollo@cree.de");
 
-    let tree = {
-        let empty = repo.treebuilder(None).unwrap();
-        let oid = empty.write().unwrap();
-        repo.find_tree(oid).unwrap()
-    };
-    let base = {
-        let oid = repo
-            .commit(
-                Some("refs/namespaces/foo/refs/heads/main"),
-                &auth,
-                &auth,
-                "initial",
-                &tree,
-                &[],
-            )
-            .unwrap();
-        repo.find_commit(oid).unwrap()
-    };
-    let next = repo
-        .commit(
-            Some("refs/namespaces/foo/refs/heads/next"),
-            &auth,
-            &auth,
-            "ng",
-            &tree,
-            &[&base],
+    let empty_tree_id = repo
+        .write_object(&git::objs::Tree::empty())
+        .unwrap()
+        .detach();
+    let base_id = {
+        repo.commit(
+            "refs/namespaces/foo/refs/heads/main",
+            &auth.to_ref(),
+            &auth.to_ref(),
+            "initial",
+            empty_tree_id,
+            git::commit::NO_PARENT_IDS,
         )
-        .unwrap();
+        .unwrap()
+    };
+    let next_commit_id = repo
+        .commit(
+            "refs/namespaces/foo/refs/heads/next",
+            &auth.to_ref(),
+            &auth.to_ref(),
+            "ng",
+            empty_tree_id,
+            [base_id],
+        )
+        .unwrap()
+        .detach();
     repo.reference(
         "refs/namespaces/foo/refs/pulls/1/head",
-        next,
-        true,
+        next_commit_id,
+        PreviousValue::Any,
         "pee arrr",
     )
     .unwrap();
@@ -60,37 +63,47 @@ fn upstream() -> TempDir {
     tmp
 }
 
-fn collect_refs(repo: &git2::Repository) -> Result<Vec<(String, git2::Oid)>, git2::Error> {
+fn collect_refs(repo: &impl git::easy::Access) -> anyhow::Result<Vec<git::refs::Reference>> {
     repo.references()?
-        .map(|x| x.map(|r| (r.name().unwrap().to_owned(), r.target().unwrap())))
+        .all()?
+        .map(|x| x.map(|r| r.detach()).map_err(|err| anyhow::anyhow!(err)))
         .collect()
 }
 
-fn update_tips<'a, T>(repo: &git2::Repository, tips: T) -> Result<(), anyhow::Error>
+fn update_tips<'a, T>(repo: &impl git::easy::Access, tips: T) -> Result<(), anyhow::Error>
 where
     T: IntoIterator<Item = &'a Ref>,
 {
-    for r in tips {
-        match r {
-            Ref::Direct { path, object } => {
-                repo.reference(
-                    path.to_str()?,
-                    git2::Oid::from_bytes(object.as_slice())?,
-                    true,
-                    "",
-                )?;
-            },
+    use std::convert::TryInto;
+    let edits = tips
+        .into_iter()
+        .map(|r| match r {
+            Ref::Direct { path, object } => Ok(RefEdit {
+                change: Change::Update {
+                    log: Default::default(),
+                    expected: PreviousValue::Any,
+                    new: object.to_owned().into(),
+                },
+                name: path.as_bstr().try_into()?,
+                deref: false,
+            }),
             x => anyhow::bail!("unexpected ref variant: {:?}", x),
-        }
-    }
-
+        })
+        .collect::<Result<Vec<_>, _>>()?;
+    repo.edit_references(edits, git::lock::acquire::Fail::Immediately, None)?;
     Ok(())
 }
 
-fn collect_history(repo: &git2::Repository, tip: &str) -> Result<Vec<git2::Oid>, git2::Error> {
-    let mut revwalk = repo.revwalk()?;
-    revwalk.push_ref(tip)?;
-    revwalk.collect()
+fn collect_history(
+    repo: &impl git::easy::Access,
+    tip: &str,
+) -> anyhow::Result<Vec<git::hash::ObjectId>> {
+    repo.find_reference(tip)?
+        .into_fully_peeled_id()?
+        .ancestors()?
+        .all()
+        .map(|res| res.map(|oid| oid.detach()).map_err(Into::into))
+        .collect()
 }
 
 fn run_ls_refs<R: AsRef<Path>>(remote: R, opt: ls::Options) -> io::Result<Vec<Ref>> {
@@ -225,12 +238,13 @@ fn empty_fetch() {
 
 fn clone_with<R, L, B, P>(remote: R, local: L, build_pack_writer: B)
 where
-    R: AsRef<Path>,
-    L: AsRef<Path>,
+    R: Into<PathBuf>,
+    L: Into<PathBuf>,
     B: FnOnce(Arc<AtomicBool>) -> P,
     P: PackWriter + Send + 'static,
     P::Output: Send + 'static,
 {
+    let (remote, local) = (remote.into(), local.into());
     let refs = run_ls_refs(
         &remote,
         ls::Options {
@@ -255,9 +269,9 @@ where
 
     assert!(out.pack.is_some());
 
-    let remote_repo = git2::Repository::open(remote).unwrap();
+    let mut remote_repo = git::open(remote).unwrap().into_easy_arc_exclusive();
     remote_repo.set_namespace("foo").unwrap();
-    let local_repo = git2::Repository::open(&local).unwrap();
+    let local_repo = git::open(local).unwrap().into_easy();
 
     update_tips(&local_repo, &out.wanted_refs).unwrap();
 
@@ -276,7 +290,7 @@ fn clone_libgit() {
     let local = tempdir().unwrap();
     let local_repo = git2::Repository::init(&local).unwrap();
 
-    clone_with(&remote, &local, move |stop| {
+    clone_with(remote.path(), local.path(), move |stop| {
         packwriter::Libgit::new(packwriter::Options::default(), local_repo, stop)
     })
 }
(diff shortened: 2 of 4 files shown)


test/src/test/integration/link_git_protocol.rs
```diff
@@ -5,19 +5,109 @@
...
+                    mode: Create::Only,
+                    new: Target::Peeled(commit_id),
+                },
+                name: "refs/namespaces/foo/refs/heads/main".try_into().unwrap(),
+                deref: false,
+            },
+            git::lock::acquire::Fail::Immediately,
+            Some(&auth),
+        )
+        .unwrap();
```
Byron, line 59: There will be an equivalent `commit()`-like method soon to reduce this boilerplate.

kim:
>  all usages of git2

When you say "all", do you mean _all_?

There are _some_ uses in the tests which function like fixtures (but have random keys, so are not easy to generate upfront). But there are also other uses which are tied to the fact that `librad` is, well, tied to `git2`. I would suggest to scope this exercise the former.

Byron:
> When you say "all", do you mean all?

😅 All in this file except for the `git2` specific tests. The strategy here might be to look at git2 usages in tests from time to time and see if it's feasible to replace those with `gitoxide`, potentially implementing the missing bits and pieces. Maybe that way one will end up with an API surface big enough to be feasible for places outside of tests as well.

> I would suggest to scope this exercise the former.

Absolutely, I want it done yesterday to finally finish that git-fetch workblock I am still in, technically. I will keep that as MVP'ish as possible too to get the upload-pack building block more quickly.

Byron:
It feels I have been working on this for an eternity, so I did hope that it will 'just work' with some success. Will see what's causing the issue the gitoxide can't find an object that is definitely there but only if the repository was handled by `thin_pack_x`, which doesn't do much with these repositories.
It's very strange and I have the feeling an even stranger bug is to be found as cause of this.

On the bright side, one can already see how the `git-repository` API is shaping up, something like 2000 LOCs later.

Byron:
I think for the most part, using `git-repository` instead of `git2` now is a pleasant experience API wise. `gitoxide` definitely was inspired by `git2` in many places, while also trying to make the API more tree-like for chaining.

To make it nicer, I would love to remove some of the `unwrap()` calls in favor of returning results, which is done in some places but not everywhere. If that's agreeable, I can patch this.

Besides that, what do you think 🥺?

Byron:
As a side-note, while preparing this PR with crates from `crates.io` I managed to break the build surprisingly as this workspace then used incompatible crates in the same build. The problem causing this is well-known and not easy to tackle, I started a discussion over at [cargo smart-release](https://github.com/Byron/gitoxide/issues/192) to explore ways to make improve the situation.

There are ways to improve the situation already:

* pin all `gitoxide` crate versions at the cost of not getting patches automatically and upgrading periodically.
* consume all `gitoxide` code through `git-repository` without using the `unstable` flag. That's not feasible today as the API surface doesn't cover what's needed to interact with low-level `git-protocol` facilities. This would be my preferred choice as it would mean that `git-repository` indeed fulfils its promise of delivering highest performance with negligible overhead for the convenience layer it provides.
* `gitoxide` pre-release plumbing crates bump their dependents (those workspace crates who use it) minor version once their minor version changes to prevent auto-updates on the consumers side. This is what's (much better) described in the [`smart-release` issue](https://github.com/Byron/gitoxide/issues/192) I mentioned earlier.

test/src/test/integration/link_git_protocol.rs
```diff
@@ -6,91 +6,106 @@
...
+        )
+        .unwrap()
     };
-    let next = repo
+    let next_commit_id = repo
         .commit(
-            Some("refs/namespaces/foo/refs/heads/next"),
-            &auth,
-            &auth,
+            "refs/namespaces/foo/refs/heads/next",
```
Byron, line 46: Now that I see the diff, there is probably no good reason to change the order of arguments here. If there is agreement that should be fixed.

test/src/test/integration/link_git_protocol.rs
```diff
@@ -6,91 +6,106 @@
...
-            },
+    use std::convert::TryInto;
+    let edits = tips
+        .into_iter()
+        .map(|r| match r {
+            Ref::Direct { path, object } => Ok(RefEdit {
+                change: Change::Update {
+                    log: Default::default(),
+                    expected: PreviousValue::Any,
+                    new: git::hash::ObjectId::from(object.as_slice()).into(), // TODO: remove type conversion when possible
```
Byron, line 87: Tried again to remove this type conversion, to no avail. I don't understand why `git_hash::ObjectId` are not the same. Sure, they come from different crates but they are from the same source, `git_hash` in the same version.

test/src/test/integration/link_git_protocol.rs
kim: I don't mind, really, I can't remember 6 arguments anyway :)

No implicit `HEAD` is probably fair, but I'm surprised that the last 4 arguments are consumed.

test/src/test/integration/link_git_protocol.rs
```diff
@@ -6,91 +6,106 @@
...
-            &[&base],
+            auth.clone(),
+            auth,
+            empty_tree_id,
+            [base_id],
         )
-        .unwrap();
+        .unwrap()
+        .detach();
     repo.reference(
```
kim, line 55: I always found it a bit odd that `reference` _creates_ a reference, but well names

test/src/test/integration/link_git_protocol.rs
```diff
@@ -285,21 +301,23 @@ fn clone_libgit() {
...
-    let local_repo = git2::Repository::init(&local).unwrap();
+    let local_repo = git::init(&local).unwrap();
 
-    clone_with(&remote, &local, move |stop| {
+    clone_with(remote.path(), &local.path(), move |stop| {
         packwriter::Standard::new(local_repo.path(), packwriter::Options::default(), stop)
     })
 }
 
+#[allow(dead_code)]
```
kim: Needed?

test/src/test/integration/link_git_protocol.rs
```diff
@@ -317,13 +335,17 @@ where
         assert!(out.pack.is_some());
     }
 
-    let remote_repo = git2::Repository::open(&remote).unwrap();
+    let mut remote_repo = git::open(remote.clone()).unwrap().into_easy_arc_exclusive(); // without GATs we need the arc version
```
kim, line 335: GATs would be needed on gitoxide's side?

test/src/test/integration/link_git_protocol.rs
```diff
@@ -341,6 +363,8 @@ where
         update_tips(&local_repo, &out.wanted_refs).unwrap();
     }
 
+    // Need to refresh it as it didn't notice the new pack
```
kim, line 363: Interesting.. what does it exactly mean to `open` a repository, in terms of I/O, file handles, etc? 

I'm asking because `git2` does a **lot** actually (including resolving and parsing config files), so we ended up using `git2::Repository` from a (database) pool to avoid fd exhaustion and other issues. I suspect that we'd not use easy mode in a lot of places in order to be able to carefully manage resource consumption, but I think documentation about what the cost of an easy repo is would help.

test/src/test/integration/link_git_protocol.rs
kim: > git_hash in the same version.

That is a bit concerning -- I'm not sure I understand how Byron/gitoxide#192 would explain this.

FintanH:
> As a side-note, while preparing this PR with crates from crates.io I managed to break the build surprisingly as this workspace then used incompatible crates in the same build. 

One of my [builds](https://buildkite.com/monadic/radicle-link-seedling/builds/3363#b2bbc323-9acf-4014-90bc-2899aeebd418) got bitten by this too :grimacing:

kim:
> * consume all `gitoxide` code through `git-repository`

If I understand the problem correctly, then `git-repository` is depending on
_newer_ git-* crates, which we don't pull in because they are breaking as per
semver. Going forward, however, I would expect `git-repository` to be **more**
conservative than lower level crates, so we may not get functionality that is
considered unstable until there is a release of `git-repository`.

So I'm not sure: either `git-repository` re-exports everything else, and we can
control via an `unstable` feature whether we want the bleeding edge or not, or
we would probably maintain our own prelude crate with pinned versions, to avoid
having to manage those versions across multiple of our own crates. But I _think_
the latter would preclude depending on `git-repository`, as we enter crate hell
through the other door then.

test/src/test/integration/link_git_protocol.rs
Byron: > I don't mind, really, I can't remember 6 arguments anyway :)

Ha, fair enough. It's not as bad as I made it sound, it's 'just' the message that moved underneath the author and committer fields. Keeping more similar would help with transitioning code, but then again, there is enough that's very different. I will just keep it like that then.

> No implicit HEAD is probably fair, but I'm surprised that the last 4 arguments are consumed.

I would be too, and there is a reason that hopefully actually is one in practice: APIs won't hide that they need to own an object by only asking for a reference. That way, if the caller has that object to spare, it's free. If the signature of a function asks for reference types, even if an owned version is passed in as `AsRef<oid>` or `AsRef<str>`, internally one has to go through a ref like `input.as_ref().to_owned()` even though the caller did something like `f(x)` and is happy that way. This adds a burden to callers as they have to clone if needed.

test/src/test/integration/link_git_protocol.rs
Byron: 🤦‍♂️ fixed!
There was a `git-hash` v0.5 lingering around and I didn't notice it when checking `cargo tree -v git-hash`. Now they are all the same version and `cargo` behaves as `cargo` behaves.

test/src/test/integration/link_git_protocol.rs
Byron: Me too! And it was aptly called `create_reference` before. But as I worked with this it dawned on me how `libgit2` (or `git2`) is thinking and I started liking it and thought it was beneficial to not unnecessarily change names.

This also means that `reference**s**` now creates an intermediate object for iterators, and you guessed it, that was called `iter_references()` before I took the U-turn there as well.

I don't know if it's _good_ to be so `git2` inspired there, and already thought providing an alternative name for the same functionality potentially even behind feature 

[2000 words later]… names _are_ hard.

test/src/test/integration/link_git_protocol.rs
Byron: Yes, just there. [Here is the code for that](https://github.com/Byron/gitoxide/blob/7c559d6e1b68bc89220bca426257f383bce586ae/git-repository/src/easy/impls.rs#L72-L88) and the reason mutable `RefMut<'argh_a_lifetime_that_cannot_be_expressed, Repository>` can't be done yet.
```suggestion
    let mut remote_repo = git::open(remote.clone()).unwrap().into_easy_arc_exclusive(); // without GATs we need the arc version
```
```suggestion
    let mut remote_repo = git::open(remote.clone()).unwrap().into_easy_arc_exclusive(); // without GATs we need the arc version
```

test/src/test/integration/link_git_protocol.rs
kim: Let me rephrase that: all arguments are `impl (Try)Into<Owned>`, which suggests that a copy occurs if a reference is passed. It might be that `git2` is lying, but it is still surprising that this cannot be done "zero copy", especially for object ids. So, why is that?

test/src/test/integration/link_git_protocol.rs
Byron: Indeed, documentation is non-existing in places and I plan to rectify that with a quick-pass of doc-writing before I move on.

Right now `open(…)` doesn't do much except for mapping all packs in full and reading the local repository config file once. Everything else is lazy.

On the server you would do the same, but instead of holding a `git::Repository` you would hold a `git::EasyArcExclusive`. That way, access to the persistent pack memory maps is shared, but caches and on-demand maps (like packed-refs) are not.

Even though it's not implemented, I would like to have a `refresh()` to reload the object database. And thinking about it, it's easy to demo this API here but would then force the `local` repo to be an `EasyArcExclusive` as well.

Probably worth sketching this out while we are here.

Note that `git-config` will be fully loaded into memory so no handles will be left open.

test/src/test/integration/link_git_protocol.rs
Byron: > […] but it is still surprising that this cannot be done "zero copy", especially for object ids. So, why is that?

Amazing, I am glad you ask because I already took it for granted even though there isn't any other reason that I only implemented serialization for owned objects, not for (zero-copy) parsed ones. 

the `gixp pack-verify` program has a flag to re-encode all objects it sees to see if its rountrippable. Amazingly this is only about a tiny amount of time smaller than not doing that, so it might not be worth optimizing for this case.

However, it can definitely be done and I'd say it's not particularly hard. Those who have owned objects could then call `to_ref()` and get the immutable version that can then be serialized.

Sounds like it's old tech-debt that should probably be fixed once and for all. What do you think?

kim: Thanks for the explanation! 

It is probably not a bottleneck, but _maybe_ something we'd want to watch out for in the future. I just thought I'll ask, because during the making-of easy a lot of effort went into avoiding allocations for object ids.

Byron:
>  I would expect `git-repository` to be **more**
conservative than lower level crates, so we may not get functionality that is
considered unstable until there is a release of `git-repository`.

That's correct, it's in [stability tier 1](https://github.com/Byron/gitoxide/blob/main/STABILITY.md#tiers). It re-exports all crates in the same stability tier and hides everything else.

> o I'm not sure: either `git-repository` re-exports everything else, and we can
control via an `unstable` feature […]

Such a feature toggle exists and is used to re-exports all lower tier crates. Along with a new version bumping scheme described [in this cargo smart-release](https://github.com/Byron/gitoxide/issues/192) ticket, if any of the plumbing crates have breaking changes with a minor bump, `git-repository` would also have a minor bump. That way you would get patches automatically but not more, i.e. breaking changes. This would allow me to even post PRs for upgrades for you without your CI breaking surprisingly.

For a lack of a better idea, this would be my preference, as it seems more powerful than pinning all `git-*` crates to their version.

Thinking into the future of post-release crates, we would eventually stop using the `unstable` feature to get all the benefits of stability tier 1 while lower tier crates may still change, but you won't be using them directly anymore.

Byron:
### Pending Changes

* [ ] Show how packs can be refreshed after pull/fetch
* [ ] See if writing ObjectRefs is as easy as it sounds for consistency and unbound performance

test/src/test/integration/link_git_protocol.rs
```diff
@@ -6,91 +6,104 @@
...
+
+    let empty_tree_id = repo
+        .write_object(&git::objs::Tree::empty())
+        .unwrap()
+        .detach();
+    let base_id = {
+        repo.commit(
+            "refs/namespaces/foo/refs/heads/main",
+            &auth.to_ref(),
+            &auth.to_ref(),
```
Byron, line 37: While I was doing breaking changes, I unified parameter order 😅

test/src/test/integration/link_git_protocol.rs
```diff
@@ -6,91 +6,104 @@
...
+    let auth = git::actor::Signature::now_local_or_utc("apollo", "apollo@cree.de");
+
+    let empty_tree_id = repo
+        .write_object(&git::objs::Tree::empty())
+        .unwrap()
+        .detach();
+    let base_id = {
+        repo.commit(
+            "refs/namespaces/foo/refs/heads/main",
+            &auth.to_ref(),
```
Byron, line 36: Unfortunately the `to_ref()` is necessary as we can't have real refs. The `*Ref` is borrowed, too, for maximum flexibility, or else those who can't move these have to clone them. That's cheap, but I guess it's the question of who has to make the call.

I could imagine introducing a trait for `ToRef` which eliminates the `to_ref()` call.

test/src/test/integration/link_git_protocol.rs
```diff
@@ -6,91 +6,104 @@
...
-                &tree,
-                &[],
-            )
-            .unwrap();
-        repo.find_commit(oid).unwrap()
+    let repo = git::init_bare(&tmp).unwrap().into_easy();
+    let auth = git::actor::Signature::now_local_or_utc("apollo", "apollo@cree.de");
+
+    let empty_tree_id = repo
+        .write_object(&git::objs::Tree::empty())
```
Byron, line 30: Now one can write object refs as well as owned objects, and there is a trait for that so calls look as good as they can.

test/src/test/integration/link_git_protocol.rs
Byron: Oh, and before I forget: the refs are a lie, internally it still creates an owned `Commit` as it's easier to work with. Namely, object ids are in hex but nobody really has these, so they would need internal conversion and memory to hold that. A possible optimization, but one I skipped 😅.

test/src/test/integration/link_git_protocol.rs
```diff
@@ -341,6 +360,8 @@ where
         update_tips(&local_repo, &out.wanted_refs).unwrap();
     }
 
+    // Need to refresh it as it didn't notice the new pack
+    local_repo.refresh_object_database().unwrap();
```
Byron, line 364: This is exactly what would have to be done on the server as well once it causes new packs to be written. Internally it currently replaces itself, which is certainly something that could be improved one day.

Byron:
I squashed everything into one commit and rebased against master. Is there anything else that would prevent merging this PR?

link-git-protocol/Cargo.toml
```diff
@@ -25,7 +25,20 @@ versions = "3.0.2"
 
 [dependencies.git-repository]
 version = "0.9.0"
-features = [ "async-network-client", "unstable" ]
+default-features = false # leave out max-performance as it doesn't build on msvc win due to sha1-asm. See https://github.com/RustCrypto/asm-hashes/issues/17 for details
+features = [ "one-stop-shop", "async-network-client", "unstable" ]
+
+[dependencies.git-features]
```
FintanH, line 31: Why are we adding these back rather than accessing them through `git-repository`?

FintanH (commented):
LGTM for the tests, just have the one question :)

link-git-protocol/Cargo.toml
Byron: It adds all of the default features back except for `max-performance`, which unfortunately breaks the build.

I was thinking that it might be better to remove `max-performance` from the default feature set for now, but felt bad about it.

However, now that I wrote this I revisited the issue and fixed it on `git-repository` side as the superior choice :).

kim: I'm confused, too. It seems like we need `git-repository` and `git-packetline`, but not the rest. Maybe a rebase went wrong or something?

Byron: Great catch! And even though I don't know how that happened, removing the superfluous packages did indeed not cause any trouble.

FintanH (approved):
I approve this message :+1:

kim:
Closed via 07013fc0e618bc112ad9019e34ed533b946eb0d5

Thanks!

Byron:
Please note in order to provide actual protection from unexpected breakage coming up through the crates exposed by `git-repository` `cargo-smart-release` will have to [learn a new trick](https://github.com/Byron/gitoxide/issues/192).  For now I have manually bumped `git-repository`'s minor version to assure of that.

PR #783, opened 2021-09-03
Distinguish between channel full and closed in error handling

adaszko:


seed/src/handle.rs
```diff
@@ -31,6 +31,15 @@ pub enum NodeError {
     RequestTimeout(#[from] time::error::Elapsed),
 }
 
+impl NodeError {
+    fn from_send_error<T>(try_send_error: TrySendError<T>) -> Self {
+        match try_send_error {
+            TrySendError::Full(_) => Self::RequestSendFailedChannelFull,
+            TrySendError::Closed(_) => Self::RequestSendFailedChannelClosed,
```
xla, line 38: Wouldn't it be more efficient to use the `from` directive from the `thiserror` crate and then let the `?` work its magic, instead of having to maintain two error variants and map them manually.

kim:
Would be nice to amend the commit message in the form:

seed: what

why

Signed-off-by:

seed/src/handle.rs
adaszko: That would entail needlessly passing around `TrySendError`'s type argument.

kim: Which you could monomorphise to `Request`?

adaszko: And keep a value of type `oneshot::Sender<>` in an error enum?  I don't think that's such a good idea.  The error value is meant to be printed and matched against.  It only complicates things to keep a channel in it.

kim: fair

kim (approved):
lgtm

PR #786, opened 2021-09-14
rad binaries

FintanH:
Since we now have a `bins` section, I'm adding the `rad`, `rad-identities-dev`, and `rad-profile-dev` binaries there. Note that I'm building this on top of `finto/rad-identities`.

### DIFF
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -69,3 +69,16 @@ steps:
       - "Cargo.lock"
     agents: *build-agent
     env: *build-docker
+
+  - label: "Build Bins"
+    timeout_in_minutes: 20
+    if: |
+        build.pull_request.repository.fork == null ||
+        build.pull_request.repository.fork == false
+    commands:
+      - ".buildkite/env"
+      - "ci/build-bins"
+    artifact_paths:
+      - "Cargo.lock"
+    agents: *build-agent
+    env: *build-docker
--- a/bins/rad-identities-dev/Cargo.toml
+++ b/bins/rad-identities-dev/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "rad-identities-dev"
+version = "0.1.0"
+authors = [ "Fintan Halpenny <fintan.halpenny@gmail.com>" ]
+edition = "2018"
+license = "GPL-3.0-or-later"
+
+[dependencies]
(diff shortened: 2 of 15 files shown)


kim:
Out of curiosity: is there a deeper reason to prefer crate-per-binary over just one crate with multiple bin targets?

FintanH:
> Out of curiosity: is there a deeper reason to prefer crate-per-binary over just one crate with multiple bin targets?

My thinking was that dependencies are separated and we continue on the small, well-defined crates path. What do you think?

bins/rad-identities-dev/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "rad-identities-dev"
+version = "0.1.0"
+authors = ["Fintan Halpenny fintan.halpenny@gmail.com"]
```
kim, line 4: does cargo not want some angle brackets around the email address?

FintanH: Yes, apparently[0]

[0]: https://doc.rust-lang.org/cargo/reference/manifest.html#the-authors-field

kim:
> My thinking was that dependencies are separated and we continue on the
> small, well-defined crates path. What do you think?

I think that it will most likely become a bit of a versioning nightmare once we
publish actual cargo versions. But we can revisit once we get there.

.buildkite/pipeline.yaml
```diff
@@ -69,3 +69,16 @@ steps:
       - "Cargo.lock"
     agents: *build-agent
     env: *build-docker
+
+  - label: "Build Bins"
```
xla, line 73: Mea culpa!

bins/rad-identities-dev/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+[dependencies]
+anyhow = "1"
+structopt = "0.3"
+
+[dependencies.rad-exe]
+path = "../../rad-exe"
+
+[dependencies.rad-identities]
+path = "../../rad-identities"
+
```
xla, line 17: Generous newline?

FintanH: Was feeling nice

xla (approved):
🛋 🔣 🍡 🌟

PR #787, opened 2021-09-16
Move crates into crates/ directory

adaszko:
The directory structure attempts to follow the guidelines laid out at https://matklad.github.io/2021/08/22/large-rust-workspaces.html

Why do this?  This is a necessary step towards merging radicle-bins into this repo.  Why merge the two repos you ask?  Having two repos means there are often two places that need to be updated to deploy a change, two CI systems.  It also means there's a need to reference radicle-link repo from radicle-bins and that reference needs to be updated and maintained unnecessarily.   AFAIK, the only reason there's a separate repository for executables was that it wasn't clear at the time that cargo supports that and it's in fact pretty widely practiced in the Rust ecosystem, see e.g. [ripgrep](https://github.com/BurntSushi/ripgrep), [rust-analyzer](https://github.com/rust-analyzer/rust-analyzer), etc.

Anything else needs to be updated that I missed?


kim:
Thanks, but no thanks. 


While that article is an interesting read, there is no interest to maintain the executables currently found in radicle-bins within this repo.

Next time you propose a large change like this, it would also be appreciated to start a discussion beforehand.

